PR #10, opened 2021-01-19
mitx bigquery pipeline

abeglova:


src/ol_data_pipelines/mitx_bigquery/solids.py
```diff
@@ -0,0 +1,208 @@
...
+        context.solid_config["outputs_dir"]
+    )
+
+    for dataset in datasets:
+        table_name = (
+            context.resources.bigquery_db.project
+            + "."
+            + dataset.dataset_id
+            + ".user_info_combo"
+        )
```
pdpinch, line 125: Where does this table_name pattern come from? Is this a convention from IRx and edx2bigquery?

abeglova: Yes, it's a convention in edx2bigquery

abeglova: I looked at the data in the bigquery database and that's how the data was organized. I don't know if there is documentation anywhere with the schema for that database

src/ol_data_pipelines/mitx_bigquery/solids.py
```diff
@@ -0,0 +1,208 @@
...
+        "certificate_status",
+        "certificate_verify_uuid",
+        "certificate_download_uuid",
+        "certificate_name",
+        "certificate_created_date",
+        "certificate_modified_date",
+        "certificate_error_reason",
+        "certificate_mode",
+        "id_map_hash_id",
+    ]
```
blarghmatey, line 113: How many of these fields do we have a current use for? I think it's best for us to minimize the amount of information that we are replicating and only pull what we actually need for this project. If we determine in the future that we need additional attributes then we can retrieve them at that point.

pdpinch:
Maybe worth a comment in the code?

> I looked at the data in the bigquery database and that's how the data was organized. I don't know if there is documentation anywhere with the schema for that database

src/ol_data_pipelines/mitx_bigquery/solids.py
```diff
@@ -0,0 +1,208 @@
...
+                )
+
+                with file_system.open_output_stream(file_path) as file:
+                    with parquet.ParquetWriter(file, arrow_table.schema) as writer:
+                        writer.write_table(arrow_table)
+
+        except NotFound:
+            pass
+
+    yield AssetMaterialization(
```
blarghmatey, line 153: It might be useful to put the asset materialization inside the loop to show which courses are being processed. It's also helpful to add additional metadata such as the number of rows processed.

abeglova:
Thanks @blarghmatey. I made the requested changes

src/ol_data_pipelines/mitx_bigquery/solids.py
```diff
@@ -147,16 +118,26 @@ def download_user_data(
                     with parquet.ParquetWriter(file, arrow_table.schema) as writer:
                         writer.write_table(arrow_table)
 
+                yield AssetMaterialization(
+                    description="Updated course file",
+                    asset_key="course_file_update",
```
blarghmatey, line 123: We should set the asset key as the file path so that each asset is tracked independently.

blarghmatey:
:+1:

PR #13, opened 2021-01-27
add enrollment creation timestamp to pipleline data

abeglova:
I did not include the enrollment creation timestamp as a field we want to pull in the original PR. This is useful information in general and also will make incremental updates to the open enrollment data possible.


blarghmatey:
:+1:

PR #20, opened 2021-02-03
Ab/pull open data

abeglova:
This pr creates a pipeline for downloading user and run data from the open database to file/s3 as parquet files.
This is part of https://github.com/mitodl/ol-data-pipelines/issues/18

Sample config (with database keys removed):
```
resources:
  postgres_db:
    config:
      dbname: opendiscussions
      host: 
      password: 
      port: 5432
      user:
solids:
  download_run_data:
    config:
      outputs_dir: /etc/dagster/open_run_data
      file_base: run_data
  download_user_data:
    config:
      file_base: run_data
      outputs_dir: /etc/dagster/open_user_data
```

### DIFF
--- a/src/ol_data_pipelines/open_discussions/solids.py
+++ b/src/ol_data_pipelines/open_discussions/solids.py
@@ -1,0 +1,184 @@
+"""Pipeline for uploading open-discussions user and run data to s3."""
+from dagster import (  # noqa: WPS235
+    Field,
+    Int,
+    ModeDefinition,
+    Noneable,
+    Output,
+    OutputDefinition,
+    PresetDefinition,
+    SolidExecutionContext,
+    String,
+    pipeline,
+    solid,
+)
+from pyarrow import fs
+from pypika import PostgreSQLQuery as Query
+from pypika import Table, Tables
+
+from ol_data_pipelines.lib.arrow_helper import stream_to_parquet_file
+from ol_data_pipelines.lib.yaml_config_helper import load_yaml_config
+from ol_data_pipelines.resources.postgres_db import postgres_db_resource
+
+
+@solid(
+    description=("Query postgres and data as parquet files "),
+    required_resource_keys={"postgres_db"},
+    config_schema={
+        "chunksize": Field(
+            Int,
+            is_required=False,
+            default_value=5000,
+            description="Number of rows per parquet file",
+        ),
+        "outputs_dir": Field(
+            String, is_required=True, description="Path for output files"
+        ),
+        "file_base": Field(String, is_required=True, description="base_name_for_file"),
+        "minimum_timestamp": Field(
+            Noneable(str),
+            is_required=False,
+            default_value=None,
+            description="Minimum minimum_timestamp for user date_joined. Format yyyy-mm-dd",
+        ),
+    },
+    output_defs=[
+        OutputDefinition(
+            name="query_folder",
+            dagster_type=String,
+            description="Path to user data rendered as parquet config_files",
+        )
+    ],
+)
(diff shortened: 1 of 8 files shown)


src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -0,0 +1,208 @@
...
+    },
+    output_defs=[
+        OutputDefinition(
+            name="query_folder",
+            dagster_type=String,
+            description="Path to user data rendered as parquet config_files",
+        )
+    ],
+)
+def download_postgres_data(context: SolidExecutionContext, query: String):
```
blarghmatey, line 131: I think that `download` is a bit of a misnomer. I also think that encoding information about an implementation detail of the systems isn't necessary. We should aim to have the solid definitions be descriptive of _what_ is happening, rather then _where_ it's happening. We should probably also establish a controlled vocabulary for our pipelines to avoid any ambiguity, but for now perhaps `transfer_courserun_ids` or `extract` or `fetch`.

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -0,0 +1,208 @@
...
+            metadata_entries=[
+                EventMetadataEntry.text(
+                    label="updated_file",
+                    description="updated data file",
+                    text=output_name,
+                ),
+            ],
+        )
+
+    yield Output(output_resource.output_folder, "query_folder")
```
blarghmatey, line 171: If the purpose of creating the class object for the Parquet output is to maintain a reference to the output attribute, perhaps we should create a standard [Pydantic](https://pydantic-docs.helpmanual.io/) model that we store as a `lib` which we can hydrate with that information and which we can attach validation logic to.

blarghmatey (changes requested):

src/ol_data_pipelines/lib/parquet_file_rendering.py
```diff
@@ -0,0 +1,27 @@
...
+from pyarrow import fs, parquet
+
+
+class ParquetOutputResource:
+    """helper class for parquet data output"""
+
+    def __init__(self, output_dir, clear_existing_files):
+        self.file_system, self.output_folder = fs.FileSystem.from_uri(output_dir)
+
+        if clear_existing_files:
```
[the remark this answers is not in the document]
blarghmatey: So, given that we are writing the data out as S3 there isn't any option for find/replace/update rows as the unit of work is at the level of a Parquet file. Since the tables are small, we should probably be writing them out as a single Parquet file, so if we simply write out the data it will automatically overwrite the data that is available in S3.

src/ol_data_pipelines/open_discussions/solids.py
blarghmatey: My previous comment was about the prior implementation of the parquet "resource", not about the output being produced.

That being said, if there is no meaningful output for a given solid then the supported pattern is to yield an `Output` of type `dagster.Nothing` https://docs.dagster.io/_apidocs/types

src/ol_data_pipelines/lib/parquet_file_rendering.py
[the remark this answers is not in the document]
abeglova: There are currently 123,054 users on production, which can definitely be one file. RC has 97352 and I had no problem querying for all of them at once but probably better not to in production. I'll replace `stream_parquet_files`  function to query in chunks than writes everything in one file

blarghmatey (approved):

blarghmatey:
I think this looks good for an initial pass. :+1:

PR #22, opened 2021-02-19
ab/fix-timestamp-format

abeglova:
closes https://github.com/mitodl/ol-data-pipelines/issues/21

Timestamps imported from parquet to athena are corrupted. By default parquet 2.0 saves timestams in int64 format but amazon glue crawlers expect int96 format.

Setting use_deprecated_int96_timestamps=True on the writer fixes the issue


blarghmatey:
:+1:

PR #24, opened 2021-02-22
add course table to open pipeline

abeglova:
We don't always have a match between run codes in the open database and the bigquery database for older course runs. It is useful to try matching on just the course code for enrollments where we cannot find a match based on the run. This PR updates the open data pipeline to pull data from the courses table. 

It also makes an update to the runs query to exclude learning resource runs that are not associated with a course object since there are learning_resource_run objects associated with non-course learning resources such as videos.

### DIFF
--- a/src/ol_data_pipelines/open_discussions/solids.py
+++ b/src/ol_data_pipelines/open_discussions/solids.py
@@ -34,7 +34,12 @@ from ol_data_pipelines.resources.postgres_db import postgres_db_resource
         "outputs_dir": Field(
             String, is_required=True, description="Path for output files"
         ),
-        "file_base": Field(String, is_required=True, description="base_name_for_file"),
+        "file_base": Field(
+            String,
+            is_required=False,
+            default_value="open_user_data",
+            description="base_name_for_file",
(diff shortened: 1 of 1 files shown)


src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -157,6 +165,87 @@ def fetch_open_run_data(context: SolidExecutionContext):  # noqa: WPS210
...
+        "chunksize": Field(
+            Int,
+            is_required=False,
+            default_value=5000,
+            description="Number of rows per parquet file",
+        ),
+        "outputs_dir": Field(
+            String, is_required=True, description="Path for output files"
+        ),
+        "file_base": Field(String, is_required=True, description="base_name_for_file"),
```
blarghmatey, line 183: Might make sense to add a default value for this field.

blarghmatey:
Other than the above comment I think this is good to go.

abeglova:
Thanks @blarghmatey. I made the requested change

blarghmatey:
:+1:

PR #25, opened 2021-02-28
Ab/enrollments bulk update

abeglova:


src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+
+@resource(
+    config_schema={
+        "aws_access_key_id": Field(
+            String, is_required=True, description="Host string for MySQL/MariaDB server"
+        ),
+        "aws_secret_access_key": Field(
+            String,
+            is_required=True,
+            description="TCP Port number for MySQL/MariaDB server",
```
blarghmatey, line 66: As above, these should be optional parameters to take advantage of instance profiles and other means of loading credentials automatically.

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        "work_group": Field(
+            String,
+            is_required=True,
+            description="Username for authenticating to MySQL/MariaDB server",
```
blarghmatey, line 71: This isn't a MySQL/MariaDB server...

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -287,3 +474,36 @@ def pull_open_data_pipeline():
     fetch_open_run_data()
     fetch_open_course_data()
     fetch_open_user_data()
+
+
+@pipeline(
+    description="Update open enrollments",
```
blarghmatey, line 480: This pipeline description is a bit incomplete. Would be good to call out the fact that the data is reflected into S3, combined with other data sources, and then data is written back to Open.

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -261,6 +267,187 @@ def fetch_open_course_data(context: SolidExecutionContext):  # noqa: WPS210
...
+        .on_conflict("user_id", "enrollments_table_run_id")
+        .do_update("updated_on", timestamp)
+        .do_update("enrollment_timestamp", Excluded("enrollment_timestamp"))
+        .do_update("run_id", Excluded("run_id"))
+        .do_update("course_id", Excluded("course_id"))
+    )
+
+    insert_query = base_insert_query
+
+    for enrollment in query_results:
```
blarghmatey, line 428: Having this write directly to the database might end up being brittle as the schema evolves. Might be worth creating an API on Open to be able to handle these updates?

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -261,6 +267,187 @@ def fetch_open_course_data(context: SolidExecutionContext):  # noqa: WPS210
...
+                athena_enrollments.enrollment_created
+                > Cast(context.solid_config["minimum_timestamp"], "DATE")
+            )
+            | (
+                users.date_joined
+                > Cast(context.solid_config["minimum_timestamp"], "DATE")
+            )
+        )
+
+    query_results = context.resources.athena_db.run_query(athena_query)
```
blarghmatey, line 401: It might be worth splitting this into two solids with the query results as the output of the first stage and a separate solid for handling the inserts.

src/ol_data_pipelines/lib/pypika_helper.py
```diff
@@ -0,0 +1,17 @@
+"""Helper class for custom pypika functions."""
+
+from pypika import CustomFunction
+
+SplitPart = CustomFunction("SPLIT_PART", ["string", "delimiter", "part"])
+
+Position = CustomFunction("POSITION", ["input"])
+
+MinBy = CustomFunction("MIN_BY", ["value1", "value2"])
```
blarghmatey, line 9: If these are specific to Presto SQL syntax then it might be worth either adding a comment to that effect, or encapsulating into a specific Query syntax class.

src/ol_data_pipelines/open_discussions/solids.py
abeglova: I think we should definitely consider this if we find that we need to change the database schema often. However, we would need to make changes to the query generating the data for the bulk update anyway if there is a schema change.

If we did create an api for bulk updates on open a good process would be uploading the raw bulk update data to s3, pinging the open api, then having open make database updates.

We can consider this as follow up work if we find that the current process is not working well enough

abeglova:
Thank @blarghmatey. This is ready for re-review

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -301,8 +296,15 @@ def fetch_open_course_data(context: SolidExecutionContext):  # noqa: WPS210
...
+    output_defs=[
+        OutputDefinition(
+            name="enrollment_data_cursor",
+            dagster_type=DictCursorDagsterType,
+            description="Cursor which returns the results of a query joining users, runs, courses and enrollments",
+        )
+    ],
 )
-def update_open_enrollments(  # noqa: WPS210
+def get_open_enrollments_cursor(  # noqa: WPS210
```
blarghmatey, line 307: I don't think that returning a cursor object would work properly. I think it makes more sense to yield the result set as an output. If we expect to have to batch the results then we can possibly perform a dynamic graph such that the downstream solid will be replicated for each result batch https://docs.dagster.io/examples/dynamic_graph

src/ol_data_pipelines/open_discussions/solids.py
```diff
@@ -425,7 +461,7 @@ def update_open_enrollments(  # noqa: WPS210
 
     insert_query = base_insert_query
 
-    for enrollment in query_results:
+    for enrollment in enrollment_data_cursor:
```
blarghmatey, line 464: As mentioned above, I don't think passing the cursor around makes sense, particularly because there is the distinct possibility that these two steps won't even execute on the same machine, let alone the same process. The dependency chain should be oriented around the data rather than an arbitrary object.

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -34,14 +34,22 @@ def __init__(  # noqa: WPS211
...
         """
-        self.connection = pyathena.connect(
-            aws_access_key_id=aws_access_key_id,
-            aws_secret_access_key=aws_secret_access_key,
-            work_group=work_group,
-            cursor_class=DictCursor,
-            region_name=region_name,
-            schema_name=schema_name,
-        ).cursor()
+        if aws_access_key_id and aws_secret_access_key:
```
blarghmatey, line 37: It should be possible to simply make the access key and secret optional and pass them as `None` when they are not present.

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+from dagster import Field, InitResourceContext, String, resource
+from pyathena.cursor import DictCursor
+from pypika import Query
+
+
+class AthenaClient:
+    def __init__(  # noqa: WPS211
+        self,
+        aws_access_key_id: Text,
+        aws_secret_access_key: Text,
```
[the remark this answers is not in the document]
blarghmatey: These should be keyword args with a default value of `None`

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,108 @@
...
+        work_group: Text,
+        region_name: Text,
```
blarghmatey, line 17: This can be a keyword arg with a default value of `"us-east-1"`

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,108 @@
...
+            description="Aws key for Athena connection",
+        ),
+        "work_group": Field(
+            String,
+            is_required=True,
+            description="Work group for Athena connection",
+        ),
+        "region_name": Field(
+            String,
+            is_required=True,
```
blarghmatey, line 83: This can be `is_required=False` with a default value of `"us-east-1"`

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@resource(
+    config_schema={
+        "aws_access_key_id": Field(
+            String, is_required=False, description="AWS key for Athena connection"
+        ),
+        "aws_secret_access_key": Field(
+            String,
+            is_required=False,
```
blarghmatey, line 73: This either needs to be passed a default value of `None` or the code in the function body needs to use a `.get` call on the `resource_config` object.

blarghmatey (changes requested):

blarghmatey:
As part of the work for https://github.com/mitodl/open-discussions/issues/3318 I took the opportunity to improve the structure of the Athena/S3 system which also involved changing database and table naming which will require some changes to the query in this PR. I think that this is also a good opportunity to add the integration with AWS Glue for registering the table/schema definitions of the data being loaded as part of the pipeline concern, rather than requiring configuration of the Glue crawler and waiting for it to execute.

abeglova:
AWS glue crawler only has to run if the data schema changes, correct? I'm not sure how useful it would be to run the crawler automatically as part of this pipeline since schema changes should occur infrequently and also the queries in the pipeline would not update automatically to a new schema.

blarghmatey:
My intent is to move away from using the crawler entirely and instead use the Glue catalog API (https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-catalog.html) directly to register tables and their schema. The crawler is intended to be used for registering data sources that are not managed programmatically and for which the schema is largely unknown.

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+
+        :returns: DictCursor
+        """
+        return self.connection.execute(str(query))
+
+
+@resource(
+    config_schema={
+        "aws_access_key_id": Field(
+            String, is_required=True, description="Host string for MySQL/MariaDB server"
```
blarghmatey: This isn't an accurate description of what the parameter is

src/ol_data_pipelines/resources/athena_db.py
blarghmatey: This is not an accurate description of what the parameter is

src/ol_data_pipelines/resources/athena_db.py
blarghmatey: This is still an inaccurate representation of what the parameter is

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        "work_group": Field(
+            String,
+            is_required=True,
+            description="Username for authenticating to MySQL/MariaDB server",
+        ),
+        "region_name": Field(
+            String,
+            is_required=True,
+            description="Password for authenticating to MySQL/MariaDB server",
```
blarghmatey: This is not an accurate representation of what the parameter is

src/ol_data_pipelines/resources/athena_db.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        "schema_name": Field(
+            String,
+            is_required=True,
+            description="Password for authenticating to MySQL/MariaDB server",
```
blarghmatey: This is not an accurate description of the parameter

abeglova: You are commenting on an old commit

src/ol_data_pipelines/resources/athena_db.py
abeglova: You are commenting on an old commit

src/ol_data_pipelines/resources/athena_db.py
blarghmatey: Sorry, my bad :/

blarghmatey:
:+1:

PR #78, opened 2021-04-26
update to dagster 0.11.5

abeglova:
Dependabot updated dagster-aws and dagster-shell to version 0.11.5 

This updates dagster and dagit to 0.11.5 as well so there is no warning message when running dagster.


blarghmatey:
:+1:

PR #146, opened 2021-06-21
add pipeline for micromasters tables

abeglova:
closes https://github.com/mitodl/ol-data-pipelines/issues/142

Test locally by running `pull_micromasters_data_pipeline` from dagit
This doesn't have a schedule yet-I will add one once the rest of the pipeline for generating reports is built out

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,15 @@ ignore =
     # WPS332: Forbid walrus (:=) operator - Allow for use of new language feature that can reduce code lines
     WPS332,
     # WPS348: Forbid starting lines with a `.` - ignored to allow for how Black occasionally breaks up chained statements
-    WPS348
+    WPS348,
+    # E501: Forbid line too long - ignored because the max line length is shorter than Black's max line length
+    E501,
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -20,7 +20,15 @@ ignore =
...
-    WPS348
+    WPS348,
+    # E501: Forbid line too long - ignored because the max line length is shorter than Black's max line length
+    E501,
+    # WPS226: Forbid string constant over-use - ignored because many solids have the same input definitions
+    WPS226,
+    # WPS432: Forbid magic number - ignored to allow input definitions to have defaults
+    WPS432,
+    # WPS221: Forbid high Jones Complexity - ignored to allow pypika SQL queries
+    WPS221
```
blarghmatey, line 31: I added some comments to the other PR around the flake8 config changes.

blarghmatey:
I think that before we merge this and start adding all of these new datasets to S3, it's worth writing the integration with the Glue data catalog to handle the registration of the tables and their schema. Otherwise it's a manual and error-prone process.

Issue #233, opened 2022-05-20
Refactor how Dagster and user pipelines are built and deployed


blarghmatey:
# User Story
- As a data platform owner I would like to easily build and deploy new releases of the core Dagster services for patches and upgrades
- As a data engineer I would like to easily build and deploy new code and pipelines to a running Dagster instance and have visibility of when my changes are running

# Description/Context
We currently have a script that uses a project called [vdist](https://github.com/dante-signal31/vdist) for building our whole Dagster project as a debian package that is installed on an EC2 box via Salt. This is functional, but very cumbersome and coarse-grained. As we scale the number of pipelines and contributors that we would like to manage, we need a more streamlined way of packaging individual or groupings of pipelines and deploying them to a running Dagster environment.

Because we are using a mono-repo approach for our data platform we would like to use [Pants](https://www.pantsbuild.org/) to bundle the pipelines up with their relevant dependencies to create more streamlined bundles. It can also easily handle loading those packaged pipelines into Docker images that we can deploy to the running Dagster environment.

Dagster has a number of supported options for running user pipelines as containers. The most relevant for our environment is to either run them with [Docker on an EC2 box](https://docs.dagster.io/deployment/guides/docker) or investigate using the [ECS integration](https://docs.dagster.io/deployment/guides/aws#deploying-in-ecs).

# Acceptance Criteria
- [x] I can package pipeline code into a Python archive for deployment using Pants
- [x] I can build Docker images that contain my pipeline code and associated dependencies for deploying to my Dagster environment
- [ ] #276
- [x] I have a Concourse pipeline that packages pipeline code and builds associated Docker images
- [ ] I have a Concourse pipeline that deploys updated images to the target environment (ECS or EC2)

PR #240, opened 2022-06-08
Update the packaging and deployment process for Dagster and bootstrap dbt project

blarghmatey:
As part of our effort to bootstrap our data platform we need to have a solid foundation to build from. This restructures the repository to be aligned with the purpose of being the central location for all data platform related code.

- Install and configure [Pants](https://www.pantsbuild.org/) as the build tool (which we use for our other mon-repos)
- Restructure the Dagster pipelines to be oriented around the core abstractions of ops, graphs, and jobs, rather than segmenting logic around specific pipelines. This will allow for us to more easily scale the definition and re-use of process logic and encapsulate business needs along the repository groupings.
- Bootstraps a dbt project with some initial configuration
- Adds build logic to allow for easily packaging Dagster repositories (collections of pipelines, sensors, and schedules) into discrete Docker images so that they can be deployed independently
- Restructure the directory/module naming to be more indicative of what the purpose is


blarghmatey:
This addresses the work needed for #233

pants.toml
```diff
@@ -0,0 +1,33 @@
...
+  'pants.backend.docker',
+  'pants.backend.docker.lint.hadolint',
+  'pants.backend.python',
+  'pants.backend.python.lint.black',
+  'pants.backend.python.lint.flake8',
+  'pants.backend.python.lint.isort',
+  'pants.backend.python.typecheck.mypy',
+  'pants.backend.shell',
+  'pants.backend.shell.lint.shellcheck',
+  'pants.backend.shell.lint.shfmt',
```
shaidar, line 13: Might be a good idea to add `bandit` to the list

.concourse/build_containers.yaml
```diff
@@ -0,0 +1,41 @@
+---
+resources:
+- name: ol-data-platform
+  type: git
+  icon: github
+  source:
+    uri: https://github.com/mitodl/ol-data-platform
+    branch: elt_refactor
```
shaidar, line 8: Not `main`?

blarghmatey: fixed

dockerfiles/orchestrate/BUILD
```diff
@@ -0,0 +1,25 @@
...
+    tags=["docker_image"],
+    image_tags=["{build_args.DAGSTER_VERSION}", "latest"],
+)
+
+docker_image(
+    name="dagster-daemon",
+    source="Dockerfile.dagit",
+    dependencies=[
+        "src/ol_orchestrate:dagit",
+        "src/ol_orchestrate:dagster-daemon",
```
shaidar, line 19: So the `dagster-daemon` docker image has a dependency on the `dagster-daemon` orchestration?

pants.toml
```diff
@@ -0,0 +1,34 @@
...
+[flake8]
+config = 'setup.cfg'
+
+[anonymous-telemetry]
+enabled = true
+repo_id = '910e8451-aeea-4389-aae8-dde8674cace2'
+
+[docker]
+build_args=["DAGSTER_VERSION"]
+default_repository = "mitodl/data-platform-{name}"
```
shaidar, line 33: What's `name` and isn't it `mitodl/ol-data-platform`?

src/ol_dbt/models/example/my_first_dbt_model.sql
```diff
@@ -0,0 +1,26 @@
+/*
```
shaidar, line 1: Not sure we need to commit any of the example models

src/ol_orchestrate/dagster.yaml
```diff
@@ -0,0 +1,54 @@
...
+  config:
+    bucket: dagster-{{ environment }}
+    prefix: compute-logs/
+run_storage:
+  module: dagster_postgres.run_storage
+  class: PostgresRunStorage
+  config:
+    postgres_db:
+      username:
+        env: DAGSTER_PG_USERNAME
```
shaidar, line 19: Don't see where this is set in the env

src/ol_orchestrate/dagster.yaml
```diff
@@ -0,0 +1,54 @@
...
+telemetry:
+  enabled: true
+scheduler:
+  module: dagster.core.scheduler
+  class: DagsterDaemonScheduler
+compute_logs:
+  module: dagster_aws.s3.compute_log_manager
+  class: S3ComputeLogManager
+  config:
+    bucket: dagster-{{ environment }}
```
shaidar, line 11: No `ol` prefix?

src/ol_orchestrate/jobs/open_edx.py
```diff
@@ -0,0 +1,44 @@
+from dagster import graph
```
shaidar, line 1: Do we want to rename this file since we're trying to settle on edxapp and doesn't this apply only to residential?

src/ol_orchestrate/ops/elt.py
```diff
@@ -0,0 +1,11 @@
+from dagster import op
```
shaidar, line 1: Not really sure what this file is

src/ol_orchestrate/ops/open_edx.py
```diff
@@ -15,26 +15,21 @@
     OutputDefinition,
```
shaidar, line 15: Similar to above, wondering about file name

dockerfiles/orchestrate/BUILD
blarghmatey: Yeah, I was thinking I should co-locate those definitions. I'll make that change.

pants.toml
blarghmatey: Name is whatever is set as the name attribute of the `docker_image` build target in the BUILD file

src/ol_orchestrate/dagster.yaml
blarghmatey: That gets pulled in through docker-compose during deployment. Some room for improvement in the next iteration

src/ol_orchestrate/jobs/open_edx.py
blarghmatey: It's for data pertaining to an open_edx installation and this pipeline code is used for all of our deployments

src/ol_orchestrate/ops/elt.py
blarghmatey: It's a stub of the pipeline for orchestrating across Airbyte and dbt

src/ol_dbt/models/example/my_first_dbt_model.sql
blarghmatey: I'll clean that up in a subsequent PR

shaidar:
👍

PR #264, opened 2022-07-05
Set up initial dbt project structure and configuration

blarghmatey:
This sets up the dbt project with an initial set of staging models to work with the raw
data that is being synchronized via airbyte. It also configures the profiles to reduce
toil for individual users to experiment with dbt from their local machines.


dseaton:
Overall structure looks good, but I can't test locally until I have starburst access. A few questions/comments before testing locally:

1. For all six sql files, should we follow the [dbt style guide](https://docs.getdbt.com/guides/best-practices/how-we-structure/2-staging#staging-models) by using CTEs for our source and transformations? I lean toward the dbt style for improved readability, debugging, and onboarding.
2. Should we select username and email in both *users_user.sql files so that we have it for enrollment mart? Were they left off for privacy concerns?
3. In redash, the created_on field for both mitxonline and xpro is in a dict/tuple/unknown (?) format. As a cleaning step in the staging phase, should we go ahead and select member0 so that we have the timestamp? Additional context: 
    - Example result from one xPro enrollment: `{member0=2019-06-27 16:48:06.912, member1=null}`
    - [Link to redash query](https://bi.odl.mit.edu/queries/1148/source).

Note, I think all three of these comments could be addressed in a follow up PR.

src/ol_dbt/README.md
```diff
@@ -1,11 +1,13 @@
...
-Try running the following commands:
-- dbt run
-- dbt test
+This project is configured to use Trino as the warehouse engine. The profile information
+is defined in the repository with environment variables for the username and
+password. There are separate profiles defined for QA and Production environments.
 
+To run a dbt build you can use the command:
+```
+DBT_TRINO_USERNAME=<your_username> DBT_TRINO_PASSWORD=<your_password> dbt build --project-dir src/ol_dbt/ --profiles-dir src/ol_dbt/ --target <qa|production>
```
shaidar, line 9: Maybe adding the `-x` (fail-fast flag) and the `--no-anonymous-usage-stats` are worth it?

src/ol_dbt/models/staging/mitxpro/stg_mitxpro__app__postgres__courses_courserun.sql
```diff
@@ -0,0 +1,10 @@
+select
```
shaidar, line 1: Should those file names have a double underscore between `stg` and the `app`?

dseaton: @shaidar - so like this? `stg__mitxpro__app__postgres__courses_courserun.sql` 

When I first saw the filenames, I had similar questions. 

dbt recommends `<type/dag_stage>_<source/topic>__<additional_context>`: [link](https://docs.getdbt.com/blog/stakeholder-friendly-model-names#embed-information-in-the-name-using-an-agreed-upon-pattern). However, I haven't seen sufficiently complex examples to say what happens within "source/topic". The double underscores between `mitxpro`, `app`, `postgres`, and `courses` made it harder for me to jump straight from source/topic to context.

src/ol_dbt/dbt_project.yml
```diff
@@ -25,14 +25,20 @@ clean-targets:         # directories to be removed by `dbt clean`
...
 models:
-  ol_data:
-    # Config indicated by + and applies to all files under models/example/
-    example:
-      +materialized: view
+  open_learning:
+    staging:
+      +materialized: incremental
+    intermediate:
+      +materialized: views
```
dseaton, line 42: For intermediate, `+materialized: views` throws an error when I test intermediate SQL:

```
17:28:52  Compilation Error in model int_finance__ol_joined_enrollments (models/intermediate/int_finance__ol_joined_enrollments.sql)
17:28:52    No materialization 'tables' was found for adapter trino! (searched types 'default' and 'trino')
```

It looks like a starburst/trino issue. I tried "views" and "tables", but incremental was the only option that worked.

src/ol_dbt/models/staging/mitxpro/stg_mitxpro__app__postgres__courses_courserun.sql
blarghmatey: Good point on the source/topic issue. We _could_ use single `_` between the service/app/source_type components and then double `__` as a separator for the table name. I think for now we'll keep it as is, but something to bring into a discussion around what semantics we want to include in our controlled vocabulary for table/model names and column identifiers.

src/ol_dbt/profiles.yml
```diff
@@ -0,0 +1,24 @@
+---
+open_learning:
+  target: dev
```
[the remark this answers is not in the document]
blarghmatey: Yeah, this current iteration doesn't support a dev target but that's something we can factor in as we make this production ready.

src/ol_dbt/dbt_project.yml
blarghmatey: Thanks for figuring that out. I'll update to incremental for now and we can dig into it further as a follow up.

shaidar:
One file is `.yml` and another one is `.yaml`

blarghmatey:
> One file is `.yml` and another one is `.yaml`

Just renamed it to be consistent

src/ol_dbt/profiles.yml
```diff
@@ -0,0 +1,24 @@
...
+      type: trino
+      method: ldap
+      user: '{{ env_var("DBT_TRINO_USERNAME") }}'
+      password: '{{ env_var("DBT_TRINO_PASSWORD") }}'
+      database: ol_data_lake_qa
+      port: 443
+      schema: ol_warehouse_data_qa
+      host: mitol-ol-data-lake-qa.trino.galaxy.starburst.io
+      http_scheme: https
+    production:
```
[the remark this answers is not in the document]
blarghmatey: It is a different target, it's saying that `dev` is the default target but we don't actually have a definition for that target.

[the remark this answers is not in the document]
shaidar: So if someone runs a `dbt build` with this profile, wouldn't run against both QA and Production since they're both under the default dev target?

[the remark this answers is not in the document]
shaidar: Nevermind. I got it

PR #291, opened 2022-07-29
Create an initial OL enrollment table using dbt

dseaton:
Initial implementation of an OL enrollment table/product. The PR establishes organizational standards for intermediate and mart SQL, noting, intermediate maintains sub directories for our products (e.g., mitxonline). I also used this PR to go through the process of adding new fields to the overall enrollment table, which required starting with updates to staging SQL, then work up to the "core" mart level.

Two considerations for reviewers of this PR:
* Organization/layout of the "intermediate", "core" directories and the contained SQL.
* The process of adding new fields, e.g., username and email in this PR. This felt cumbersome at times, i.e., many updates in many files to add two fields.


src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
```diff
@@ -0,0 +1,29 @@
+---
+version: 2
+
+models:
+- name: int_mitxonline__enrollments
+  description: Intermediate model of enrollments in MITx Online.
+  columns:
```
Ferdi, line 7: It would be great to have the course_id here (e.g. course-v1:MITx+14.740x) . it's a convention used across teams and products.

dseaton: Agreed. I wanted to get this work reviewable before investing in adding/refining fields for the final enrollment product. Adding fields for the first time is a bit cumbersome, so I wanted to make sure my staging, intermediate, and mart organization style is in good shape before refining all the fields.

src/ol_dbt/models/marts/core/_core__models.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: updated_on
+    description: timestamp, specifying when the enrollment was last updated
+  - name: courseware_url_path
+    description: str, url for course where the user is enrolled
+  - name: course_title
+    description: str, title of the course where the user is enrolled
+  - name: user_active
+    description: boolean, specifying whether the user is active within a platform
+      (e.g., xPro, MITx Online, etc)
+  - name: org
```
blarghmatey, line 27: I think `org` is a bit misleading as a specifier for this information. I think `service`, `product`, `application`, `platform`, etc. is more indicative of what this is signifying. While it's true that some of these applications map to a given business unit in the department, the first-order reference is more aligned with the system that the information is coming from. If we want to tie things into a given business unit then we can build a mapping table for OU -> service in the core mart.

blarghmatey (commented):
This structure looks good. I think it would be worthwhile to add a README in the `src/ol_dbt` directory that specifies the guidelines and reasoning behind what transformations should be done at what stages so that other developers understand how to structure things.

src/ol_dbt/models/staging/mitxonline/_mitxonline__sources.yml
```diff
@@ -27,6 +27,8 @@ sources:
   - name: mitxonline__app__postgres__users_user
     columns:
     - name: id
+    - name: username
+    - name: email
```
shaidar, line 31: I'm always hesitant about including usernames and emails unless extremely necessary. So, would be interested in hearing the reason behind this requirement.

src/ol_dbt/models/marts/core/_core__models.yml
dseaton: I am going to change this to "platform". But we should continue discussing in the near future, while this is straightforward to change.

src/ol_dbt/models/staging/mitxonline/_mitxonline__sources.yml
dseaton: Thanks for this question. I took quite a bit of time thinking through whether it was appropriate. It came down to whether I could envision a product that would "not" need these or at least emails. An overall ol_enrollment report intended for internal business units needs these for most scenarios. When we start talking about other products that are facing more diverse audiences internally (possibly externally?) we have to be careful. 

From a dbt organization perspective, including username and email in the staging SQL gives a developer the option to use those fields in future reporting, but it doesn't mean all reports/products will have those fields. Right now, I am allowing email/username to flow into the marts -> core ol_enrollment product, but I think this is ok for us to test and discuss.

Does that help?

dseaton:
@shaidar @abeglova I am new to sqlfluff, but have a check failing. Unfortunately, the output from the checks is truncated and I am still trying to figure out how to get sqlfluff working correctly locally. Are either of you concerned about that failed check above? Or do you know a way for me to see the full output for the failed check?

shaidar:
It's mainly just a formatting issue:

src/ol_dbt/models/marts/core/_core__models.yml
dseaton: I also just added a minimal READ.me in the ol_dbt directory. Will iterate on this as the project evolves.

PR #307, opened 2022-08-09
Move documentation to yml files.

dseaton:
Realized I should be writing my documentation in the *.yml files, not in the SQL itself. Using the *.yml should let us create a site, generate docs, etc. more easily.

@blarghmatey Mind taking a look?


blarghmatey (approved):
:+1:

PR #309, opened 2022-08-19
Dseaton/bootcamps to ol enrollment

dseaton:
This PR adds Bootcamps enrollments to our Open Learning wide Enrollment Product. Additional contexgt:
 - @blarghmatey added all data for a few or our data sources, meaning Bootcamps was available to add.
 - We have use the `raw__` prefix as a new naming convention for source tables. 
 - I relied heavily on the [dbt-codegen package](https://github.com/dbt-labs/dbt-codegen) to generate .yml (source) and .sql (table) files.
 - I am leaving description documentation for another PR.


dseaton:
I have moved the dbt-codegen notes from the comments section of this PR to a `docs` folder in the ol-dbt project.

src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
```diff
@@ -0,0 +1,29 @@
+---
+version: 2
+
+models:
+- name: int_bootcamps__enrollments
```
blarghmatey, line 5: I'm wondering if we want to stick with 2 underscores (`__`) for all separators? The idea being that it serves as a visual indicator of the different pieces of information. The tables each have single underscores (`_`) in their name to indicate they are a unit of information. This is another thing that is useful to figure out early and make sure that we enforce for consistency.

blarghmatey: Specifically, changing `int_bootcamps__` -> `int__bootcamps__`

blarghmatey (commented):
In general this looks good. One thing that is worth considering (likely for a subsequent PR) is to refactor the enrollments model to be an aggregate across the more foundational/atomic concepts that we are working with. The overall idea being that we have those atomic concepts as models in the `core` mart and then build higher order concepts from those atoms. In the case of `enrollments` those would likely include (non-exhaustive list) `learner`, `course`, `course_run`, etc.

src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
dseaton: We do want to use double underscores between stg, int, mart and the data source. I noticed this issue when working through this PR, it also occurs for all the staging sql.  As a reviewer, if you are ok with that change being added to this PR, happy to make that now.

blarghmatey: Yes, I think it's worth including in this PR

dseaton: > Yes, I think it's worth including in this PR

Done!

dseaton:
> In general this looks good. One thing that is worth considering (likely for a subsequent PR) is to refactor the enrollments model to be an aggregate across the more foundational/atomic concepts that we are working with. The overall idea being that we have those atomic concepts as models in the core mart and then build higher order concepts from those atoms. In the case of enrollments those would likely include (non-exhaustive list) learner, course, course_run, etc.

I'd like to spin this work up as separate PR(s) for two reasons: 1) an additional product/goal would help figure out how core will work with other reporting, 2) I still have some questions as to how intermediate and core will work together. Right now, we have join logic for enrollments in intermediate (per dbt guidelines), if I end up doing these kinds of joins to pull together  those core concepts, I really don't need to do that in the intermediate layer. In any event, thinking through this in another PR might be a good strategy.

blarghmatey:
:+1:

PR #332, opened 2022-09-26
mitxonline user models

abeglova:
## Description
This adds staging and intermediate models for mitx user data 

## Motivation and Context
https://github.com/mitodl/hq/issues/266

## How Has This Been Tested?
I spot checked the data. I also have automated tests to check that the int__mitxonline__users table 1) Has the same number of rows as stg__mitxpro__app__postgres__users_user 2) Has unique ids 3) Has no missing ids to make sure that no extra rows are accidentally dropped or created during data normalization

## Types of changes

## Checklist:


src/ol_dbt/models/intermediate/combined/int__combined__users.sql
```diff
@@ -1,5 +1,9 @@
 with mitxonline_users as (
-    select * from {{ ref('int__mitxonline__users') }}
+    select
+        id
+        , username
+        , user_email as email
```
blarghmatey, line 5: What's the reason for recasting this as `email`? We should aim to keep column names and semantics consistent across layers.

src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
```diff
@@ -41,11 +41,36 @@ models:
     description: ""
 
 - name: int__mitxonline__users
-  description: ""
+  description: denormalized mitxonline users
   columns:
   - name: id
```
blarghmatey, line 46: What's the purpose of maintaining the database-internal ID column after denormalizing? At this layer we should be identifying other join keys that are semantically meaningful at the domain level, rather than relying on specifics of the source database. Once we merge source tables into a denormalized structure it becomes unclear and arbitrary which `ID` to use here.

src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user.sql
```diff
@@ -1,18 +1,23 @@
...
     select
         id
         , username
-        , email
-        , is_active
-        , created_on
-        , updated_on
+        , name as full_name
+        , email as user_email
+        , created_on as user_joined_on_utc
```
blarghmatey, line 14: For any timestamp columns we should cast them all to a standard representation (e.g. ISO8601) so that we don't have to wonder what the structure is in subsequent layers.

src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user_groups.sql
```diff
@@ -0,0 +1,16 @@
...
+
+with source as (
+    select * from {{ source('ol_warehouse_raw_data', 'raw__mitxonline__app__postgres__users_user_groups') }}
+)
+
+, cleaned as (
+
+    select
+        id
+        , group_id
```
blarghmatey, line 11: nit: What kind of group is this? This might be something we move to the intermediate layer to formalize the naming, but without the application context this identifier is very vague.

blarghmatey (commented):
Looking good so far. Just some comments/questions as we figure out our vocabularies.

src/ol_dbt/models/intermediate/combined/int__combined__users.sql
abeglova: I recast it so the union works (Ferdi asked for email to be user_email in the mitxonline data). I haven't worked on importing full data from other sources yet. I think we can just remove this whole file. It's just a placeholder

src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
abeglova: At this layer the id is useful for ensuring we don't accidentally add or drop columns during denormalization (really easy to do). We can drop it in the next layer

src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user.sql
abeglova: So, the timestamps are already ISO8601 strings but I will change this to
```   , to_iso8601(from_iso8601_timestamp(created_on)) as user_joined_on
        , to_iso8601(from_iso8601_timestamp(last_login)) as last_login
```
Just so that the format is definitely 100% standerdized and also anything that isn't an iso8061 timestamp throws an error

abeglova:
Ferdi and Peter asked me to break this into smaller pieces so closing it for now

Issue #334, opened 2022-09-26
Stage mitxonline users_user


pdpinch:
- [ ] what are the raw fields
- [ ] write up a plan for each field, including modifications to the field name, data types, data to omit, etc. 
- [ ] open a PR the implements the plan for `users_user`

Ferdi:
```
id
username
email
created_on_utc
updated_on_utc
is_active
last_login_utc
```

We are dropping `is_staff` , `is_supperuser` since they are simply used by django [internally](https://github.com/mitodl/hq/issues/266#issuecomment-1258253936) to specify who is [allowed](https://stackoverflow.com/a/20470185) to view certain pages. 

I'm not sure we need to prefix fields with `user_` .   I think the table/model name will serve as a namespace.

abeglova:
I don't think the `updated_on` is useful - it's just a field that django automatically updates. Should be just after last_login most of the time, whenever the script to update the last_login actually runs

abeglova:
We can just filter out users where is_active=False. That flag is used to delete accounts.

Ferdi:
> We can just filter out users where is_active=False. That flag is used to delete accounts.

I think we should keep them around.  The might be needed in lookups (e.g customer service) , they might have enrollments in the past etc.

abeglova:
Ok I think the plan should be to create a separate table for users where is_active = False in the staging layer so those users aren't accidentally included

PR #343, opened 2022-09-27
users_user staging

abeglova:
## Description
This pr creates the staging model for users user
The users_user table has the following columns

id - int postgres id for the table, keeping it as is
name - varchar renamed to full_name since user_legaladdress has first_name and last name
email - varchar keeping as is
is_staff - boolean dropping as per the discussion 
password - varchar dropping
username - varchar keeping
is_active - boolean flag used to "delete" users. Keeping as is for now. We should separate users with is_active=False into a separate table in the intermediate layer 
created_on- iso8601 string renamed to joined_on. Postgres sets this automatically when a row is created in the users table
last_login -  iso8601 string keeping
updated_on - iso8601 string, this is a field updated by postgres when the row is updated, Dropping, since it mostly tracks app internals. It should be just after last_login most of the time, whenever the script that updates that field actually runs
is_superuser - boolean dropping as per the discussion

## Motivation and Context
https://github.com/mitodl/ol-data-platform/issues/334

## How Has This Been Tested?
Tested locally

## Types of changes

## Checklist:

### DIFF
--- a/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
+++ b/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
@@ -1,0 +1,31 @@
+---
+version: 2
+
+models:
+- name: stg__mitxonline__app__postgres__users_user
+  columns:
+  - name: id
+    description: int, sequential ID representing one user in MITx Online
+    tests:
+    - unique
+    - not_null
+  - name: username
+    description: str, name chosen by user
+    tests:
+    - unique
+    - not_null
+  - name: email
+    description: str, user email associated with their account
+    tests:
+    - unique
+    - not_null
+  - name: is_active
+    description: boolean, used to soft delete user accunts
+  - name: full_name
+    description: str, the user's full name
+    tests:
+    - not_null
+  - name: created_on
+    description: timestamp, specifying when a user account was initially created
+  - name: last_login
+    description: timestamp, specifying when a user last logged in
--- a/src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user.sql
+++ b/src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user.sql
@@ -5,13 +5,15 @@ with source as (
 )
 
 , cleaned as (
+
     select
         id
         , username
         , email
         , is_active
-        , created_on
-        , updated_on
+        , name as full_name
+        , to_iso8601(from_iso8601_timestamp(created_on)) as joined_on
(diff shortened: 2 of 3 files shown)


src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__enrollments.sql
```diff
@@ -17,7 +17,7 @@ with enrollments as (
     select
         id
         , username
-        , email
+        , user_email as email
```
abeglova, line 20: This will be substantially changed soon. This is a temporary change to avoid braking the combined enrollments view

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
```diff
@@ -2,7 +2,7 @@ with users as (
     select
         id
         , username
-        , email
+        , user_email as email
```
abeglova, line 5: This is a temporary change to avoid breaking the current intermediate views, which will be replaced

src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__users_user.sql
```diff
@@ -1,18 +1,20 @@
...
-        , email
-        , is_active
-        , created_on
-        , updated_on
+        , name as full_name
+        , email as user_email
+        , to_iso8601(from_iso8601_timestamp(created_on)) as created_on
+        , to_iso8601(from_iso8601_timestamp(last_login)) as last_login
     from source
+    where is_active = true
```
abeglova, line 17: We set is_active=False instead of deleting them. The email is scrambled when that happens. We can ignore records with active=Fasle

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
```diff
@@ -0,0 +1,29 @@
...
+    - not_null
+  - name: email
+    description: str, user email associated with their account
+    tests:
+    - unique
+    - not_null
+  - name: is_active
+    description: boolean, used to soft delete user accunts
+  - name: full_name
+    description: str, the user's full name
```
rachellougee, line 25: The same comment as legaladdress PR, do we need to test full name as not_null?

rachellougee (approved):
Looks good to me

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
abeglova: I'll add the constraint

PR #344, opened 2022-09-27
Create staging_guidelines.md

dseaton:
Proposed guidelines associated with the "staging" process.

## Motivation and Context
These guidelines are intended to help reduce ambiguity as we stage data in our dbt project. The guidelines come from Chapters 10 and 11 of the "Informed Company", dbt documentation, and a desire to support new people that will pick up data staging work.

### DIFF
--- a/src/ol_dbt/docs/staging_guidelines.md
+++ b/src/ol_dbt/docs/staging_guidelines.md
@@ -18,13 +18,9 @@ Decisions on which tables and fields to "stage" should be determined by the issu
 - PII (Personally Identifying Information): _we want to ask ourselves internally, are these data worth risking PII exposure?_ If concerned, ask for clarifications.
 
 ### Renaming fields
-Naming convention guidelines will be highlighted here, but ambiguity should be addressed in the discussion of the issue being worked on. Following [Emily Riederer's "field names as contracts"](https://emilyriederer.netlify.app/post/column-name-contracts/), she establishes a `<type>_<entity>_<column_name/details>`. Fields in the MITx Online users_user table would become:
-- id: `id_user`
-- username: `txt_user_username`
-- email: `txt_user_email`
-- is_active: `is_user_active`
-- created_on: `tm_user_created_on`
-- updated_on: `tm_user_updated_on`
+Naming conventions, and any related ambiguity, should be addressed in the discussion of the issue being worked on. 
+
+_A discussion post will be created for whether we can establish our own version of [Emily Riederer's "field names as contracts"](https://emilyriederer.netlify.app/post/column-name-contracts/), using `<type>_<entity>_<column_name/details>`_. 
 
 ### Transforming timestamps
 All timestamp fields should be converted to [ISO8601](https://en.wikipedia.org/wiki/ISO_8601). For example, 2022-09-26T18:51:23Z.
@@ -35,9 +31,9 @@ In some situations, we may want to tranform field values to specific types or to
 - Ints, floats, strings to their proper format where needed. We can likely leverage the dbt testing framework to help with this (as oppossed to making it a manual transformation).
 
 ### Testing with dbt
-dbt has a [robust testing framework](https://docs.getdbt.com/docs/building-a-dbt-project/tests) that includes four
+dbt has a [testing framework](https://docs.getdbt.com/docs/building-a-dbt-project/tests) that includes four
 out of the box tests: `uniqueness`, `not_null`, `accepted_values`, and `relationships`. Guidelines:
-- Include at least one test per table, focusing particularly on "primary keys" and the outcomes of transformed data.
+- Include testing where appropriate, espcially in cases where staged fields play an important role in intermediate and mart layers.
 
 ### Documentation
 Raw tables, staging models, and staging fields can all be documented directly in the `.yml` files maintained in directories containing our model `.sql`. Guidelines:
--- a/src/ol_dbt/docs/staging_guidelines.md
+++ b/src/ol_dbt/docs/staging_guidelines.md
@@ -12,7 +12,7 @@ The following considerations are all necssary phases of the "staging" process in
 The guidelines below are an attempt to help prepare contributers on what to consider when they begin "staging" data.
 
 ### Selecting tables and fields to stage
-Decisions on which tables and fields to "stage" should be determined by the issue being addressed. At times, table and field selections will be based upon downstream use, and at other times, part of building out our initial project structure. Guidelines: 
+Decisions on which tables and fields to "stage" should be determined by the issue being addressed. At times, table and field selections will be based upon downstream use, and at other times, part of building out our initial project structure. Guidelines:
 - Seek clarity and ask questions via the issue being worked on.
 - Explore data to help determine which tables are needed ([Redash](https://bi.odl.mit.edu/queries?order=-created_at&page=1&page_size=20) is a great tool for exploring tables).
 - PII (Personally Identifying Information): _we want to ask ourselves internally, are these data worth risking PII exposure?_ If concerned, ask for clarifications.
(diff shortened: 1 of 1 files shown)


src/ol_dbt/docs/staging_guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+### Selecting tables and fields to stage
+Decisions on which tables and fields to "stage" should be determined by the issue being addressed. At times, table and field selections will be based upon downstream use, and at other times, part of building out our initial project structure. Guidelines:
+- Seek clarity and ask questions via the issue being worked on.
+- Explore data to help determine which tables are needed ([Redash](https://bi.odl.mit.edu/queries?order=-created_at&page=1&page_size=20) is a great tool for exploring tables).
+- PII (Personally Identifying Information): _we want to ask ourselves internally, are these data worth risking PII exposure?_ If concerned, ask for clarifications.
+
+### Renaming fields
+Naming convention guidelines will be highlighted here, but ambiguity should be addressed in the discussion of the issue being worked on. Following [Emily Riederer's "field names as contracts"](https://emilyriederer.netlify.app/post/column-name-contracts/), she establishes a `<type>_<entity>_<column_name/details>`. Fields in the MITx Online users_user table would become:
+- id: `id_user`
+- username: `txt_user_username`
```
abeglova, line 23: I don't think this is a good idea. More verbose column names are harder to remember, tyoe and work with. I think if there is ambiguity we can add table prefixes in the intermediate layer. It is is pretty easy to check the type in the database and the only situations where ambiguous types really cause issues in sql queries is surprise nulls and errors in converting between types (ei converting between string timestamps and dates). Also all the dates are stored as iso string anyway. I do like the is_ prefix for booleans

src/ol_dbt/docs/staging_guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+### Transforming timestamps
+All timestamp fields should be converted to [ISO8601](https://en.wikipedia.org/wiki/ISO_8601). For example, 2022-09-26T18:51:23Z.
+
+### Transforming values
+In some situations, we may want to tranform field values to specific types or to more human interpretable formats. Examples might include:
+- Writing "CASE" logic to combine multiple boolean fields into a signle "string" field. Chapter 10 of the [Informed Company](https://learning.oreilly.com/library/view/the-informed-company/9781119748007/c10.xhtml#head-2-1) details some of these types of cases.
+- Ints, floats, strings to their proper format where needed. We can likely leverage the dbt testing framework to help with this (as oppossed to making it a manual transformation).
+
+### Testing with dbt
+dbt has a [robust testing framework](https://docs.getdbt.com/docs/building-a-dbt-project/tests) that includes four
```
abeglova, line 38: I don't see the point of this either. The transformations we do in the staging layer generally won't introduce the kinds of errors that can be caught by dbt tests. We can check that all the primary keys are  unique and present but if they are not the problem is either in the original data or in copying the raw files. I agree that we should have tests to check that we aren't accidentally dropping or creating rows in the intermediate and mart layers

abeglova (changes requested):

src/ol_dbt/docs/staging_guidelines.md
dseaton: For the task at hand (bringing over postgres tables), your logic is spot on. I don't know about cases with more complex data (e.g., log events), or data that we don't directly own (e.g., IRx data), but we aren't there yet. A good compromise is:

"Test any field value transformations you make". So if you change a couple boolean fields to a single categorical field, we would write a test for that.

src/ol_dbt/docs/staging_guidelines.md
dseaton: > add table prefixes in the intermediate layer

dbt guidelines: [Rename and Cast Fields Once (in the staging layer)](https://docs.getdbt.com/guides/legacy/best-practices#rename-and-recast-fields-once).

> More verbose column names are harder to remember, type and work with.

My take on Emily's recommendations are: 1) interpretability by analysts, data scientists, consumers and 2) settling on a reproducible convention so we don't have to ask each other about naming practices (especially when multiple people are working on a single dbt project). 

It might seem like overhead to create <type>_<entity>_<col_name/details>, but it will save a lot of overhead in other places. For example, by leaving "staging" models with the original names, it means that an analyst working in the intermediate layer has to remember 1) the original names and 2) the naming conventions.

abeglova: I think adding the table name to the prefix will make it more confusing not less down the line. Once you are joining data from  multiple databases it's not clear which "user" table you are talking about and preserving the database structure of the original source in the name makes things more confusing not less. For example, once I'm working with intermediate data I don't need to remember that the mitxonline postgress database stores first name in the user_legaladdress table and not users_user

dseaton: > I think adding the table name to the prefix will make it more confusing not less down the line.

We will have to get deeper in the work, or create a couple examples in Redash, in order to test out Emily's codification. It's not meant to be just a table name.

> As for types - Maybe we actually add dbt tests to ensure that the types are what we expect them to be? That sounds a lot easier to work with than adding prefixes

I am less certain on "types", but predict it will serve everyone well in the future. Especially when we do more complex staging work and when we get to marts that require specific aggregation (other measures).

abeglova: I don't think type mismatches cause problems in sql queries very often and more verbose and less readable names do add to confusion

abeglova: And the kind of problems that are caused by something being not the type you expect in a sql query would be caught immediately anyway

abeglova: I'm just worried that column names will work similarly to the prefixes for table names which are descriptive but also pretty much make it impossible to get the correct table name without copy pasting

dseaton: I am going to put together a couple more examples to demonstrate how this would work. I have realized that I misinterpreted part of Emily's suggestions with `txt`, which may have led to more confusion. Stay tuned.

pdpinch: Anastasia has done a lot of work with writing reports in redash, and I expect she'll be doing more. I trust her experience with column names. 

And we can change it later if we have to, right?

dseaton: @pdpinch I added a contextual statement to the PR. [See above](https://github.com/mitodl/ol-data-platform/pull/344#issue-1387828038).

src/ol_dbt/docs/staging_guidelines.md
abeglova: So I thought about this and I changed my mind. We shouldn't write tests for the sake of writing tests but it's important to documents unique and non-null constraints from the original database somewhere in ol-data-platform and dbt tests seems like an easy way to do it. 

For example in the user prs I'm working on,  it's critical information that  the mitxonline enforces that `email` and `username` are  unique for users and the user_legaladdress table is one to one with users (so user_id in that table is unique). We can document that information by adding tests for it.

dseaton: I like it. Will update language and use your email/username example.

src/ol_dbt/docs/staging_guidelines.md
dseaton: @abeglova This should be all set. My updates say "discuss naming conventions in the issue".

src/ol_dbt/docs/staging_guidelines.md
dseaton: @abeglova This should be all set. State that we should test fields that will be used in intermediate/mart layers.

abeglova (approved):
LGTM

Issue #351, opened 2022-09-28
Stage xpro users_user


abeglova:
- [ ] what are the raw fields
- [ ] write up a plan for each field, including modifications to the field name, data types, data to omit, etc. 
- [ ] open a PR the implements the plan for `users_user`

abeglova:
The table has the following fields:
The users_user table has the following columns

id - int postgres id for the table, keeping it as is
name - varchar renamed to full_name since user_legaladdress has first_name and last name
email - varchar keeping as is
is_staff - boolean dropping as per the discussion we had for is_staff for mitxonline
password - varchar dropping
username - varchar keeping
is_active - boolean flag used to "delete" users. Keeping as is for now. We should separate users with is_active=False into a separate table in the intermediate layer
created_on- iso8601 string renamed to joined_on. Postgres sets this automatically when a row is created in the users table
last_login - iso8601 string keeping
updated_on - iso8601 string, this is a field updated by postgres when the row is updated, Dropping, since it mostly tracks app internals. It should be just after last_login most of the time, whenever the script that updates that field actually runs
is_superuser - boolean dropping as per the discussion we had for is_superuser for mitxonline

PR #352, opened 2022-09-28
add xpro users_user staging

abeglova:
## Description
This pr creates the staging model for users user
The users_user table has the following columns

id - int postgres id for the table, keeping it as is
name - varchar renamed to full_name since user_legaladdress has first_name and last name
email - varchar keeping as is
is_staff - boolean dropping as per the discussion 
password - varchar dropping
username - varchar keeping
is_active - boolean flag used to "delete" users. Keeping as is for now. We should separate users with is_active=False into a separate table in the intermediate layer 
created_on- iso8601 string renamed to joined_on. Postgres sets this automatically when a row is created in the users table
last_login -  iso8601 string keeping
updated_on - iso8601 string, this is a field updated by postgres when the row is updated, Dropping, since it mostly tracks app internals. It should be just after last_login most of the time, whenever the script that updates that field actually runs
is_superuser - boolean dropping as per the discussion

## Motivation and Context
https://github.com/mitodl/ol-data-platform/issues/351

## How Has This Been Tested?
Tested locally

## Types of changes

## Checklist:

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
@@ -62,7 +62,7 @@ models:
     tests:
     - unique
     - not_null
-  - name: user_joined_on
+  - name: joined_on
     description: timestamp, user join timestamp
   - name: last_login
     description: timestamp, user last log in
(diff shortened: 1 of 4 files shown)


rachellougee:
@abeglova There is a failing test related to "Env var required but not provided: 'DBT_TRINO_USERNAME'"? Did the profile have DBT_TRINO_USERNAME?

dseaton (approved):
LGTM.

Issue #232, opened 2022-05-19
Standardize ELT steps as a Dagster pipeline


blarghmatey:
A large fraction of our pipelines will be a standard pattern of:
- [ ] Trigger an [Airbyte](https://airbyte.com/) sync for a given connection
- [ ] Ensure that the [AWS Glue catalog](https://docs.aws.amazon.com/prescriptive-guidance/latest/serverless-etl-aws-glue/aws-glue-data-catalog.html) is up to date with the replicated schema of the raw data
- [ ] Trigger a [dbt](https://www.getdbt.com/) job to transform the replicated data

We need to have modularized [`ops`](https://docs.dagster.io/concepts/ops-jobs-graphs/ops#ops) that capture each of the above steps so that we can parameterize them easily as a [`job`](https://docs.dagster.io/concepts/ops-jobs-graphs/jobs-graphs#jobs--graphs) that accepts a definition of which connection to sync and which dbt job to run.

There are existing plugins for Dagster for interfacing with [Airbyte](https://docs.dagster.io/_apidocs/libraries/dagster-airbyte) and [dbt](https://docs.dagster.io/_apidocs/libraries/dagster-dbt). We have also created an initial [library](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_data_pipelines/lib/glue_helper.py) for interfacing with glue, though it will likely need some updates.

shaidar:
Wrote the code to handle basic creation of AWS Glue crawler using the API. Ran into an issue trying to communicate with the Airbyte API. Ended up torching my docker setup and rerunning it which seemed to work so am now writing the code to trigger airbyte sync.

shaidar:
Initially went down the wrong path of writing python code from scratch to handle crawler creation and started working on the Airbyte bit as I assumed that we would be calling a python script through Dagster. Then I realized there are integration in Dagster to handle airbyte. Setup a test db, S3 bucket, local airbyte, and local dagster and went through some of the dagster tutorials to get familiar with the tool. Ran an airbyte sync which seems to work as expected as long as the passed connection ID is correct.

shaidar:
Ran into some roadblocks and trying to figure out how to work through them in Dagster. Conceptually feels like dagster is very similar to to Concourse and unfortunately their documentation is also as sparse. 
[...]

PR #356, opened 2022-09-29
Asset materialization

shaidar:
## Description

Addresses [Issue#232](https://github.com/mitodl/ol-data-platform/issues/232)
## Motivation and Context

[Issue#232](https://github.com/mitodl/ol-data-platform/issues/232)
## How Has This Been Tested?

Ran locally with a local Airbyte instance running.
## Types of changes

## Checklist:


src/ol_orchestrate/repositories/asset_materialization.py
```diff
@@ -0,0 +1,138 @@
...
+
+from ol_orchestrate.jobs.sync_assets_and_run_models import sync_assets_and_run_models
+
+airbyte_resource = airbyte_resource.configured(
+    {"host": "airbyte-qa.odl.mit.edu", "port": "443"}
+)
+
+
+@repository
+def bootcamps_qa_app_db_to_datalake_raw():
```
blarghmatey, line 13: We should name the repositories according to the service that it is managing data for. So, we would have a repository for bootcamps, one for xpro, one for MITx Online, etc. Within each repository we'll have a collection of pipelines, where e.g. MITx Online will have "stage_application_database_tables" and "stage_openedx_database_tables", etc.

We also want to abstract the connection IDs to pass the proper value based on environment. As an example of that, we can use a dictionary that keys off of an environment variable to pass in the proper connection ID (e.g. https://docs.dagster.io/guides/dagster/transitioning-data-pipelines-from-development-to-production#part-two-deployment)

blarghmatey:
This is looking good so far. The other piece this needs is an update to the BUILD file to package the different repositories (e.g. https://github.com/mitodl/ol-data-pipelines/blob/main/src/ol_orchestrate/BUILD#L8-L18) and adding the appropriate entries in the workspace.yaml (https://github.com/mitodl/ol-data-pipelines/blob/main/src/ol_orchestrate/workspace.yaml#L3-L6). The docker-compose file is currently defined in ol-infrastructure [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/dagster/templates/docker-compose.yaml.j2#L1)

This is an area that we should work on improving as a follow-on task.

src/ol_orchestrate/repositories/asset_materialization.py
shaidar: @blarghmatey Just to make sure I'm on the right path of what u have in mind regarding repo grouping, u are looking at something similar to this?
```
@repository
def mitxonline():
    return [
        sync_assets_and_run_models.to_job(
            name="raw_mitxonline_application_database_tables",
            resource_defs={
                "airbyte": airbyte_resource,
                "dbt": dbt_cli_resource.configured(
                    {
                        "project_dir": "/opt/dbt/models/staging/mitxonline",
                        "profiles_dir": "/opt/dbt",
                    }
                ),
            },
            config={
                "ops": {
                    "sync_airbyte": {
                        "config": {
                            "connection_id": "6e8cdba9-c711-4118-a196-91038cee78a7"
                        }
                    }
                },
            },
        ),
        sync_assets_and_run_models.to_job(
            name="raw_mitxonline_openedx_database_tables",
            resource_defs={
                "airbyte": airbyte_resource,
                "dbt": dbt_cli_resource.configured(
                    {
                        "project_dir": "/opt/dbt/models/staging/mitxonline",
                        "profiles_dir": "/opt/dbt",
                    }
                ),
            },
            config={
                "ops": {
                    "sync_airbyte": {
                        "config": {
                            "connection_id": ""
                        }
                    }
                },
            },
        )
    ]
```

src/ol_orchestrate/BUILD
```diff
@@ -28,3 +28,27 @@ docker_image(
     tags=["data_pipeline", "docker_image"],
     image_tags=["{build_args.DAGSTER_VERSION}", "latest"],
 )
+
+python_distribution(
+    name="asset-materialization",
+    dependencies=[
+        "//:requirements#dagster-airbyte",
+        "//:requirements#dagster-dbt",
```
blarghmatey, line 36: These dependencies will be inferred by the imports in the module that defines the graph so you can leave them out.

src/ol_orchestrate/BUILD
```diff
@@ -28,3 +28,25 @@ docker_image(
...
+    provides=python_artifact(
+        name="dagster-pipeline",
+        version="0.5.2",
+    ),
+)
+
+docker_image(
+    name="asset-materialization-pipeline",
+    dependencies=[
+        "src/ol_dbt:dbt_project"
```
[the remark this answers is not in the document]
shaidar: Made the change, but could you elaborate a bit on this as I don't think I understand the reason for it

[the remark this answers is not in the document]
blarghmatey: This tells pants that, when you say `./pants build src/ol_orchestrate/:asset-materialization-pipeline` it will see that there's a dependency on the Python package, so it will build that first, and then inject the built package into the build context that it generates for the Dockerfile so it can be copied and installed into the image. https://www.pantsbuild.org/docs/docker#adding-dependencies-to-your-docker_image-targets

blarghmatey:
:+1:

PR #362, opened 2022-10-05
Ab/mitx online users intermediate

abeglova:
## Description
This adds an intermediate model to mitxonline users

## How Has This Been Tested?
There are automated tests to ensure columns that we expect to be unique or not null are that way
There is also a test to ensure that the number of rows in the intermediate model equals the number of of rows in users_user where is_active=True

## Types of changes

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ requires = ["poetry_core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.sqlfluff.core]
-templater = "jinja"
+templater = "dbt"
 dialect = "hive"
 sql_file_exts = ".sql,.sql.j2,.dml,.ddl"
 
@@ -75,3 +75,4 @@ apply_dbt_builtins = true
 
 [tool.sqlfluff.templater.dbt]
 project_dir = "src/ol_dbt/"
+profiles_dir = "src/ol_dbt/"
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
@@ -1,9 +1,23 @@
 with users as (
-    select
-        id
-        , username
-        , email
+    select *
     from {{ ref('stg__mitxonline__app__postgres__users_user') }}
 )
 
-select * from users
+, users_legaladdress as (
+    select *
+    from {{ ref('stg__mitxonline__app__postgres__users_legaladdress') }}
+)
+
+select
+    users.id
+    , users.username
+    , users.full_name
+    , users.email
+    , users.joined_on
+    , users.last_login
+    , users_legaladdress.user_address_country
+    , users_legaladdress.first_name
+    , users_legaladdress.last_name
+from users
+left join users_legaladdress on users_legaladdress.user_id = users.id
+where users.is_active = true
--- a/src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
+++ b/src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
@@ -1,0 +1,36 @@
+-- Modifies equal_rowcount from dbt_utils --
+-- https://github.com/dbt-labs/dbt-utils/blob/7eab49274943964f46a5d2ddd326fac11a16598a/macros/generic_tests/equal_rowcount.sql --
+
+{% test equal_rowcount_with_filters(model, compare_model, filter=null, compare_filter=null) %}
+
+with a as (
+
+    select count(*) as count_a from {{ model }}
+    {% if filter %}
+    where {{filter}}
+    {% endif %}
+
+),
+b as (
+
+    select count(*) as count_b from {{ compare_model }}
+    {% if compare_filter %}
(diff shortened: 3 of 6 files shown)


abeglova:
I'm not sure what's causing the pre-commit.ci error. Also doesn't look like tests are running for new prs I just opened. I don't think the issue is my pr but I switched the sql-fluff templater to "dbt" from "jinga" so it might be

pyproject.toml
```diff
@@ -58,7 +58,7 @@ requires = ["poetry_core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.sqlfluff.core]
-templater = "jinja"
+templater = "dbt"
```
abeglova, line 61: Needed to change this so the "test" tag didn't throw an error in sqlfluff

src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
```diff
@@ -0,0 +1,36 @@
...
+
+{% test equal_rowcount_with_filters(model, compare_model, filter=null, compare_filter=null) %}
+
+with a as (
+
+    select count(*) as count_a from {{ model }}
+    {% if filter %}
+    where {{filter}}
+    {% endif %}
+
```
dseaton, line 12: rm extra line?

src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
```diff
@@ -0,0 +1,36 @@
...
+),
+final as (
+
+    select
+        count_a,
+        count_b,
+        abs(count_a - count_b) as diff_count
+    from a
+    cross join b
```
rachellougee, line 30: Is there a reason it's choosing a cross join here vs a full join? there is only 1 row returning from a and b. I haven't used cross join much, just curious the choice here

src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
```diff
@@ -0,0 +1,36 @@
+-- Modifies equal_rowcount from dbt_utils --
```
dseaton, line 1: Awesome! Really like seeing how this works.

dseaton (approved):
Looks good to me. 

I asked a potentially complex question about filtering out `is_active = false`, but I don't think it should slow down this PR.

pyproject.toml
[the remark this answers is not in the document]
abeglova: I needed to change it because the "test" tag threw an unknown tag error.  But also, it makes sense to use the dbt template since this is a dbt project and there will probably be other similar issues

src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
abeglova: A full join wouldn't work because there isn't a key to match on. A full join matches on a key but returns items from both tables when there isn't a match. To use a full join here I would have to create a fake "join_id" column with the same value in both tables

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
```diff
@@ -1,9 +1,23 @@
...
+    , users.full_name
+    , users.email
+    , users.joined_on
+    , users.last_login
+    , users_legaladdress.user_address_country
+    , users_legaladdress.first_name
+    , users_legaladdress.last_name
+from users
+left join users_legaladdress on users_legaladdress.user_id = users.id
+where users.is_active = true
```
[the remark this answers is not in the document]
abeglova: I think maybe we create separate intermediate a tables with "deleted" users if that data is needed? For what it's worth a lot of deleted users looked like they were duplicates of users that were active but I didn't look too closely

src/ol_dbt/tests/generic/test_equal_rowcount_with_filters.sql
rachellougee: Ah.. I forgot that this is returning count, so no key match. I apparently didn't think too much.

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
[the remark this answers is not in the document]
abeglova: I think the only circumstance where we would care about "deleted" users is if for some reason there is a "deleted" user that made a purchase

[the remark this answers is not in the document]
dseaton: I agree with both your responses. 

We need to make sure to apply the same filters to other open edx sources (e.g., xpro). I'll see if we have that information in the edX/2U data.

blarghmatey:
It looks like the error from pre-commit is because it's defaulting to the `~/.dbt` directory for the profile, which doesn't exist. https://docs.sqlfluff.com/en/stable/configuration.html#dbt-project-configuration

abeglova:
Thanks @blarghmatey, that fixed the issue

rachellougee (approved):
Looks good to me

PR #364, opened 2022-10-06
fix precommit

abeglova:
## Description
Currently sqlfluff is broken on master - this fixes it.

## Motivation and Context
In https://github.com/mitodl/ol-data-platform/pull/362 I switched the templater from sqlfluff from `jinja` to `dbt` because the `jinja` templater wasn't able to parse the `test` keyword. It looks like the dbt templater needs credentials to connect to our database to work properly. I don't like the idea of something that we're using as a linter having those credentials so for now I switched the templater back to `jinja` and just added  src/ol_dbt/tests/  to .sqlfluffignore so it isn't parsed. 

I'm not sure if this is the right call but I want to fix  the automated tests before merging the xpro prs and going on vacation next week. I'd like to discuss this with Tobias when I'm back from vacation.


pyproject.toml
```diff
@@ -76,3 +76,5 @@ apply_dbt_builtins = true
 [tool.sqlfluff.templater.dbt]
 project_dir = "src/ol_dbt/"
 profiles_dir = "src/ol_dbt/"
+profile ="open_learning"
```
abeglova, line 79: Not needed if we use templater = jinja but I added it in case we want to switch the templater back to dbt

rachellougee (approved):
I will just approve it for now since we need to fix the automated test. I am not sure if it's a good idea to put test directories in .sqlfluffignore. We can revisit it later

abeglova:
I agree - I'm not sure if this is the best solution. Feel free to change it while I'm away next week if Tobias thinks that creating credentials for pre-commit on github is a good idea

PR #394, opened 2022-10-18
intermediate xpro users

abeglova:
This pr adds an intermediate model for xpro users. 

Also, since we decided to prefix course and program fields with course_ and program_ to avoid name collisions when joining intermediate models, I added user_ to fields from user tables.

I wasn't sure which columns to exclude for privacy reasons. I ended up including everything from users_legaladdress and  users_profile except address information that is more granular than state/region from users_legaladdress and user_birth_year from users_profile. I still don't have a good sense of what user information is too private to include and and at what point in the process we should exclude it.

We have a bunch of placeholder combined intermediate models in the codebase right now - I made the minimal changes to those so that the sql doesn't error. Those will be all completely re-written later

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/int__combined__enrollments.sql
+++ b/src/ol_dbt/models/intermediate/combined/int__combined__enrollments.sql
@@ -6,10 +6,6 @@ with mitxonline_enrollments as (
     select * from {{ ref('int__mitxpro__enrollments') }}
 )
 
-, bootcamps_enrollments as (
-    select * from {{ ref('int__bootcamps__enrollments') }}
-)
-
 , join_ol_enrollments as (
     select
         *
@@ -22,13 +18,6 @@ with mitxonline_enrollments as (
         *
         , 'xPro' as platform
     from mitxpro_enrollments
-
-    union all
-
-    select
-        *
-        , 'Bootcamps' as platform
-    from bootcamps_enrollments
 )
 
 select * from join_ol_enrollments
--- a/src/ol_dbt/models/intermediate/combined/int__combined__users.sql
+++ b/src/ol_dbt/models/intermediate/combined/int__combined__users.sql
@@ -1,18 +1,20 @@
 with mitxonline_users as (
     select
-        id
-        , username
-        , email
+        user_id
+        , user_username
+        , user_email
     from {{ ref('int__mitxonline__users') }}
 )
 
 , mitxpro_users as (
-    select * from {{ ref('int__mitxpro__users') }}
+    select
+        user_id
+        , user_username
+        , user_email
+    from {{ ref('int__mitxpro__users') }}
 )
 
-, bootcamps_users as (
-    select * from {{ ref('int__bootcamps__users') }}
-)
+
 
 , join_ol_users as (
     select
@@ -26,13 +28,6 @@ with mitxonline_users as (
         *
         , 'xPro' as platform
     from mitxpro_users
-
-    union all
-
-    select
-        *
-        , 'Bootcamps' as platform
-    from bootcamps_users
 )
 
 select * from join_ol_users
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__enrollments.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__enrollments.sql
@@ -15,9 +15,9 @@ with enrollments as (
 
 , users as (
     select
-        id
-        , username
-        , email
+        user_id
+        , user_username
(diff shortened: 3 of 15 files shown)


src/ol_dbt/models/intermediate/combined/int__combined__enrollments.sql
```diff
@@ -6,10 +6,6 @@ with mitxonline_enrollments as (
     select * from {{ ref('int__mitxpro__enrollments') }}
```
abeglova, line 6: This table is just a placeholder for now. I removed bootcamps since the column names no longer match

src/ol_dbt/models/intermediate/combined/int__combined__users.sql
```diff
@@ -1,18 +1,20 @@
 with mitxonline_users as (
     select
-        id
-        , username
-        , email
+        user_id
```
abeglova, line 3: This model is just a placeholder for now. I removed bootcamps since the column names no longer match

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__enrollments.sql
```diff
@@ -15,9 +15,9 @@ with enrollments as (
```
abeglova, line 15: This table/model is still a placeholder

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__enrollments.sql
```diff
@@ -14,9 +14,9 @@ with enrollments as (
```
abeglova, line 14: This model is still a placeholder

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__users.sql
```diff
@@ -1,9 +1,40 @@
...
+    , users.user_last_login
+    , users_legaladdress.user_address_country
+    , users_legaladdress.user_address_state_or_territory
+    , users_legaladdress.user_first_name
+    , users_legaladdress.user_last_name
+    , users_profile.user_gender
+    , users_profile.user_company
+    , users_profile.user_job_title
+    , users_profile.user_industry
+    , users_profile.user_job_function
```
abeglova, line 32: I wasn't sure which columns to exclude for privacy reasons. I ended up including everything except address information that is more granular than state/region from users_legaladdress and user_birth_year from users_profile. I still don't have a good sense of what user information is too private to include and and at what point in the process we should exclude iot

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -35,17 +35,104 @@ models:
...
   description: Intermediate model of users in MIT xPro.
   columns:
-  - name: id
-    description: ""
-  - name: username
-    description: ""
-  - name: email
-    description: ""
+  - name: user_id
+    description: int, primary key
```
rachellougee, line 47: nitpick, I think this description needs more descriptive to be consistent

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__users.sql
rachellougee: yeah, this might need discussion or requirement as to what column to include in the intermediate table. @dseaton thought?

rachellougee (approved):

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__users.sql
dseaton: I would match what we have for MITx Online: https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql

You might also consider removing `users_legaladdress.first_name` and `users_legaladdress.last_name` since we have a `users_legaladdress.full_name` field. 

As of right now, I have not seen a report actively utilized by xPro or Business Ops that requires anything beyond "country". There will be requests for PII in the future (e.g., legal address), but I say we avoid shipping any PII more granular than email/name/country to the intermediate layer.

dseaton (approved):
LGTM. 

One thing about documentation. We should talk to Tobias about field inheritance. I tried to find a dbt solution that would let intermediate and mart models inherit field descriptions from the staging layer (i.e., less writing out of doc strings). 

https://github.com/dbt-labs/dbt-core/issues/2995

PR #397, opened 2022-10-20
test that the documentation columns match the table

abeglova:
## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the code style of this project. (Did you install and run the pre-commit hooks?)


abeglova:
A less verbose alternative is to replace `dbt_expectations.expect_column_to_exist` after every column with
```
models: # or seeds:
  - name: my_model
    tests:
    - dbt_expectations.expect_table_columns_to_match_set:
        column_list: ["col_a", "col_b"]
```
once per model. That would also make the tests run faster. But then the is the possibility that a column name in the documentation won't match column_list, especially if we're doing lots of renaming

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
```diff
@@ -9,97 +9,142 @@ models:
...
   - name: user_is_active
     description: boolean, used to soft delete user accounts
   - name: user_full_name
     description: str, the user's full name
     tests:
     - not_null
+    - dbt_expectations.expect_column_to_exist
   - name: user_joined_on
     description: timestamp, specifying when a user account was initially created
-  - name: last_login
```
abeglova, line 30: I successfully caught a case where the column name in the documentation no longer matches the table while writing this pr

abeglova:
Those tests don't affect ` dbt run` at all but they do slow down `dbt test` (like any tests). I think that's ok, it's still relatively fast for now and eventually, as we add more test we will probably mostly run `dbt test` locally with  --select with or without this change

rachellougee (approved):
LGTM. 
There are some useful tests we could add later such as the multi-columns for compound key https://hub.getdbt.com/calogica/dbt_expectations/0.2.0/
It is a good start for us

PR #432, opened 2022-11-04
mitxonline enrollments

abeglova:
## Description

This pr adds the enrollment models for mitxonline

This pr adds the following models:
int__mitxonline__courserunenrollments
int__mitxonline__programenrollments
stg__mitxonline__app__postgres__courses_courserunenrollment
stg__mitxonline__app__postgres__courses_courserunenrollment


src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
```diff
@@ -316,3 +316,117 @@ models:
...
+    description: int, unique ID for each user on the MITx Online platform
+    tests:
+    - dbt_expectations.expect_column_to_exist
+    - not_null
+  - name: courserunenrollment_enrollment_mode
+    description: str, enrollment mode for user. One of "audit" and "verified"
+    tests:
+    - dbt_expectations.expect_column_to_exist
+    - accepted_values:
+        values: ['audit', 'verified']
```
rachellougee, line 370: This will likely fail in production because there are other values other than these two if we run  `SELECT DISTINCT change_status FROM courses_courserunenrollment;`

Although our app currently only expects these two, there are more options on the edx side https://courses-qa.mitxonline.mit.edu/admin/course_modes/coursemode/31/change/. 
image

Not entirely sure if we need to add all these options here, but something to keep in mind if these values flow into MITxOnline

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
```diff
@@ -316,3 +316,117 @@ models:
...
+  - name: courserunenrollment_has_edx_email_subscription
+    description: bool, indicates whether the user should receive emails from edx
```
rachellougee, line 372: Minor comment, just need to be consistent if it's bool or boolean

rachellougee (commented):
Just a couple of comments. Overall it looks good.

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
abeglova: good catch! Thanks

abeglova: So this enrollment mode is different from the course enrollment mode or the change status.
It's the enrollment_mode here:
https://rc.mitxonline.mit.edu/admin/courses/courserunenrollment/4/change/?_changelist_filters=o%3D1.4.5

Looks like I can change it to any value from the admin and some enrollments in production have enrollment_mode "honor"

I'm just going to remove the constraint

src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
```diff
@@ -2,28 +2,132 @@
...
+    tests:
+    - dbt_expectations.expect_column_to_exist
+    - not_null
+  - name: courserunenrollment_created_on
+    description: timestamp, specifying when an enrollment was initially created
+    tests:
+    - dbt_expectations.expect_column_to_exist
+    - not_null
+  - name: courserunenrollment_enrollment_mode
+    description: str, enrollment mode for user. One of "audit" and "verified"
```
rachellougee, line 35: might as well update the description to indicate other values are possible here

rachellougee (approved):
Looks good

Issue #443, opened 2022-11-09
MicroMasters users and profiles


rachellougee:
# User Story
Bring over MicroMasters Users and profiles to staging and intermediate so that we can combine them into a denormalized `users` table 

# Description/Context

Involved tables:

- `auth_user`
- `profiles_profile`
- `social_auth_usersocialauth` (provider='mitxonline' could be used to match with users in MITxOnline)

PR #448, opened 2022-11-15
MicroMasters users

rachellougee:
## Description

Staging and intermediate users model for MicroMasters
stg__micromasters__app__postgres__auth_user
stg__micromasters__app__postgres__auth_usersocialauth
stg__micromasters__app__postgres__profiles_profile
int__micromasters__users (adding two additional fields that could be used to match with mitxonline/edx.org users)

## Motivation and Context
https://github.com/mitodl/ol-data-platform/issues/443

## How Has This Been Tested?
Ran `dbt run` 
and the following commands:
`dbt test --select stg__micromasters__app__postgres__auth_user`
`dbt test --select stg__micromasters__app__postgres__auth_usersocialauth`
`dbt test --select stg__micromasters__app__postgres__profiles_profile`
`dbt test --select int__micromasters__users`

(Note that there are a few errors related to the MITxOnline programcertificate when running a full `dbt test` because duplicates exist in `raw__mitxonline__app__postgres__courses_programcertificate` Tobias said it's related to the S3 connector they are working on)

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
@@ -8,6 +8,26 @@ with users as (
     from {{ ref('stg__mitxonline__app__postgres__users_legaladdress') }}
 )
 
+, micromasters_profiles as (
+    select * from {{ ref('stg__micromasters__app__postgres__profiles_profile') }}
+)
+
+, micromasters_auth as (
+    select *
+    from {{ ref('stg__micromasters__app__postgres__auth_usersocialauth') }}
+    where user_auth_provider = 'mitxonline'
+)
+
+--- this profile is pull from MicroMasters so that we can match MM profile with MITxOnline users based on username
+, users_profiles as (
+    select
+        micromasters_profiles.user_profile_id as user_micromasters_profile_id
+        , micromasters_auth.user_username
+        , micromasters_profiles.user_address_country
+    from micromasters_profiles
+    inner join micromasters_auth on micromasters_auth.user_id = micromasters_profiles.user_id
+)
+
 select
     users.user_id
     , users.user_username
@@ -17,5 +37,7 @@ select
     , users.user_last_login
     , users_legaladdress.user_address_country
     , users.user_is_active
+    , users_profiles.user_micromasters_profile_id
 from users
 left join users_legaladdress on users_legaladdress.user_id = users.user_id
+left join users_profiles on users_profiles.user_username = users.user_username
--- a/src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
+++ b/src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
@@ -1,0 +1,270 @@
+---
+version: 2
+
+sources:
+- name: ol_warehouse_raw_data
+  loader: airbyte
+  database: 'ol_data_lake_{{ target.name }}'
+  schema: 'ol_warehouse_{{ target.name }}_raw'
+  tables:
+  - name: raw__micromasters__app__postgres__auth_user
+    columns:
+    - name: id
+      description: int, sequential ID representing a user in MicroMasters
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: password
+      description: str, hashed password
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: username
+      description: str, unique string to identify user
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: email
+      description: str, user email associated with their account
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: first_name
+      description: str, user's first name (blank for the majority of users on production)
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: last_name
+      description: str, user's last name (blank for the majority of users on production)
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: is_active
+      description: boolean, indicate whether this user is active (use it to soft delete
+        account)
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: last_login
+      description: timestamp, user's last login
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: date_joined
+      description: timestamp, specifying when a user account was initially created
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: is_staff
+      description: boolean, indicate whether the user can log into admin site (internally
+        used)
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: is_superuser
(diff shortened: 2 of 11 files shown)


Ferdi:
> int__micromasters__users (adding a field to match with mitxonline users)

Can you elaborate a bit about *adding a field to match with mitxonline users*

The main reason we are bringing MM users is to be able to augment the users/profiles of learners from edx.org and mitxonline.

rachellougee:
> int__micromasters__users (adding a field to match with mitxonline users)

> Can you elaborate a bit about adding a field to match with mitxonline users
> 
> The main reason we are bringing MM users is to be able to augment the users/profiles of learners from edx.org and mitxonline.

I am simply adding a field 'user_mitxonline_username' to the MM users intermediate table so that it could be used for matching MITxOnline users (currently no profile in MITxOnline). It is not limited to MITxOnline users, just a separate field for later use

Ferdi:
> user_mitxonline_username

Is this specifically for matching with mitxonline users?  Lots of useres in MM come from edx.org.  we might want to match with those profiles as well.

rachellougee:
> Is this specifically for matching with mitxonline users? Lots of useres in MM come from edx.org. we might want to match with those profiles as well.

Yes, this field is specifically for matching mitxonline users. I can add another field if we want to match it with edx.org

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
```diff
@@ -0,0 +1,43 @@
...
+            chr(32)
+            , nullif(first_name, '')
+            , nullif(last_name, '')
+        ) as user_full_name
+        , case
+            when edx_level_of_education = 'p' then 'Doctorate'
+            when edx_level_of_education = 'm' then 'Master''s or professional degree'
+            when edx_level_of_education = 'b' then 'Bachelor''s degree'
+            when edx_level_of_education = 'a' then 'Associate degree'
+            when edx_level_of_education = 'hs' then 'High school'
```
abeglova, line 27: I think it makes sense to standardize this with the xpro education levels https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxpro/_mitxpro__sources.yml#L612.

abeglova: Also either rename the column to user_highest_education or rename the column in xpro

src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
```diff
@@ -0,0 +1,270 @@
...
+      - dbt_expectations.expect_column_to_exist
+    - name: edx_employer
+      description: str, user's employer
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: phone_number
+      description: str, user's phone number
+      tests:
+      - dbt_expectations.expect_column_to_exist
+    - name: edx_job_title
```
abeglova, line 190: Looks like there are lots of columns here that you did not include in the staging data. Is that because those columns are blank? Or PII concerns? For xpro we included all the profile data in the staging table but not the intermediate tables

abeglova (changes requested):
Back to you with a few comments. Also, it probably makes sense to add a intermediate model with mitxonline data joined to MM profile data, since that was the motivation for importing this data

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
rachellougee: Renamed it.
(I was hesitant to rename this field to user_highest_education, but if we want to be consistent with xpro, then makes sense)

src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
rachellougee: I added all the fields to the staging

rachellougee:
@abeglova Please take a look. I've also added the MM profile to intermediate, so that it's available to pull from.

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
```diff
@@ -19,3 +38,5 @@ select
     , users.user_is_active
```
abeglova, line 38: Maybe add users_profiles.user_profile_id here?

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
abeglova: It probably makes sense to rename the labels to

"Secondary/high school", "Junior secondary/junior high/middle school" "Elementary/primary school"

to match xpro as well. Either that, or change the xpro levels to be less wordy.

abeglova (approved):
Two small comments, otherwise LGTM

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
rachellougee: so I don't know if we should change the value to match with xpro.
These options are taken from MicroMasters app, if we change it, it's not how user chose when they sign up
```
    LEVEL_OF_EDUCATION_CHOICES = (
        (DOCTORATE, 'Doctorate'),
        (MASTERS, "Master's or professional degree"),
        (BACHELORS, "Bachelor's degree"),
        (ASSOCIATE, "Associate degree"),
        (HIGH_SCHOOL, "High school"),
        (JUNIOR_HIGH_SCHOOL, "Junior high school"),
        (ELEMENTARY, "Elementary school"),
        (NO_FORMAL_EDUCATION, "No formal education"),
        (OTHER_EDUCATION, "Other education"),
    )
```

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__users.sql
rachellougee: I am not sure the purpose of adding user_profile_id here ?  it might cause confusion as that's not MITxOnline profile id. I made CTL users_profiles here so we can add any field from the MM profile as the data is already matched at this point.

abeglova: Maybe add it as micromasters_profile_id then? The idea is that we want queries that are using the intermediate tables as inputs can easily join to that data

abeglova (commented):
Back to you with a few comments. Also, it probably makes sense to add a intermediate model with mitxonline data joined to MM profile data, since that was the motivation for importing this data

rachellougee:
added `user_micromasters_profile_id`. and the test ran fine

PR #511, opened 2023-01-05
fix formating

abeglova:
This changes the max line length back to 120 and fixes the linting to the new version of sqlfluff.

Sqlfluff now has slightly new rules for indents since version 2.0.0 no longer supports hanging indents https://docs.sqlfluff.com/en/stable/layout.html#hanging-indents


rachellougee (approved):
LGTM

Issue #329, opened 2022-09-21
Synchronize edx.org course exports to S3


blarghmatey:
# User Story
As an engineer working on MIT Open I would like to be able to integrate course contents from edx.org along with our other Open edX installations.

# Description/Context
IRX is maintaining a Google Cloud Storage bucket with the `.tar.gz` course bundles for courses being hosted on edx.org. All of our other Open edX installations have pipelines configured to load course bundles into date-stamped paths in S3 on a nightly basis. In order to reduce the level of development effort we would like to replicate that interface by synchronizing the GCS course exports to S3 using the same path formats. The pipeline code that manages that export is located at https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/ops/open_edx.py#L482-L576

The bucket in GCS that has the course bundles is `[simeon-mitx-course-tarballs](gs://simeon-mitx-course-tarballs)`

# Acceptance Criteria
- [x] There is a pipeline that replicates the course exports from edx.org from GCS to S3
- [x] The pipeline is triggered by a sensor that monitors for changes/new contents in the source GCS bucket

shaidar:
The pipeline code that manages the export is located here - https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/ops/open_edx.py#L630-L704

shaidar:
Looked into using our existing `bigquery` resource and mainly just changing the last line in yielding a storage client, however it appears that we have to create a separate resource (which was mainly copy/paste). Ran into a permissions issues using the existing creds to read from GCS and am working with Jon to see what permissions need to be tweaked on their end to resolve it.

shaidar:
Chatted with Jon and it seems like this is not something he can do himself and is contacting IST to create a custom role for this to work.

shaidar:
Checked in with Jon on Monday and he still hadn't heard anything. Will ping him again today to see if there are any updates.

shaidar:
We now have access to the GCS bucket.

shaidar:
Have most of the pipeline code written except for connecting the output from one op (downloading the files from GCS) to the input into the other op (uploading the assets to S3) which I will be working on today and then testing out on QA.

shaidar:
Ran into some issues with the dagster container builds after adding the gcs pipeline. Tobias ended up making some changes to how we're building containers for pipeline use in dagster which ended up being a blocker, and now we're back to testing the pipeline on QA.

shaidar:
[...]

PR #520, opened 2023-01-11
Sync GCS edx course assets to S3

shaidar:
## Description

PR addresses following [issue](https://github.com/mitodl/ol-data-platform/issues/329)
## Types of changes

## Checklist:

### DIFF
--- a/src/ol_orchestrate/jobs/edx_gcs_courses.py
+++ b/src/ol_orchestrate/jobs/edx_gcs_courses.py
@@ -1,0 +1,53 @@
+from dagster import graph
+from dagster_aws.s3.resources import s3_resource
+
+from ol_orchestrate.lib.hooks import (
+    notify_healthchecks_io_on_failure,
+    notify_healthchecks_io_on_success,
+)
+from ol_orchestrate.lib.yaml_config_helper import load_yaml_config
+from ol_orchestrate.ops.edx_gcs_courses import (  # noqa: WPS235
+    download_edx_gcs_course_data,
+    upload_edx_gcs_course_data_to_s3,
+)
+from ol_orchestrate.resources.healthchecks import (
+    healthchecks_dummy_resource,
+    healthchecks_io_resource,
+)
+from ol_orchestrate.resources.gcp_gcs import gcp_gcs_resource
+from ol_orchestrate.resources.outputs import daily_dir
+
+@graph(
+    description=(
+        "Extract Open edX data maintained by institutional research "
+        "from a gcs bucket into a date-stamped paths in S3 on a "
+        "nightly basis."
+    ),
+    tags={
+        "source": "gcs",
+        "destination": "s3",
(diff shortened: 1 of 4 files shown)


src/ol_orchestrate/ops/edx_gcs_courses.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@op(
+    name="edx_course_tarballs",
+    description="Download edx course tarballs from GCS bucket",
+    required_resource_keys={"results_dir"},
+    out={"edx_course_tarball_directory": Out(dagster_type=DagsterPath)},
+)
+def download_edx_gcs_course_data(context):
+    credentials = service_account.Credentials.from_service_account_info(access_json)
```
blarghmatey, line 25: This appears to be redundant to the credential assignment being done in the resource definition.

src/ol_orchestrate/resources/gcp_gcs.py
```diff
@@ -0,0 +1,79 @@
...
+)
+def gcp_gcs_resource(resource_context: InitResourceContext):
+    """Create a connection to gcs.
+
+    :param resource_context: Dagster execution context for configuration data
+    :type resource_context: InitResourceContext
+
+    :yields: A gcs client instance for use during pipeline execution.
+    """
+    access_json = {
```
blarghmatey, line 60: Do we want to require specifying each of these nested attributes, or do we prefer to take a service account JSON blob wholesale?

src/ol_orchestrate/jobs/edx_gcs_courses.py
```diff
@@ -0,0 +1,50 @@
...
+production_resources = {
+    "gcp_gcs": gcp_gcs_resource,
+    "s3": s3_resource,
+    "results_dir": daily_dir,
+    "healthchecks": healthchecks_io_resource,
+}
+
+sync_gcs_to_s3_job = sync_gcs_to_s3.to_job(
+    name="sync_gcs_to_s3_job_pipeline",
+    resource_defs=dev_resources,
```
blarghmatey, line 49: This is hardcoding the dev resources for this job. We will want to do a conditional based on the `DAGSTER_ENV` variable

src/ol_orchestrate/resources/gcp_gcs.py
shaidar: Think in the long-term it might come in handy being able to tweak one of the attributes as opposed to taking it as wholesale JSON blob. Maybe most of the fields are the same except for `project_id` for example.

shaidar:
@blarghmatey Can we go ahead and merge this?

blarghmatey (approved):
:+1:

blarghmatey:
This will need some updates to the BUILD and workspace definitions to package the pipeline and set it up to run in docker-compose

Issue #479, opened 2022-12-09
MicroMasters alumni report


pdpinch:
# User Story

As an admin in MIT Alumni Association, I need periodic updates of learners who who have completed MicroMasters programs so I can invite them to join the MITAA. 

# Description/Context

Learners who have earned a MicroMasters program credential are eligible for benefits from the MIT Alumni Association (MITAA). To make these benefits available, we deliver a list of contact & profile data to the MITAA, roughly 4 times per year. (The schedule is timed to work well with course end dates). 

The data is sourced from several places: 

1. mitxonline
   1.  course certificates for DEDP 
   2. program certificates for DEDP 
   3. profile: email, first name, last name and country)
4. edx.org bigquery data:
   1. course certificates for SCM, SDS, POM and FIN programs
   2. profile: email, name, address, country
3. As far as I know, there is no source of program certificates for SCM, SDS, POM and FIN programs, so we have to calculate them, based on the requirements of each program 
6. micromasters:
   1. augmented profile information: ascii first name, ascii last name, address

The MITAA has requirements for data they can ingest:
1. separate first and last names
2. ASCII text only. No unicode 
3. incremental deliveries each time

pdpinch:
The most recent report for this is at https://bi.odl.mit.edu/queries/1115 which pulls data from https://docs.google.com/spreadsheets/d/13tIKdblVxdfrdfe_dhIwtbCYWxTEi903-ahLE6DJ8w8/edit#gid=987139461

PR #533, opened 2023-01-20
Ab/micromasters certificates

abeglova:
This PR adds a mart query to generate the micromasters alumni report (excluding DEDP for now)

## Motivation and Context
Part of https://github.com/mitodl/ol-data-platform/issues/479 

## How Has This Been Tested?
dbt run and dbt test

Extensive testing with production data to make sure that the generated data matches the existing alumni report

### DIFF
--- a/src/ol_dbt/models/marts/micromasters/marts__micromasters_program_certificates.sql
+++ b/src/ol_dbt/models/marts/micromasters/marts__micromasters_program_certificates.sql
@@ -1,0 +1,146 @@
+with edx_course_certificates as (
+    select *
+    from {{ ref('int__edxorg__mitx_courserun_certificates') }}
+)
+
+, edx_users as (
+    select *
+    from {{ ref('int__edxorg__mitx_users') }}
+)
+
+, micromasters_program_requirements as (
+    select *
+    from {{ ref('int__micromasters__program_requirements') }}
+)
+
+, micromasters_users as (
+    select *
+    from {{ ref('int__micromasters__users') }}
+)
+
+
+, micromasters_profiles as (
+    select *
+    from {{ ref('stg__micromasters__app__postgres__profiles_profile') }}
+)
+
+, micromasters_programs as (
+    select *
+    from {{ ref('int__micromasters__programs') }}
+)
+
+, micromasters_courses as (
+    select
+        course_id
+        , program_id
+        , course_edx_key
+        , course_number
+        , course_edx_key as course_readable_id
+    from {{ ref('stg__micromasters__app__postgres__courses_course') }}
+)
+
+, micromasters_user_requirement_completions as (
+    select
+        edx_course_certificates.user_id as user_edxorg_id
+        , edx_course_certificates.user_username as user_edxorg_username
+        , edx_course_certificates.courserun_readable_id
+        , edx_course_certificates.courseruncertificate_status
+        , edx_course_certificates.courseruncertificate_created_on
+        , edx_course_certificates.courserun_title
+        , micromasters_courses.program_id as micromasters_program_id
+        , micromasters_courses.course_id as micromasters_course_id
+        , micromasters_program_requirements.programrequirement_type
+        , micromasters_program_requirements.program_num_required_courses
+        -- The next line makes electiveset_required_number set for both rows for elective completions
+        -- and rows for core course completions
+        , avg(
+            micromasters_program_requirements.electiveset_required_number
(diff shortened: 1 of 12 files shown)


src/ol_dbt/models/marts/micromasters/marts__micromasters_program_certificates.sql
```diff
@@ -0,0 +1,148 @@
...
+    where user_course_certificate_number = 1
+
+)
+
+, program_completions as (
+    select
+        user_edxorg_id
+        , user_edxorg_username
+        , program_id
+        , min(courseruncertificate_created_on) as program_completion_timestamp
```
rachellougee, line 109: Should it be max instead of min? we are taking the earliest courserun certificate as program completion here

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
```diff
@@ -37,6 +36,9 @@ with source as (
         , edx_name as user_edx_name
         , edx_goals as user_edx_goals
         , fake_user as user_profile_is_fake
+        , to_iso8601(from_iso8601_timestamp(date_of_birth)) as user_birth_date
```
rachellougee, line 39: I am not sure if this is a good idea to convert the birth date as a timestamp field, if we just want to get the birth year, we should be able to get the year from the date since the year is just the first four digits

rachellougee (commented):
I left a few comments. Overall the logic looks ok but I didn't look into the report

src/ol_dbt/models/marts/micromasters/marts__micromasters_program_certificates.sql
abeglova: We do want the min - this is the earliest course run certificate that the user meets all the requirements for the certificate. Some users continue taking courses after  they earn the certificate. if a user takes additional courses in the program, we don't want to change their certificate earn date

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
rachellougee: I don't remember if we ever convert the date field to timestamp, just the timestamp fields. For birthdate, It feels different as it should always be a date as we are not recording a person's exact birth time. If we indeed convert date field to timestamp, I guess it's fine

src/ol_dbt/models/marts/micromasters/marts__micromasters_program_certificates.sql
rachellougee: I imagine the logic work in this way
- taking the earliest course run certificate for each course
- talking the max of each course certificate aggregated above that meets all the requirements

abeglova: No because of the where clause we pick the first course where the running total of both core and electives courses is enough to fulfill the  program requirements. So if the the program requires 2 core courses and 2 electives, and the user than takes a 3rd elective after already earning certificate, we count the date when they just met the program requirement with 2 core courses and 2 electives not the date when they earned the extra elective.

abeglova: That's why I do the thing with the running total instead of grouping and taking the maximum of courseruncertificate_created_on as the program_completion_timestamp

abeglova: I try to explain in the comment in line 77 but I'll add an extra comment here as well

rachellougee: Ok, looking at the logic the second time, It now makes sense. I saw the comment on line 77, but I think my confusion was about how we calculated "first completes the program requirement", which was in the where clause in lines 111 to 116. That was the thing I missed the first time looking at it.

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
rachellougee: ok, if we want to convert birth date to timestamp, please use the macro

rachellougee (approved):
👍

Issue #514, opened 2023-01-06
Review and update int__combined__* tables


rachellougee:
# User Story
There are a few intermediate combined tables that were created as initial dbt works, one of them is removed from our code, but we should review these tables and update accordingly

- int__combined__enrollments
- int__combined__course_runs
- int__combined__users

These tables are supposed to combine the data from different platforms - mitxonline, Bootcamp, xpro

rachellougee:
These combined models are removed from the codebase as they are simply UNION ALL not very useful. We can add them once we know the use cases

rachellougee:
Per meeting discussion, we want these combined tables from different platform

PR #543, opened 2023-01-26
config int__combined__ models as view

rachellougee:
## Description

config int__combined__ models as view instead of the table as the underlying models are already tables, so combined model requires no additional space

## Motivation and Context

https://github.com/mitodl/ol-data-platform/issues/514

## How Has This Been Tested?

ran `dbt run` and `dbt test`


abeglova (approved):
LGTM

PR #564, opened 2023-02-14
Ab/dbt meta testing

abeglova:
Our dbt model yml files have many many repetitive `dbt_expectations.expect_column_to_exist` and `dbt_expectations.expect_table_column_count_to_equal` tests. The purpose of these tests is to ensure that we document every column in the models. Unfortunately these repetitive tests make the documentation harder to read and make it harder to notice the tests that convey important information about the data, such as uniqueness and not null constraints when reading the documentation.

Instead of checking for complete documentation in the tests we can use [https://github.com/tnightengale/dbt-meta-testing](the dbt-meta-testing package) to ensure that we document every column in the models and don't have typos/name discrepancies between the models and the tests. 

to run the package run
`dbt deps`
`dbt run-operation required_docs`

This will return 
1. A list of models are missing descriptions. We can probably fix this, but all of these models have self descriptive names 
2. A list of columns missing from  the models.yml file. This is the reason for installing the package
3. A list of columns missing descriptions. We generally don't have a problem with this


rachellougee (approved):
👍

Issue #569, opened 2023-02-16
stg__micromasters__app__postgres__profiles_profile fails to update on production


abeglova:
the refresh for models/staging/micromasters/stg__micromasters__app__postgres__profiles_profile.sql
Fails with
 TrinoUserError(type=USER_ERROR, name=INVALID_ARGUMENTS, message="Timezone with non-zero seconds offset cannot be rendered as ISO8601: -04:56:02", query_id=20230216_143116_48453_ki6gj)

The date_of_birth field is the problem

PR #570, opened 2023-02-16
fix date_of_birth error

abeglova:
closes https://github.com/mitodl/ol-data-platform/issues/569

In order to standardize date formatting we run to_iso8601(from_iso8601_timestamp()) on all dates. Apparently this breaks for dates without timestamps with years prior to 1883. This is preventing stg__micromasters__app__postgres__profiles_profile and all tables that depend on it from being updated on production.

Since birth days from the 1800s aren't real anyway, we can just replace those dates with null


rachellougee (approved):
LGTM

Issue #558, opened 2023-02-10
Create intermediate model for MITx users


rachellougee:
# User Story
Create an intermediate model for MITx Users, where users are from edx and MITx Online
- edx users are mostly from BigQuery table `mitx_user_info_combo`, but it also includes users from `MITProfessionalX` and `MITxPRO` which should probably be excluded
- MITx Online users should be directly from our mitxonline app, but there is no profile data on mitxonline database. For these users who also have accounts in MicroMasters, we could get their profile data from MicroMasters if their account is linked to MITxOnline 

# Description/Context

# Acceptance Criteria
- [ ] MITx Users intermediate model is created in OL data lake

rachellougee:
https://github.com/mitodl/ol-data-platform/issues/568 should be addressed before we work on this

rachellougee:
As for my understanding, MITx user's data come from:

- MITx users on edx.org - this is in the BigQuery table `mitx_user_info_combo` we get from IRx
- MITx Online user's data from our application 

But the BigQuery table `mitx_user_info_combo` also includes users from
- `MITProfessionalX` e.g. MITProfessionalX/IOTx/2017_T2
-  MITxPRO courses. e.g. MITxPRO/SysEngxB2/1T2019
-  Our open edx users - we can exclude them because we can get these directly from our app

Do we exclude these users from `MITProfessionalX` and old MITxPRO courses listed at https://bi.odl.mit.edu/queries/1233/source?
@Ferdi @pdpinch 
Let me know if there is something else I am missing

pdpinch:
MITProfessionalX and MITxPRO learners are not MITx learners. They should not be included in the MITx user data.

PR #576, opened 2023-02-22
add MITx Users model

rachellougee:
## Description

It adds int__mitx__users model, which combines MITx Online and edX.org users. 
It filters out those users from old MITProfessionalX and xPro courses

## Motivation and Context

https://github.com/mitodl/ol-data-platform/issues/558

## How Has This Been Tested?

ran dbt run on staging and intermediate models, and dbt test

## Types of changes

## Checklist:

### DIFF
--- a/src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql
+++ b/src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql
@@ -1,0 +1,34 @@
+---MITx users from MITx Online and edX
+
+{{ config(materialized='view') }}
+
+with mitxonline_users as (
+    select * from {{ ref('int__mitxonline__users') }}
+)
+
+, edxorg_users as (
+    select * from {{ ref('int__edxorg__mitx_users') }}
+)
+
+, mitx_users as (
+    select
+        '{{ var("mitxonline") }}' as platform
+        , user_id
+        , user_username
+        , user_email
+        , user_address_country
+    from mitxonline_users
+
+    union distinct
+
+    select
+        '{{ var("edxorg") }}' as platform
+        , user_id
+        , user_username
+        , user_email
+        , user_country as user_address_country
+
+    from edxorg_users
+)
+
+select * from mitx_users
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql
```diff
@@ -0,0 +1,34 @@
...
+, mitx_users as (
+    select
+        '{{ var("mitxonline") }}' as platform
+        , user_id
+        , user_username
+        , user_email
+        , user_address_country
+    from mitxonline_users
+
+    union distinct
```
abeglova, line 22: This can just be union instead of union distinct

abeglova (approved):
LGTM other than the one small comment

src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql
rachellougee: sqlfluff added it automatically when I just write union

abeglova (changes requested):
Acutally, looks like you are not deduping between mitxonline users and edxorg users (and probably can't with the data we have). I don't really see the point of this pr then? Having them both in one table without deduping just makes joins harder

src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql
abeglova: Ok it's fine then

rachellougee:
I guess the question is how do we dedup users between two platforms - MITx Online and edx.org? @abeglova @pdpinch  @Ferdi?

abeglova:
For non-micromasters users we probably can't dedupe (other than by using email which is unreliable) but then why create this table?

abeglova (approved):
LGTM

rachellougee:
Per discussion, we don't need to worry about deduplication right now as these are two different platforms. We will deal with duplicates later if any. We still want a model for MITx users, thus this PR is needed.

PR #580, opened 2023-02-24
config: Add per-developer schemas to dbt workflow

blarghmatey:
In order to eliminate the potential for engineers to accidentally overwrite each-other's work while developing, this adds an automatic per-developer schema creation for data generated from the raw sources. This will cause all tables rendered via `dbt run`, etc. to be created in their own namespaced locations in Trino/Glue/S3. This also allows for working directly against production data without altering the tables used by active consumers of the data.

## Description
Adds templating of the schema names used based on a `developer_name` variable so that each developer can have a namespaced database location to generate their tables into while developing.

## Motivation and Context
As analytics engineers are iterating on models, particularly in intermediate and mart layers, they need to be able to work against production data to properly validate their work. This change allows them to build against production raw data without having to generate multiple copies of it, and also without modifying "live" data that is being consumed by production users.

## How Has This Been Tested?
I ran this against the QA environment and validated that the databases and tables were created in the proper location.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [x] Enhancement (improves on existing behavior)

## Checklist:


src/ol_dbt/README.md
```diff
@@ -35,7 +35,11 @@ dbt models (SQL files) are structured into 3 directories under ol_dbt/models, wi
   export DBT_TRINO_PASSWORD=<PASSWORD>
   ```
   You need Starburst credentials to complete this step, which can be requested from OL Devops
-
+- Set your `developer_name` variable in `src/ol_dbt/.user.yml`. This will be used to create a namespaced database schema that will hold the tables that are generated during your development workflow. This should look like:
+```yaml
+vars:
+  developer_name: tmacey
```
rachellougee, line 41: This doesn't seem to work for me, not sure if I am missing something.

I can only make it work by hardcoding in profiles.yml ` schema: ol_warehouse_qa_rlougee`. Once I change it to ` schema: ol_warehouse_qa{{ "_{}".format(var('developer_name', '')).rstrip("_")
        }}`, dbt run just creates tables under ol_warehouse_qa. 
 
Other than that, dbt run seems work for me

blarghmatey: So, I had assumed that adding the var in `.user.yml` would cause it to be loaded but it seems I'm wrong. Perhaps you have your dbt user config in a different location? Alternatively you can pass `--var 'developer_name: rlougee'` at the command line. We definitely don't need to hardcode the schema

rachellougee: yes, I have a local .user.yml file that is under the same directory .dbt as profiles.yml
```
id: fd3c10cf-142a-46a7-xxxxx
vars:
  developer_name: rlougee
```

I tried a few times, but I can only get it to work by either hardcoding or using the command line as you suggested --var 'developer_name: rlougee'. I guess for now it's fine.

rachellougee: Looking at https://docs.getdbt.com/docs/build/project-variables#variable-precedence, It seems like the var needs to be in dbt_project.yml if not using the command line. var in .user.yml doesn't seem to work. Can we update the README for the instruction? Otherwise, it looks good to me

setup.cfg
```diff
@@ -1,54 +1,3 @@
...
-    test_*.py:S101,WPS218
-inline-quotes = double
-i_control_code = True
-exclude =
-    .git,
-    .venv,
-    __pycache__
-
 [mypy]
 python_version=3.9
```
shaidar, line 2: Can we bump this version?

blarghmatey:
So, exploring the idea of pulling in the git branch via Jinja is coming up short. That leads me to the possibility of writing a CLI wrapper that automatically exports the current git branch as an env variable to be used in templating the schema name. Alternatively, we create an additional pair of targets like `dev_qa` and `dev_production` that include the schema templating and require explicitly passing the `--var` parameter to specify the schema namespace, which has a high potential for typos and generating lots of extraneous schemas (which is the same situation as the git branch approach and just requires an automated way of cleaning up those schemas on a scheduled basis).

rachellougee (approved):
Tested against qa `dbt run --select stg__bootcamps__app__postgres__courses_course --var 'schema_suffix: TEST_BRANCH' ` (modified my local profiles.xml), that seems fine. 

I noted a small comment on missing closing bracket on production schema, but I can't test against dev_qa/dev_production yet as ` Catalog 'ol_data_lake_dev_qa' does not exist"` which I believe it's in other PR.  I assume there will be separate PR to automate cleaning up those schemas on a scheduled basis.

I am going to approve it for now but waiting for Sar to take a look at as well.

setup.cfg
blarghmatey: We can revisit that in a separate PR. There might be some limiting factors on that still.

src/ol_dbt/profiles.yml
```diff
@@ -22,3 +22,23 @@ open_learning:
...
       http_scheme: https
+    dev_qa:
+      type: trino
+      method: ldap
+      user: '{{ env_var("DBT_TRINO_USERNAME") }}'
+      password: '{{ env_var("DBT_TRINO_PASSWORD") }}'
+      database: ol_data_lake_qa
+      port: 443
+      schema: ol_warehouse_qa{{ "_{}".format(var('schema_suffix')) }}
+      host: mitol-ol-data-lake-qa-0.trino.galaxy.starburst.io
```
shaidar, line 33: nitpick: My personal preference would be to have the hostnames set up as env vars.

shaidar:
Looks good overall

Issue #563, opened 2023-02-14
MicroMasters alumni report since 12/20/2022


pdpinch:
# User Story

As an admin in MIT Alumni Association, I need periodic updates of learners who have completed MicroMasters programs so I can invite them to join the MITAA.

As part of the work on #479, we delivered an update through 12/19/2022. It's time now to deliver an update through 2/15/2023, including the DEDP credential holders added in #539. 

# Acceptance Criteria

Our practice has been to drop an Excel file on [dropbox](https://www.dropbox.com/sh/uk9op9acclrahf5/AAAKxhBotwaTzzKHSmlZfr2Pa?dl=0) with the list of updated credential holders and their profile data. If there is a more convenient and self-service way to do this, please propose something here.

pdpinch:
@abeglova is there a PR for this (maybe a merged one)?

I've noticed a couple of issues in the data you sent me earlier: 

- Some learners are missing their full names. They also happen to be learners with a null for `user_username` which I believe means that they don't have an edx.org account. It looks like the full name is available in `ol_warehouse_production_intermediate.int__mitxonline__users.user_full_name` 
- The same learners that are missing full names are also missing their country. It looks like that's in the same intermediate table, in `user_address_country`

PR #581, opened 2023-02-24
micromasters mart fixes

abeglova:
## Description
closes https://github.com/mitodl/ol-data-platform/issues/563

This makes the following updates
1) We replace ESD.SCM1x' with 'CTL.SC1x' in the courserun_readable_ids in stg__edxorg__bigquery__mitx_user_info_combo so that that course matches the course id in micromasters and the course id naming scheme for other courses. We need to also do this in stg__edxorg__bigquery__mitx_person_course or the certificates earned in that course are missing from int__edxorg__mitx_courserun_certificates
2) Some mitxonline DEDP learners don't have a edx account (or we can't match them to edx user info) we should get country and full name for those users from the mitxonline user data
3) Peter asked for a specific column order in the final mart


rachellougee (approved):
👍

Issue #587, opened 2023-03-01
Create micromasters ecommerce staging and intermediate models


abeglova:
Turns out we never created staging and intermediate ecommerce models for micromasters. We should create them now

PR #588, opened 2023-03-02
micromasters ecommerce staging

abeglova:
This creates the staging models for micromasters ecommerce
Closes https://github.com/mitodl/ol-data-platform/issues/587
As discussed in the meeting, we're not going to create intermediate models for these tables


src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
```diff
@@ -2,6 +2,212 @@
...
+    - not_null
+  - name: order_updated_on
+    description: timestamp, specifying when the order was most recently updated
+    tests:
+    - not_null
+  - name: order_state
+    description: string, order state. Options are "fulfilled", "failed", "created"
+      "refunded", "partially_refunded"
+    tests:
+    - not_null
```
rachellougee, line 99: To be consistent, this should add accepted_values same way as other platforms

src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
```diff
@@ -2,6 +2,212 @@
...
+  - name: coupon_object_id
+    description: int, id in either coures_course or courses_program for the course
+      entity for which the coupon can be redeemed
+    tests:
+    - not_null
+  - name: coupon_amount_type
+    description: string, type of coupon discount and meaning of the "amount" column.
+      May be "fixed-discount" "fixed-price" or "percent-discount"
+    tests:
+    - not_null
```
rachellougee, line 178: same here. the list is fixed in MM AMOUNT_TYPES = [PERCENT_DISCOUNT, FIXED_DISCOUNT, FIXED_PRICE] so we should add accepted_values

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__ecommerce_redeemedcoupon.sql
```diff
@@ -0,0 +1,20 @@
...
+)
+
+, renamed as (
+
+    select
+        id as redeemedcoupon_id
+        , order_id
+        , coupon_id
+        , {{ cast_timestamp_to_iso8601('created_on') }} as redeemedcoupon_created_on
+        , {{ cast_timestamp_to_iso8601('created_on') }} as redeemedcoupon_updated_on
```
rachellougee, line 14: `{{ cast_timestamp_to_iso8601('updated_on') }} as redeemedcoupon_updated_on`

rachellougee (changes requested):
Left a few comments, one of them is typo so it needs to be fixed. Otherwise it looks ok

src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
abeglova: there are some weird values for this field on staging. I don't want to add accepted_values so staging test won't fail

rachellougee: ok. looks like from two old test records on staging

rachellougee (approved):
LGTM

PR #589, opened 2023-03-03
micromasters intermediate refactor

abeglova:
As discussed at the meeting yesterday int__micromasters__program_certificates and int__micromasters__users because they don't include all micromasters students and certificate earners. Also mart_components is a bad name

This PR
1) Renames `intermediate/mart_components` as `intermediate/subqueries`
2) Moves int__micromasters__users into subqueries, since not all micromasters learners are included in that table
3) replaces int__micromasters__program_certificates with the content of marts_micromasters__program_certificates


rachellougee (approved):

Ferdi:
> int__micromasters__users into subqueries, since not all micromasters learners are included in that table

if these models are incomplete,  do we want to have a special notation, like `__int__micromasters__users`. . Then we say "dont use models starrtign with double underscore"

abeglova:
@Ferdi sorry, the PR description is confusing. The model is now called subqueries__micromasters__users

rachellougee:
One thing to remind you, these renamed models will not be dropped in production, we might need to manually drop the old name https://docs.getdbt.com/faqs/models/removing-deleted-models

abeglova:
@Ferdi Oh, sorry you meant that instead of having "subquery" in the table name i can add __ to the subquery tables.

Yeah! I like that

rachellougee (approved):

Issue #528, opened 2023-01-19
MicroMasters Summary report


pdpinch:
# User Story

As a member of a MicroMasters team, I'd like to have a top-line summary of data so we can track the health of the program overall. 

# Description/Context

We currently have a report at https://bi.odl.mit.edu/queries/1056 that uses data collected in a google sheet https://docs.google.com/spreadsheets/d/13tIKdblVxdfrdfe_dhIwtbCYWxTEi903-ahLE6DJ8w8/edit#gid=0 

The [script](https://github.mit.edu/anindyar/autorep_mm) that is used to produce this summary has to be run manually and can no longer be maintained. We need something that is run on a regular basis (at least once a month) and is more reliable. In particular, we expect the number of total enrollments and unique learners to be always increasing.

# Acceptance Criteria

- [ ] A new BI report that uses data from the data lake instead of relying on the google sheet.

Summary data for:
- [ ] enrollments 
- [ ] verified learners
- [ ] certificates 

# Related issues 

- is it possible to have some kind of alert when the number of enrollments and unique learners decrease between runs?

blarghmatey:
#538 isn't a blocker, that's just an optimization to cut down on overall sync times for the data.

abeglova:
This is completed for enrollments but not verified learners and certificates

abeglova:
Also, we currently don't import marts into bi. That should be an easy fix for devops

pdpinch:
@abeglova I got a question from the DEDP team that I could probably answer, but I'd like to verify with you: 

> I'm writing with a follow-up question to the data you shared at the last MicroMasters meeting. If I recall correctly, your table listed ~300,000+ unique DEDP learners. Our course team has counted ~50,000 learners, so we're curious to know how you're arriving at your number, and what we're doing differently.

Can you briefly explain our sources and method? 

I suspect they don't have access to the full extent of edx enrollment data. I think the first run of 14.73x alone had more than 40,000 enrollments.

pdpinch:
This is partially done, which can be seen in the data mart `ol_warehouse_production_mart.marts__micromasters_summary`
 
| program_title | total_enrollments | unique_users | unique_countries
| -- | -- | -- |  -- | 
| Supply Chain Management | 1,038,566 | 493,487 | 251 |  
| Data, Economics, and Development Policy | 591,913 | 329,810 | 245 |  
| Statistics and Data Science | 848,292 | 512,995 | 251 |  
[...]

PR #594, opened 2023-03-07
enrollments summary

abeglova:
This is the first pr of https://github.com/mitodl/ol-data-platform/issues/528

With the total enrollments, unique enrollments and unique countries columns. We will add additional columns in future report. Data generated with 
[ Data generated with this code vs the existing report](https://github.com/mitodl/ol-data-platform/files/10911666/mm_summary.1.csv)

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courseruns.sql
+++ b/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courseruns.sql
@@ -39,4 +39,8 @@ from
 --- courserun_readable_id here is formatted as {org}/{course_number}/{run}
 left join
     micromasters_courses
-    on runs.courserun_readable_id like micromasters_courses.course_readable_id || '%'
+    on
+        replace(
+            replace(runs.courserun_readable_id, '6.041x_3', '6.431x'), '6.041x_4', '6.431x'
+        ) like micromasters_courses.course_readable_id
+        || '%'
--- a/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
@@ -2,6 +2,49 @@
 version: 2
 
 models:
+- name: int__micromasters__course_enrollments
+  description: MicroMasters enrollments
+  columns:
+  - name: program_title
+    description: str, tite of micromasters program the course is part of
+    tests:
+    - not_null
+  - name: user_id
+    description: str, either the edx.org or the mitxonline id for the user
+    tests:
+    - not_null
+  - name: user_username
+    description: str, either the edxorg or the mitxonline username for the user
+    tests:
+    - not_null
+  - name: user_country
+    description: str, country code for the user's address
+  - name: user_email
+    description: str, user email
+    tests:
+    - not_null
+  - name: courserun_readable_id
+    description: str, course ID on edx.org or openedx
+    tests:
+    - not_null
+  - name: platform
+    description: str, the platform of the course. Either 'MITx Online' or 'edX.org'
+    tests:
+    - not_null
+  - name: courserunenrollment_created_on
+    description: timestamp, timestamp of user's enrollment in the course
+  - name: courserunenrollment_is_active
+    description: boolean, whether the enrollment is active
+    tests:
+    - not_null
+  - name: courserun_title
+    description: str, title of the course
+    tests:
+    - not_null
+  tests:
+  - dbt_expectations.expect_compound_columns_to_be_unique:
+      column_list: ["user_id", "platform", "courserun_readable_id"]
+
 - name: int__micromasters__programs
   description: Intermediate model for programs in MicroMasters
   columns:
--- a/src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml
@@ -2,6 +2,98 @@
 version: 2
 
 models:
+- name: __micromasters_course_enrollments_from_mitxonline
+  description: MicroMasters enrollments on mitxonline
+  columns:
+  - name: program_title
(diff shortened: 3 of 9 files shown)


src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courseruns.sql
```diff
@@ -39,4 +39,8 @@ from
 --- courserun_readable_id here is formatted as {org}/{course_number}/{run}
 left join
     micromasters_courses
-    on runs.courserun_readable_id like micromasters_courses.course_readable_id || '%'
+    on
+        replace(
+            replace(runs.courserun_readable_id, '6.041x_3', '6.431x'), '6.041x_4', '6.431x'
```
abeglova, line 44: I these two runs of 6.041 (but not other runs) count as course runs of 6.431 according to the micromasters database

src/ol_dbt/models/marts/micromasters/_marts_micromasters__models.yml
```diff
@@ -2,6 +2,28 @@
...
 models:
+- name: marts__micromasters_summary
+  description: MicroMasters course enrollments
+  columns:
+  - name: program_title
+    description: str, Micromasters program tile
+    tests:
+    - not_null
+  - name: total_enrollments
+    description: str, count of course enrollments in the program
```
pdpinch, line 13: should we add a note here that this counts who users who _ever_ enrolled, and does not omit users who later unenrolled?

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
```diff
@@ -2,6 +2,45 @@
 version: 2
 
 models:
+- name: int__micromasters__course_enrollments
```
rachellougee, line 5: What's the uniqueness of this table? Is it user_id + courserun_readable_id + courserunenrollment_is_active + platform?should we add that?

src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml
```diff
@@ -2,6 +2,92 @@
 version: 2
 
 models:
+- name: __micromasters_course_enrollments_from_mitxonline
+  description: MicroMasters enrollments on mitxonline
```
rachellougee, line 6: same here. It's probably worth adding the uniqueness in this model

src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml
```diff
@@ -2,6 +2,92 @@
...
+      course through another platform than edx, such as micromasters are still listed
+      as 'audit' users
+  - name: courserun_readable_id
+    description: str, the canonical ID of the course run in the format as {org}/{course}/{run}
+    tests:
+    - not_null
+  - name: platform
+    description: str, the platform of the course- always 'edX.org'
+    tests:
+    - not_null - not_null
```
rachellougee, line 80: dup?

src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_enrollments_from_edxorg.sql
```diff
@@ -0,0 +1,41 @@
...
+    , edx_enrollments.courserun_readable_id
+    , edx_enrollments.courserunenrollment_created_on
+    , edx_enrollments.courserunenrollment_is_active
+    , edx_enrollments.courserun_title
+    , '{{ var("edxorg") }}' as platform
+from edx_enrollments
+left join edx_runs
+    on edx_enrollments.courserun_readable_id = edx_runs.courserun_readable_id
+left join edx_users on edx_users.user_id = edx_enrollments.user_id
+inner join micromasters_programs on micromasters_programs.program_id = edx_runs.micromasters_program_id
```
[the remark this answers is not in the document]
abeglova: The inner join filters to only those course enrollments that are part of a micromasters program and filters out all others. I can change the other two joins to inner joins too

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
abeglova: Sure I can add that

abeglova: Probably just user_id + courserun_readable_id + platform should be enough since a user who deactivates then reactivates should be the same enrollment

src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml
abeglova: fixed

src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_enrollments_from_edxorg.sql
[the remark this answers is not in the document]
rachellougee: right. inner joins make sense here, otherwise, program_title wouldn't always be not null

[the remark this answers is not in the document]
abeglova: Since the last join is an inner join, all the non-null program titles are already filtered out, but you're right all inner joins is clearer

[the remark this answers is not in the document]
rachellougee: Maybe we think the same thing, but the last join won't guarantee not-null program titles if this doesn't match edx_enrollments.courserun_readable_id = edx_runs.courserun_readable_id with left join. The data probably makes no difference here, but logically there are differences.

rachellougee (approved):
LGTM

PR #721, opened 2023-06-07
add secondary sorting to MM course certificates and grades

rachellougee:
N/A
I noticed this issue when testing DEDP course certificates

This PR adds a secondary sorting to the subqueries in the MM DEDP course certificate/grade models to ensure that if learners get the highest grade from different runs within the same DEDP course, the query would then pick the latest grade from these runs but this run needs to be before the DEDP course certificates were generated. This logic only applies to DEDP course certificates/grades from MicroMasters

In MM, DEDP course certificates were generated based on courses not runs, so we have to populate runs with their highest grade.

e.g Let's say a learner gets the highest grade for both `course-v1:MITxT+JPAL102x+3T2021` and `course-v1:MITx+JPAL102x+3T2020` with 0.94, and the DEDP course certificate was generated on 2022-02-01, we would choose  `course-v1:MITxT+JPAL102x+3T2021` for the DEDP course certificate

# How can this be tested?
dbt build on +int__micromasters__course_certificates/grades
dbt tests passed

### DIFF
--- a/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_certificates_dedp_from_micromasters.sql
+++ b/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_certificates_dedp_from_micromasters.sql
@@ -1,5 +1,7 @@
---- Unlike other platforms, DEDP course certificate from MM is based on course not run, we try to match it with
---- course run from learner's highest grades
+--- Unlike other platforms, MicroMasters DEDP course certificates are based on course not run, we try to find the run
+--- with highest grade for the course. If there are multiple runs with the highest grade, pick the latest grade from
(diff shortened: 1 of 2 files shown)


pdpinch:
I’m not sure I understand the circumstances here, but normally when there is more than one we grade we choose the highest, not the most recent.

rachellougee:
@pdpinch PR description is confusing, I will update it to make it clear. 

The query has already chosen the highest grade, the issue is when learners get the highest grade for different runs in the same course, the query just randomly picked the run with the highest grade. This PR is intended to add secondary sorting to ensure we get consistent results when there are multiple runs with the highest grade

quazi-h (approved):
Logic here looks good, just finishing up running the dbt build.

PR #727, opened 2023-06-12
Fix MM none-dedp program certificates to match with MM courses

rachellougee:

https://github.com/mitodl/hq/issues/1471#issuecomment-1585051201

This PR fixes the none-dedp program certificate subquery to match with MicroMasters courses so that it doesn't accidentally drop the course certificates where edx course key (run) doesn't match with edx key (course) in the string. e.g.`MITx/6.041x_3/2T2016` and  `MITx/6.041x_4/1T2017`

It also removes logic where the course ID `MITx/ESD.SCM1x/3T2014` was replaced with  `MITx/CTL.SC1x/3T2014` in edxorg models as this edx_course_key is `MITx/ESD.SCM1x/3T2014` in MM and there is already logic to take care of course_number matching logic for this course, so we shouldn't need to change the course ID for `ESD.SCM1x` in edX.org

# How can this be tested?

Run dbt build on `+int__micromasters__program_certificates` (need to refresh related staging and intermediate models)
Run 
``` 
select count(*), program_title 
from "ol_data_lake_production"."ol_warehouse_production_rlougee_intermediate"."int__micromasters__program_certificates"
group by program_title
 ```
Make sure the non-DEDP program certificate counts are:

- Finance - 153
- **Statistics and Data Science - 1269**
- Principles of Manufacturing - 249
- Supply Chain Management - 4500  (this should be the same as production)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -186,21 +186,6 @@ dev = ["cogapp", "pre-commit", "pytest", "wheel"]
 tests = ["pytest"]
 
 [[package]]
-name = "arrow"
-version = "1.2.3"
-description = "Better dates & times for Python"
-category = "main"
-optional = false
-python-versions = ">=3.6"
-files = [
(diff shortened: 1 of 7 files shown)


abeglova (approved):
lgtm

PR #746, opened 2023-06-27
create marts for MITx Online certificate and profile data

rachellougee:

https://github.com/mitodl/hq/issues/1716
https://github.com/mitodl/hq/issues/1748

This adds `marts__mitxonline_course_certificates` and `marts__mitxonline_user_profiles` for IRx. 

# How can this be tested?

Run `dbt build --select marts.mitxonline` if you have all intermediate models, otherwise add + in front of marts.mitxonline to run upstream models
Run `dbt test --select int__mitxonline__users` as well as `int__bootcamp__users` and `int__mitxpro__users` to ensure the test pass since I made the change to accepted_values for user_highest_education


rachellougee:
@blarghmatey These two marts are created for IRx. I didn't specifically create directories named IRx just to be consistent with the existing marts for micromasters. But let me know if you have trouble applying proper permission to these two marts

src/ol_dbt/models/marts/mitxonline/_marts_mitxonline__models.yml
```diff
@@ -0,0 +1,77 @@
...
+        values: ['Male', 'Female', 'Transgender', 'Non-binary/non-conforming', 'Other/Prefer
+            Not to Say', '']
+  - name: user_highest_education
+    description: str, user highest education
+    tests:
+    - accepted_values:
+        values: ['Doctorate', "Master''s or professional degree", "Bachelor''s degree",
+          'Associate degree', 'Secondary/high school', 'Junior secondary/junior high/middle
+            school', 'Elementary/primary school', 'No formal education', 'Other education',
+          '']
```
pdpinch, line 77: Is this just copied & pasted from the other intermediate model? Is there a way to more formally connect these and avoid copying and pasting?

pdpinch:
@KatelynGit @blarghmatey can you both review this?

Kate, I want to make sure you understand and agree with the SQL and the schema. 

Tobias, you'll have to share this with Jon Daries so he can pull it regularly. You know the best what's needed for that.

src/ol_dbt/models/marts/mitxonline/_marts_mitxonline__models.yml
rachellougee: Yes, it's a test copied from the intermediate. I can put these in a var that can be reused across different apps and models

blarghmatey:
At the mart level the affiliation with products is not as important as in staging or even intermediate layers. For the marts we should scope them by use case/consumer. In this case I'd argue for a layout like `marts/irx/mitxonline_certificates` with a schema that is just a combination of the two tables as a single flat representation. That eliminates the need for joins on the consumer side, as well as scoping the tables to an `irx` namespace. Similarly for when we start building the `finance` marts.

rachellougee:
I've renamed marts to `mitxonline` per discussion with Tobias. @KatelynGit @blarghmatey This is ready to be reviewed

KatelynGit:
Looks like there are 6,515 users that are in the ol_warehouse_production_intermediate.int__mitxonline__users but not in the new mart user table. I'm assuming this might just be a timing thing or due to proper filtering but wanted to double check. Can you take a look?

rachellougee:
@KatelynGit Yes, if you were looking at my schema, it's stale and has not been updated for a while. Since this mart is simply a select from intermediate, I didn't run all the upstream models as we have test in intermediate to ensure count match with staging, etc.

KatelynGit (approved):
The changes to add user and certificate data to the marts looks good. The schema matches what I'd expect based on the requirements/other tickets. And the data is flowing through as expected. The difference in counts was explained and expected as well.

src/ol_dbt/models/marts/mitxonline/marts__mitxonline_course_certificates.sql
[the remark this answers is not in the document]
rachellougee: We are using `marts__micromasters_*` in other mart models

blarghmatey (approved):

Issue #738, opened 2023-06-23
Update DEDP program certificates model to not use hard-coded program name


rachellougee:
# Description/Context

There will be a new DEDP Program "Public Policy" launches on 6/28, and the existing DEDP program name will likely be changed as part of the launch. https://github.com/mitodl/hq/issues/1662

There are some hard-coded program names `Data, Economics and Development Policy` in the intermediate MicroMasters program certificate model. That logic will break once the course team updates the existing DEDP program name, so we need to update that to use `dedp_micromasters_program_id` (2) and `dedp_mitxonline_program_id` (1) since the primary ID will unlikely be changed

# Plan/Design

Update the hardcoded DEDP program title in the intermediate MicroMasters program certificate model

rachellougee:
https://github.com/mitodl/ol-data-platform/pull/747 is merged and code is deployed

PR #747, opened 2023-06-28
Update hardcoded dedp program title

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/738

The existing DEDP program has been renamed to `Data, Economics and Development Policy: International Development` just now. That would break the DEDP certificate in `int__micromasters__program_certificates` when dbt is refreshed tomorrow as it relies on the program title. This PR is to fix that.

# How can this be tested?

run dbt build on `int__micromasters__program_certificates`


abeglova (approved):
lgtm

PR #753, opened 2023-07-10
Add new MITXPRO tables to int & mart

KatelynGit:
This adds a new int table int__mitxpro__ecommerce_combinedcoupons  and a new mart table marts__mitxpro_ecommerce_combinedcoupons

https://github.com/mitodl/hq/issues/818

creating data mart tables as described in https://github.com/mitodl/hq/issues/1803

# How can this be tested?

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -22,19 +22,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -60,7 +60,7 @@ models:
       "refunded" all associated with the coupon
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
-      column_list: ["coupon_id", "order_ids", "product_id"]
+      column_list: ["coupon_id", "order_ids", "product_id", "couponpaymentversion_payment_transaction"]
 - name: int__mitxpro__ecommerce_couponversion
   description: A version of a coupon linked to a specific couponpaymentversion record
   columns:
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -22,19 +22,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order  
+    description: string, contract number used to identify the order
   - name: product_id
-    description: int, foreign key in ecommerce_product 
+    description: int, foreign key in ecommerce_product
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies 
+      by companies
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run   
+    description: str, title of the course run
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -11,24 +11,26 @@ models:
     description: string, coupon code for the coupon
   - name: coupon_id
     description: int, foreign key referencing ecommerce_coupon
+  - name: couponversion_updated_on
+    description: timestamp, specifying when the coupon version was most recently updated
   - name: coupon_created_on
     description: timestamp, specifying when the coupon was initially created
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -36,9 +38,9 @@ models:
     description: str, title of the course
   - name: productversion_readable_id
     description: string, the readable_id field from the product object
-  - name: combinedcoupon_table_source
-    description: string, Describes whether the data on the combioned coupon table
-      is coming from b2b or elsewhere
+  - name: combinedcoupon_table_source   
+    description:  string, Describes whether the data on the combioned coupon table is 
+      coming from b2b or elsewhere
   - name: couponpaymentversion_coupon_type
     description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
       used multiple times
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -18,19 +18,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order  
+    description: string, contract number used to identify the order
   - name: product_id
-    description: int, foreign key in ecommerce_product 
+    description: int, foreign key in ecommerce_product
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies 
+      by companies
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run   
+    description: str, title of the course run
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -38,9 +38,9 @@ models:
     description: str, title of the course
   - name: productversion_readable_id
     description: string, the readable_id field from the product object
-  - name: combinedcoupon_table_source   
-    description:  string, Describes whether the data on the combioned coupon table is 
-      coming from b2b or elsewhere
+  - name: combinedcoupon_table_source
+    description: string, Describes whether the data on the combioned coupon table
+      is coming from b2b or elsewhere
   - name: couponpaymentversion_coupon_type
     description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
       used multiple times
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -9,28 +9,32 @@ models:
     description: string, email associated to the coupon
   - name: coupon_code
     description: string, coupon code for the coupon
+    tests:
+    - not_null
   - name: coupon_id
     description: int, foreign key referencing ecommerce_coupon
-  - name: couponversion_updated_on
-    description: timestamp, specifying when the coupon version was most recently updated
+    tests:
+    - not_null
   - name: coupon_created_on
     description: timestamp, specifying when the coupon was initially created
+    tests:
+    - not_null
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -38,9 +42,6 @@ models:
     description: str, title of the course
   - name: productversion_readable_id
     description: string, the readable_id field from the product object
-  - name: combinedcoupon_table_source
-    description: string, Describes whether the data on the combioned coupon table
-      is coming from b2b or elsewhere
   - name: couponpaymentversion_coupon_type
     description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
       used multiple times
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -16,19 +16,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order  
+    description: string, contract number used to identify the order
   - name: product_id
-    description: int, foreign key in ecommerce_product 
+    description: int, foreign key in ecommerce_product
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies 
+      by companies
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run   
+    description: str, title of the course run
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -36,9 +36,9 @@ models:
     description: str, title of the course
   - name: productversion_readable_id
     description: string, the readable_id field from the product object
-  - name: combinedcoupon_table_source   
-    description:  string, Describes whether the data on the combioned coupon table is 
-      coming from b2b or elsewhere
+  - name: combinedcoupon_table_source
+    description: string, Describes whether the data on the combioned coupon table
+      is coming from b2b or elsewhere
   - name: couponpaymentversion_coupon_type
     description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
       used multiple times
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -22,19 +22,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -60,7 +60,7 @@ models:
       "refunded" all associated with the coupon
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
-      column_list: ["coupon_id", "order_ids"]
+      column_list: ["coupon_id", "order_ids", "product_id"]
 - name: int__mitxpro__ecommerce_couponversion
   description: A version of a coupon linked to a specific couponpaymentversion record
   columns:
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -22,19 +22,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -58,6 +58,9 @@ models:
   - name: combined_order_state
     description: string, order state. Options are "fulfilled", "failed", "created"
       "refunded" all associated with the coupon
+  tests:
+  - dbt_expectations.expect_compound_columns_to_be_unique:
+      column_list: ["coupon_id", "order_ids"]
 - name: int__mitxpro__ecommerce_couponversion
   description: A version of a coupon linked to a specific couponpaymentversion record
   columns:
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -9,24 +9,26 @@ models:
     description: string, email associated to the coupon
   - name: coupon_code
     description: string, coupon code for the coupon
+  - name: coupon_id
+    description: int, foreign key referencing ecommerce_coupon
   - name: coupon_created_on
     description: timestamp, specifying when the coupon was initially created
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
@@ -34,9 +36,9 @@ models:
     description: str, title of the course
   - name: productversion_readable_id
     description: string, the readable_id field from the product object
-  - name: combinedcoupon_table_source
-    description: string, Describes whether the data on the combioned coupon table
-      is coming from b2b or elsewhere
+  - name: combinedcoupon_table_source   
+    description:  string, Describes whether the data on the combioned coupon table is 
+      coming from b2b or elsewhere
   - name: couponpaymentversion_coupon_type
     description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
       used multiple times
@@ -49,10 +51,10 @@ models:
   - name: user_address_countries
     description: str, country codes for the user(s)
   - name: order_ids
-    description: int, foreign key for ecommerce_orders
+    description: int, foreign key for ecommerce_orders all associated with the coupon
   - name: combined_order_state
     description: string, order state. Options are "fulfilled", "failed", "created"
-      "refunded"
+      "refunded" all associated with the coupon
 - name: int__mitxpro__ecommerce_couponversion
   description: A version of a coupon linked to a specific couponpaymentversion record
   columns:
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -22,19 +22,19 @@ models:
   - name: couponpayment_name
     description: string, human readable name for the coupon payment
   - name: b2border_contract_number
-    description: string, contract number used to identify the order
+    description: string, contract number used to identify the order  
   - name: product_id
-    description: int, foreign key in ecommerce_product
+    description: int, foreign key in ecommerce_product 
   - name: couponpaymentversion_payment_transaction
     description: string, string that identifies the payment invoice for coupon purchases
-      by companies
+      by companies 
   - name: req_reference_number
     description: b2b receipt req reference number for a payment
   - name: programrun_readable_id
     description: str, unique program run ID formatted as program-v1:{org}+{program
       code}+{run_tag} e.g. program-v1:xPRO+MLx+R1
   - name: program_title
-    description: str, title of the course run
+    description: str, title of the course run   
   - name: courserun_readable_id
     description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}
       e.g. course-v1:xPRO+MLx1+R0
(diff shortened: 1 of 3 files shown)


src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
```diff
@@ -0,0 +1,33 @@
...
+)
+
+, ecommerce_productversion as (
+    select *
+    from {{ ref('int__mitxpro__ecommerce_productversion') }}
+)
+
+select
+    commerce_coupon.coupon_id
+    , ecommerce_productcouponassignment.product_id
```
abeglova, line 18: This table is supposed to be usable to end users so you should include information about the course run/program that is the product

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_userorder_details.sql
```diff
@@ -0,0 +1,48 @@
...
+    , couponpaymentversion.couponpaymentversion_coupon_type
+    , couponpaymentversion.couponpaymentversion_discount_amount
+    , couponpaymentversion.couponpayment_name
+    , couponpaymentversion.couponpaymentversion_id
+    , b2becommerce_b2border.b2border_contract_number
+    , JSON_EXTRACT_SCALAR(b2becommerce_b2breceipt.b2breceipt_data, '$.req_reference_number') as req_reference_number
+from ecommerce_order
+inner join users
+    on ecommerce_order.order_purchaser_user_id = users.user_id
+inner join couponpaymentversion
```
abeglova, line 43: I don't think all orders use coupons, so this is excluding the orders where ecommerce_order.couponpaymentversion_id is null

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_userorder_details.sql
```diff
@@ -0,0 +1,48 @@
...
+)
+
+, couponpaymentversion as (
+    select *
+    from {{ ref('int__mitxpro__ecommerce_couponpaymentversion') }}
+)
+
+select
+    ecommerce_order.coupon_id
+    , b2becommerce_b2border.product_id
```
abeglova, line 28: Again this is meant to be used by end users so  coupon_id/product_id/order_id won't be useful information for them

abeglova (changes requested):

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
abeglova: I think the product id people want to see is the readable id of the course run/program that the ecommerce_product record is associated with

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
```diff
@@ -0,0 +1,33 @@
...
+    , commerce_coupon.coupon_code
+    , commerce_coupon.couponpayment_name
+    , commerce_coupon.coupon_created_on
+    , max(ecommerce_productcouponassignment.productcouponassignment_created_on) as maxproductcouponassignment_created_on
+    , ecommerce_productcouponassignment.productcouponassignment_is_redeemed
+    , ecommerce_productcouponassignment.productcouponassignment_email
+    , ecommerce_productversion.productversion_readable_id
+from commerce_coupon
+left join ecommerce_productcouponassignment
```
abeglova, line 27: So b2b coupons don't have a ecommerce_productcouponassignment which is why https://bi.odl.mit.edu/queries/983/source has the union

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -0,0 +1,75 @@
+---
+version: 1
+
+models:
```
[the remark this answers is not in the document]
KatelynGit: removed additional 'models:' text

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
KatelynGit: I haven't included any course or program info in the table so far. This would require new joins to other tables with very different data. If we need it I'll add it but may have to research and add as a separate table maybe.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_userorder_details.sql
KatelynGit: changed to a left join thanks

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_userorder_details.sql
KatelynGit: I only added these so they could join between the two tables but I agree alone they are not useful

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
KatelynGit: That's definitely good to know. For the other marts__mitxpro_ecommerce_userorder_details table we're pulling in the b2b order data that doesn't have a product coupon assignment and I've adjusted the  marts__mitxpro_ecommerce_productcoupon_details to use a left join to the product coupon assignment as well so hopefully that covers the issue.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
KatelynGit: Talked it over Anastasia and we'll add the additional tables and fields and I'll re-commit

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productcoupon_details.sql
```diff
@@ -0,0 +1,51 @@
...
+    , commerce_coupon.coupon_created_on
+    , ecommerce_productcouponassignment.productcouponassignment_is_redeemed
+    , ecommerce_productcouponassignment.productcouponassignment_email
+    , ecommerce_productversion.productversion_readable_id
+    , max(ecommerce_productcouponassignment.productcouponassignment_created_on)
+        as maxproductcouponassignment_created_on
+from commerce_coupon
+left join ecommerce_productcouponassignment
+    on commerce_coupon.coupon_id = ecommerce_productcouponassignment.coupon_id
+left join ecommerce_productversion
```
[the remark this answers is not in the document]
abeglova: You should join directly to courses_courserun and courses_program instead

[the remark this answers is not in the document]
abeglova: Is this why you needed the max()?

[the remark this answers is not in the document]
KatelynGit: The only table with the productversion_readable_id they requested is on the product version table so I had to pull that in unfortunately. And I don't see a join to product version through courses_courserun and courses_program. to eliminate duplicate rows I had a query above that one to filter them out - 
ecommerce_productversion as (
    select 
        product_id
        , productversion_readable_id 
    from productversion 
    where productversion_readable_id is not null
) 
The max piece was just handling a few records where we had multiple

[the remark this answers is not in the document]
abeglova: So I don't think you need productversion_readable_id (unless I missed where someone specifically asked for it) You need courserun_readable_id or product_readable id from the courserun or program tables/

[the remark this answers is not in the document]
abeglova: Taking the max is wrong anyway, since the latest product version is not nessesarily the correct one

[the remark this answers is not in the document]
abeglova: And if the productversion_readable_id (I'm not sure if it does) changes between product versions you
'll have a row for each one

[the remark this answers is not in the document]
KatelynGit: this field is needed per https://github.com/mitodl/hq/issues/1803 but new tables created instead so resolving this for now

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+        couponversion.coupon_id
+        , ecommerce_couponpaymentversion.couponpaymentversion_payment_transaction
+    from (
+        select
+            couponpayment_name
+            , max(couponpaymentversion_id) as couponpaymentversion_id
+        from int__mitxpro__ecommerce_couponpaymentversion
+        group by couponpayment_name
+    ) as couponpaymentversion
+    inner join ecommerce_couponpaymentversion as ecommerce_couponpaymentversion
```
abeglova, line 82: The as here is redundant

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+        on couponpaymentversion.couponpaymentversion_id = ecommerce_couponpaymentversion.couponpaymentversion_id
+    inner join couponversion as couponversion
```
abeglova, line 84: couponversion is never defined should this be ecommerce_coupon version?

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    select distinct
+        couponversion.coupon_id
+        , ecommerce_couponpaymentversion.couponpaymentversion_payment_transaction
+    from (
+        select
+            couponpayment_name
+            , max(couponpaymentversion_id) as couponpaymentversion_id
+        from int__mitxpro__ecommerce_couponpaymentversion
+        group by couponpayment_name
+    ) as couponpaymentversion
```
abeglova, line 81: This code is broken -  int__mitxpro__ecommerce_couponpaymentversion is never defined should id be ecommerce_couponpaymentversion? Also can the named subquery be defined before this query and called something other than couponpaymentversion since you are joining to ecommerce_couponpaymentversion?

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    inner join ecommerce_couponpaymentversion as ecommerce_couponpaymentversion
+        on couponpaymentversion.couponpaymentversion_id = ecommerce_couponpaymentversion.couponpaymentversion_id
+    inner join couponversion as couponversion
+        on couponpaymentversion.couponpaymentversion_id = couponversion.couponpaymentversion_id
+)
+
+select
```
abeglova, line 88: It would be helpful if you break this out into two named queries to make it clear that there are seperate subqueries for b2becommerce coupons and other coupons

abeglova (changes requested):
I wasn't able to run the query - still figuring out if it gives the right results

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
abeglova: Maybe max_couponpaymentversion? But also why are you getting a max by couponpayment_name and not couponpayment_id?

abeglova: I guess the grouping is ok since the couponpayment_name is unique and non_null https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxpro/_stg_mitxpro__models.yml#L463 and functionally an id

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: I'll remove this

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: yes ill change it

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: I changed it from int__mitxpro__ecommerce_couponpaymentversion to ecommerce_couponpaymentversion.  And changed the title to max_couponpaymentversion for the subquery.

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: to make this clear i added a field called combinedcoupon_table_source to show which data is which

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    end as coupon_redeemed
+    , array_join(array_distinct(array_agg(users_user.user_email)), ', ') as combined_user_emails
+    , array_join(array_distinct(array_agg(users_user.user_address_country)), ', ') as user_address_countries
+    , array_join(array_distinct(array_agg(ecommerce_order.order_id)), ', ') as order_ids
+    , array_join(array_distinct(array_agg(ecommerce_order.order_state)), ', ') as combined_order_state
+from b2b_ecommerce_b2border as b2b_ecommerce_b2border
+inner join ecommerce_couponpaymentversion as ecommerce_couponpaymentversion
+    on ecommerce_couponpaymentversion.couponpaymentversion_id = b2b_ecommerce_b2border.couponpaymentversion_id
+inner join coupon as coupon
+    on coupon.coupon_id = b2b_ecommerce_b2border.b2bcoupon_id
```
abeglova, line 193: This join is not correct b2bcoupon_id refers to theint__mitxpro__b2becommerce_b2bcoupon table not theint__mitxpro__ecommerce_coupon table

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    , b2b_ecommerce_b2border.b2border_contract_number
+    , productcouponassignment.product_id
+    , payment_transactions.couponpaymentversion_payment_transaction
+    , null
+    , program_runs.programrun_readable_id
+    , program_runs.program_title
+    , course_runs.courserun_readable_id
+    , courses.course_title
+    , productversion.productversion_readable_id
+    , 'regular coupon'
```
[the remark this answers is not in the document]
KatelynGit: fixed this i'll upload the latest

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+            json_extract(b2b_ecommerce_b2breceipt.b2breceipt_data, '$.req_reference_number')
+            as varchar
+        ), '"', ''
+    )
+    , program_runs.programrun_readable_id
+    , program_runs.program_title
+    , course_runs.courserun_readable_id
+    , courses.course_title
+    , productversion.productversion_readable_id
+    , 'b2b coupon'
```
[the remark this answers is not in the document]
KatelynGit: fixed this i'll upload the latest

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    , case
+        when (count(couponredemption.couponredemption_id) >= 1)
+            then 'True'
+        else 'False'
+    end as coupon_redeemed
+    , array_join(array_distinct(array_agg(users_user.user_email)), ', ') as combined_user_emails
+    , array_join(array_distinct(array_agg(users_user.user_address_country)), ', ') as user_address_countries
+    , array_join(array_distinct(array_agg(ecommerce_order.order_id)), ', ') as order_ids
+    , array_join(array_distinct(array_agg(ecommerce_order.order_state)), ', ') as combined_order_state
+from b2b_ecommerce_b2border as b2b_ecommerce_b2border
```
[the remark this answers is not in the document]
KatelynGit: fixed this i'll upload the latest

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    , ecommerce_couponpaymentversion.couponpaymentversion_discount_amount
+    , case when (count(couponredemption.couponredemption_id) >= 1) then 'True' else 'False' end as coupon_redeemed
+    , array_join(array_distinct(array_agg(users_user.user_email)), ', ') as combined_user_emails
+    , array_join(array_distinct(array_agg(users_user.user_address_country)), ', ') as user_address_countries
+    , array_join(array_distinct(array_agg(ecommerce_order.order_id)), ', ') as order_ids
+    , array_join(array_distinct(array_agg(ecommerce_order.order_state)), ', ') as combined_order_state
+from coupon
+left join productcouponassignment
+    on productcouponassignment.coupon_id = coupon.coupon_id
+left join ecommerce_couponversion
```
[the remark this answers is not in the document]
KatelynGit: fixed and uploading soon

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    on coupon.coupon_id = ecommerce_couponversion.couponversion_id
+left join ecommerce_couponpaymentversion
+    on ecommerce_couponpaymentversion.couponpaymentversion_id = ecommerce_couponversion.couponpaymentversion_id
+left join b2b_ecommerce_b2border
```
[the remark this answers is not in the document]
KatelynGit: I removed it updating soon

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: remove this join per other comments

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,235 @@
...
+    from {{ ref('int__mitxpro__ecommerce_productversion') }}
+)
+
+, b2b_ecommerce_b2breceipt as (
+    select *
+    from {{ ref('int__mitxpro__b2becommerce_b2breceipt') }}
+)
+
+, payment_transactions as (
+    select distinct
```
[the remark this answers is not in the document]
KatelynGit: i added a comment to the code explaining this

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,231 @@
...
+            as varchar
+        ), '"', ''
+    )
+    , program_runs.programrun_readable_id
+    , program_runs.program_title
+    , course_runs.courserun_readable_id
+    , courses.course_title
+    , productversion.productversion_readable_id
+    , ecommerce_couponpaymentversion.couponpaymentversion_coupon_type
+    , ecommerce_couponpaymentversion.couponpaymentversion_discount_amount
```
rachellougee, line 231: Instead of grouping 15+ fields in one query, could we break it down using CTEs with only required columns in the grouping first before passing it down to this query? I think it's both for readability and performance concerns. Just a suggestion.

KatelynGit: I made the change to break it down a bit using CTEs

KatelynGit:
I've made the requested changes, compared to the bi query and looks good so far. It's ready for review

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,313 @@
...
+        on ecommerce_couponversion.couponversion_id = couponredemption.couponversion_id
+    left join ecommerce_order
+        on ecommerce_order.order_id = couponredemption.order_id
+    left join users_user
+        on ecommerce_order.order_purchaser_user_id = users_user.user_id
+    group by
+        coupon.coupon_id
+)
+
+, reg_coupon_fields as (
```
abeglova, line 102: It looks like you are grouping by every single field in this query? Can you remove the group by entirely?

abeglova: I think you're essentially getting unique?

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,313 @@
...
+    left join program_runs
+        on ecommerce_line.programrun_id = program_runs.programrun_id
+    left join course_runs
+        on ecommerce_line.courserun_id = course_runs.courserun_id
+    left join courses
+        on course_runs.course_id = courses.course_id
+    left join productversion
+        on productversion.productversion_id = b2b_ecommerce_b2border.productversion_id
+    group by
+        b2b_ecommerce_b2border.b2border_email
```
abeglova, line 235: I think you are grouping by every field here too?

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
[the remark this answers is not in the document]
abeglova: I don't see the comment. Also, could you please not mark past comments as resolved?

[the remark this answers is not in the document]
abeglova: It makes it hard to check if previous issues were addressed in your update

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: Yes, that's correct I'm getting unique. If I remove the group by I'd get a bunch of duplicates in the table.

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: Yes, that's right

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
abeglova: Ok so you can use the distinct keyword instead of grouping by every single column. But also the fact that you need to is a code smell - do you know which one of the joins is creating the duplicate rows?

abeglova: Could you get rid of the group by/distinct by changing the order of the joins?

KatelynGit: It's a bunch of them but I'm basically joining down from coupon to more detailed tables(different grain) to pull some of the requested fields and it causes dups.

KatelynGit: As far as I know changing the order would still result in dups unfortunately

KatelynGit: I can definitely change it to distinct though that'll work great

abeglova: :( Ok let me see if I can rewrite this - this is definitely not ok to merge as is

abeglova: I think that's why the original bi query had a separate subquery for coupons with no redemptions - so the query could join from redemptions back to coupons in a way that doesn't create duplicates

KatelynGit: I don't know if there's much to be re-written...there was basically a list of fields and we were told to create a big denormalized table...this is it...but definitely open to ideas

KatelynGit: yea we can get rid of duplicated different ways but probably same results

abeglova: You get why the fact that you are doing a distinct there is a problem, right...

abeglova: Like for readability/maintainability of this query

KatelynGit: I think part of the problem is I was told to add even lower grain fields not part of the bi query...but I don't know of a way to do it without a distinct definitely open to ideas though

abeglova:
Looks like there are a bunch of coupons that are showing up as duplicates once in the b2b subquery and once in the regular subquery - For example  coupon_codes  269152f683c84636976fa66eede3a322 and f984451b4b6245ec802525b63acd4bf0 but there are lots more. If you check the bi query those appear only once

KatelynGit:
Yes that's true I believe it's because we're pulling more fields from lower grain tables

abeglova:
SELECT coupon_code
FROM "ol_data_lake_production"."ol_warehouse_production_abeglova_intermediate"."int__mitxpro__ecommerce_combinedcoupons"
where combinedcoupon_table_source = 'regular coupon'
and couponpaymentversion_coupon_type = 'single-use'
group by coupon_code
having count(*) > 1

Returns 606 rows

KatelynGit:
> SELECT coupon_code FROM "ol_data_lake_production"."ol_warehouse_production_abeglova_intermediate"."int__mitxpro__ecommerce_combinedcoupons" where combinedcoupon_table_source = 'regular coupon' and couponpaymentversion_coupon_type = 'single-use' group by coupon_code having count(*) > 1
> 
> Returns 606 rows

I changed a left join to an inner to eliminate most of the extra records but some are due to data differences since the table isn't at a coupon grain

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,313 @@
...
+        , payment_transactions.couponpaymentversion_payment_transaction
+        , program_runs.programrun_readable_id
+        , program_runs.program_title
+        , course_runs.courserun_readable_id
+        , courses.course_title
+        , productversion.productversion_readable_id
+        , ecommerce_couponpaymentversion.couponpaymentversion_coupon_type
+        , ecommerce_couponpaymentversion.couponpaymentversion_discount_amount
+    from coupon
+    inner join productcouponassignment
```
abeglova, line 119: Ok - in addition to the fact that this shouldn't be an inner join because not every coupon has a product assignment

there is one coupon code with more than one product coupon assignment. That coupon should not appear in the report twice with different product ids

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,313 @@
...
+        on productcouponassignment.coupon_id = coupon.coupon_id
+    left join ecommerce_couponversion
+        on coupon.coupon_id = ecommerce_couponversion.coupon_id
+    left join ecommerce_couponpaymentversion
+        on ecommerce_couponpaymentversion.couponpaymentversion_id = ecommerce_couponversion.couponpaymentversion_id
+    left join couponredemption
+        on ecommerce_couponversion.couponversion_id = couponredemption.couponversion_id
+    left join ecommerce_order
+        on ecommerce_order.order_id = couponredemption.order_id
+    left join payment_transactions
```
abeglova, line 129: So this join is one of the ones causing duplicates - Prior to this you had a uniqecouponpaymentversion_id per line (except for the weird edge case with productcouponassignment) Now you are joining every order associated with the coupon to every payment transaction associated with the coupon even when those are not matching

abeglova: This creates lines in your report with junk data

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: ok switched it back to a left join

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: I couldn't find any joins causing issues but I'm look at the counts as each one is added

KatelynGit:
I deployed a new version of the table without distincts/groupbys except for one section where there is aggregation. You may still find multiple records for one coupon but they are different coupons versions and previous versions can be filtered out at a report level. I think it looks good let me know if you see any issues.

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,264 @@
...
+                then 'true'
+            else 'false'
+        end as coupon_redeemed
+    from b2b_ecommerce_b2border
+    inner join ecommerce_couponpaymentversion
+        on ecommerce_couponpaymentversion.couponpaymentversion_id = b2b_ecommerce_b2border.couponpaymentversion_id
+    inner join ecommerce_couponversion
+        on ecommerce_couponversion.couponpaymentversion_id = ecommerce_couponpaymentversion.couponpaymentversion_id
+    inner join coupon
+        on coupon.coupon_id = ecommerce_couponversion.coupon_id
```
rachellougee, line 187: These two joins don't seem to match with BI, can you confirm if this is the correct join condition? `couponpaymentversion_id` to `coupon_id` is 1 to many, On BI,`coupon` to `couponpayment` joined on `couponpayment_id`, not via `coupon_id`

rachellougee (changes requested):
I left a few comments, I also checked a couple of coupons, data don't seem to match with BI. 

e.g. coupon_code ='efcafaa7868f4d5691ce38a4a41fa092'. 
- It should has order_id 26329, but its null in`int__mitxpro__ecommerce_combinedcoupons`.  
- There are two rows in `int__mitxpro__ecommerce_combinedcoupons` for this coupon_code as regular coupon, but there should be just 1 row with `program-v1:xPRO+SysEngx`
 
image

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,264 @@
...
+        , try_cast(users_user.user_address_country as BigInt) as user_address_countries
+        , try_cast(ecommerce_order.order_id as BigInt) as order_ids
+        , try_cast(ecommerce_order.order_state as BigInt) as combined_order_state
+        , replace(
+            cast(
+                json_extract(b2b_ecommerce_b2breceipt.b2breceipt_data, '$.req_reference_number'
+                )
+                as Varchar
+            ), '"', ''
+        ) as req_reference_number
```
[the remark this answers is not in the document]
KatelynGit: Changing this causes errors: L: 177 | P:  68 | RF02 | Unqualified reference 'omit' found in select with more
                       | than one referenced table/view.
                       | [references.qualification]
L: 177 | P:  73 | RF02 | Unqualified reference 'quotes' found in select with more
                       | than one referenced table/view.
                       | [references.qualification]

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
```diff
@@ -0,0 +1,264 @@
...
+        , program_runs.program_title
+        , course_runs.courserun_readable_id
+        , courses.course_title
+        , productversion.productversion_readable_id
+        , ecommerce_couponpaymentversion.couponpaymentversion_coupon_type
+        , ecommerce_couponpaymentversion.couponpaymentversion_discount_amount
+        , try_cast(users_user.user_email as BigInt) as combined_user_emails
+        , try_cast(users_user.user_address_country as BigInt) as user_address_countries
+        , try_cast(ecommerce_order.order_id as BigInt) as order_ids
+        , try_cast(ecommerce_order.order_state as BigInt) as combined_order_state
```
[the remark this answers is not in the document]
KatelynGit: I'll fix this

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
KatelynGit: This is the correct order it produced the same results

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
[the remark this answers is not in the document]
rachellougee: Take a look at the usage of https://trino.io/docs/current/functions/json.html#json-query and syntax, I used it for tracking log and it worked great, not sure why it would throws error if the original query is manually replacing the quote. I can take a look later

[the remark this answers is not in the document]
rachellougee: I don't see any error with this line `, json_query(b2b_ecommerce_b2breceipt.b2breceipt_data, 'lax $. req_reference_number' omit quotes) as req_reference_number`. It produces the correct req_reference_number. Were you running the same code?

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,61 @@
 version: 2
 
 models:
+- name: int__mitxpro__ecommerce_combinedcoupons
```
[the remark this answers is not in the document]
KatelynGit: I've added tests to the fields and reuploaded

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_combinedcoupons.sql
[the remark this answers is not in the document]
KatelynGit: yes the sql can run like that but the pre-commit doesn't like that code and throws an error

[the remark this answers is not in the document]
rachellougee: sqlfluff throws a validation error for this line because of the reference to `b2b_ecommerce_b2breceipt.b2breceipt_data`. The subquery b2b_coupon_fields is doing a lot of joins on top of extracting a field from one of the tables, which sqlfluff didn't like.  As mentioned, in my opinion, it's best to add this field to staging to be able to reuse it. Alternatively, you can put that at the top when you import this model, so that in b2b_coupon_fields just reference this new field.
```
, b2b_ecommerce_b2breceipt as (
    select
        b2border_id
        , json_query(b2breceipt_data, 'lax $. req_reference_number' omit quotes) as req_reference_number
    from {{ ref('int__mitxpro__b2becommerce_b2breceipt') }}
)
```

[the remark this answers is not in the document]
KatelynGit: I made this change

KatelynGit:
All the issues mentioned above have been addressed and the newly designed table is available to query

KatelynGit:
I've fixed the error it was being caused by 2 strange records in the data

abeglova:
I checked and the bi report has 21542 but your report has 24038. Do you know where the extra rows are coming from?

KatelynGit:
I believe I'm using more left joins where the report has inner joins and it is causing different counts

rachellougee (commented):
There are still dups that need to be looked into, especially the unique fields for this table.

KatelynGit:
### a new PR has been opened and competed for combined coupon work

PR #776, opened 2023-07-24
xpro model additions

abeglova:
None

This pr adds xpro columns that have been added since the data was staged


src/ol_dbt/models/staging/mitxpro/_stg_mitxpro__models.yml
```diff
@@ -1104,6 +1112,9 @@ models:
     description: timestamp, specifying when a course run was initially created
   - name: courserun_updated_on
     description: timestamp, specifying when a course run was most recently updated
+  - name: courserun_certificate_available_date
```
rachellougee, line 1115: Is this added to sql file?

abeglova: Oops, yeah it should be

abeglova: Actually nm I was working on both mitxonline and xpro changes  but decided to break it up into separate prs. certificate_available_date was actually added to mitxonlike i updated the wrong documentation file

rachellougee (approved):
👍

PR #784, opened 2023-08-02
extract course ID for xPro tracking log

rachellougee:

NA

This PR extracts the course ID from `event_type` when `course_id` is either blank or missing from the `context` field. There are still some events without course ID but most of these have no course ID anywhere.

# How can this be tested?

This is a large model, do the following before running dbt build
- delete from `stg__mitxpro__openedx__tracking_logs__user_activity` 
- replace dev_production [host](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/profiles.yml#L460) with the production host in your profile.xml


abeglova (approved):
lgtm

PR #785, opened 2023-08-02
654 Migrate Dagster to Pythonic Config

### LINKED ISSUE #654: Update Dagster code to use Pythonic Config
# User Story
- As a data engineer I want to stay up to date with Dagster's API design

# Description/Context
Dagster recently updated their APIs to support Pydantic-based config models and are promoting them in place of the previous dictionary objects passed into decorators. This is a more robust and easy to use/understand interface that we would like to adopt. They have created a migration guide [here](https://docs.dagster.io/guides/dagster/migrating-to-pythonic-resources-and-config)

# Acceptance Criteria
- [ ] All Dagster code uses the new Pythonic config interface

quazi-h:
Closes #654

Converts the ops and resources in our Dagster pipelines to use Pydantic-based config models instead of dictionary objects passed through decorators. This will keep us up to date with Dagster's API design and make our configs more robust and easier to understand.

# How can this be tested?
Dagster pipelines should be run locally and on QA and they should function as they were prior to these changes.

# Additional Context
Dagster's new API encourages us to make incremental changes and allows for the old configs and new configs to exist at the same time. This means we should be able to run the pipelines normally, even if all of the configs have not been migrated yet.

## Checklist:
- [ ] Test updated ops by running the open edx, edx gas courses, and log normalization pipelines.
- [ ] Test updated resources by running the open edx, edx gas courses, and log normalization pipelines.


src/ol_orchestrate/ops/normalize_logs.py
```diff
@@ -48,29 +59,32 @@ def load_files_to_table(context: OpExecutionContext, log_date: str) -> Nothing:
...
     :log_date: S3 Bucket log date prefix to load logs from (Format 'YYYY-MM-DD/')
     :type log_date: str
 
     """
-    source_bucket = context.op_config["tracking_log_bucket"]
+    source_bucket = config.tracking_log_bucket
     # DuckDB Glob Syntax: ** matches any number of subdirectories (including none)
     s3_path = f"s3://{source_bucket}/logs/{log_date}**"
-    context.log.info(s3_path)
+    config.log.info(s3_path)
```
blarghmatey, line 72: I think this needs to remain as the `context` object, since the logger instance isn't part of the config model

src/ol_orchestrate/ops/normalize_logs.py
```diff
@@ -180,7 +193,7 @@ def write_file_to_s3(
         for file_name in files:
             new_file_name = file_name.replace("logs", "valid")
             local_file_name = new_file_name.rsplit("/", maxsplit=1)[-1]
-            context.log.info(new_file_name)
+            config.log.info(new_file_name)
```
blarghmatey, line 196: Same here, this should remain as `context`

src/ol_orchestrate/ops/normalize_logs.py
```diff
@@ -190,7 +203,7 @@ def write_file_to_s3(
             # copy to S3
             context.resources.s3.upload_file(
                 Filename=local_file_name,
-                Bucket=context.op_config["tracking_log_bucket"],
+                Bucket=config.op_config.tracking_log_bucket,
```
blarghmatey, line 206:
```suggestion
                Bucket=config.tracking_log_bucket,
```

src/ol_orchestrate/ops/open_edx.py
```diff
@@ -30,70 +30,138 @@
 from ol_orchestrate.lib.file_rendering import write_csv
 
 
+class ListCoursesConfig(Config):
+    edx_client_id: str = Field(
```
blarghmatey, line 34: These `Field` attributes should be replaced with Pydantic `Field` instances

src/ol_orchestrate/ops/open_edx.py
```diff
@@ -627,10 +633,10 @@ def export_edx_courses(
             if task_status["state"] in {"Failed", "Canceled", "Retrying"}:
                 failed_exports.add(course_id)
     for course_id in successful_exports:
-        context.log.info("Moving course %s to %s", course_id, daily_extracts_dir)
+        config.log.info("Moving course %s to %s", course_id, daily_extracts_dir)
```
blarghmatey, line 636:
```suggestion
        context.log.info("Moving course %s to %s", course_id, daily_extracts_dir)
```

blarghmatey (changes requested):

blarghmatey (approved):

PR #787, opened 2023-08-03
1308 Dagster Pipeline to Retrieve and Stage edX.org Data Exports

quazi-h:
Closes #[1308](https://github.com/mitodl/hq/issues/1308)

# Description
This pipeline retrieves edX.org data exports stored by IRx, unpacks the compressed files, and stages the csv, tracking log, and course export files to their respective staging buckets on S3. This is done so that downstream tasks can process and ingest the data into our data lakehouse.

# How can this be tested?
Changes were tested by running the dagster pipeline locally.
Use a local config file ("/Users/qhoque/GIT/ol-data-platform/etc/edxorg_gcp.yaml") with google service account credentials to be able to access the IRx bucket.
`dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/retrieve_irx_edx_exports.py`

When running, the _files_to_sync_ param in the run config was updated in dagster for quicker iteration:
```ops:
  download_edx_data_exports:
    config:
      file_type: courses
      files_to_sync:
      - COLD/internal-2023-09-10.zip
      - COLD/internal-2023-09-26.zip
      - COLD/mitx-edx-events-2012-09-10.log.gz
      irx_edxorg_gcs_bucket: simeon-mitx-pipeline-main
  upload_edx_data_exports:
    config:
      bucket_prefix: pipeline-storage
      edx_irx_exports_bucket: ol-devops-sandbox
```

Files named "internal-YYYY-MM-DD.zip" contain csv, forum, and course data.
Files named "mitx-edx-events-YYYY-MM-DD.log.gz" contain tracking logs.

## Checklist:
- [ ] Update hardcoded local paths in code.
- [ ] Confirm that the proper credentials are being used for the qa and prod environments respectively.


src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -0,0 +1,168 @@
...
+            context.log.info(file)
+            try:
+                with zipfile.ZipFile(file, "r") as zippedFile:
+                    zippedFile.extractall(path=f"{exports_path}")
+            except zipfile.BadZipfile:
+                context.log.exception(
+                    "zipfile.BadZipfile: %s is not a zip file", file.name
+                )
+                bad_files.append(file.name)
+            context.log.info("Bad Zip Files: %s", bad_files)
```
quazi-h, line 110: Is there a better way to handle this error, and is it useful to keep a list of the offending zip files?

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -0,0 +1,168 @@
...
+            Path(file).unlink()
+            yield AssetMaterialization(
+                asset_key=f"edxorg_{file_type}/",
+                description=f"Export directory for IRx edx {file_type} data",
+                metadata={"bucket_path": MetadataValue.path(s3_path)},
+            )
+    yield Output(
+        f"s3://{config.edx_irx_exports_bucket}/{config.bucket_prefix}/",
+        "s3_upload_uri",
+    )
```
quazi-h, line 168: I was getting Dagster errors when I tried running the pipeline without yielding an Output or including an Out param in the decorator.

src/ol_orchestrate/ops/retrieve_edx_exports.py
blarghmatey: If we're going to do anything with them I'd say maybe upload them to a sub-path of the landing zone bucket to act as a dead-letter queue in case we need to manually intervene?

src/ol_orchestrate/definitions/edx/retrieve_irx_edx_exports.py
```diff
@@ -0,0 +1,140 @@
...
+file_regex = {
+    "courses": r"COLD/internal-\d{4}-\d{2}-\d{2}.zip$",
+    "logs": r"COLD/mitx-edx-events-\d{4}-\d{2}-\d{2}.log.gz$",
+}
+
+retrieve_edx_exports = Definitions(
+    resources={
+        "gcp_gcs": GCSConnection(
+            **load_yaml_config(
+                "/Users/qhoque/GIT/ol-data-platform/etc/edxorg_gcp.yaml"
```
blarghmatey, line 93: This needs to be changed back to `/etc/dagster/edxorg_gcp.yaml`

src/ol_orchestrate/definitions/edx/retrieve_irx_edx_exports.py
```diff
@@ -0,0 +1,140 @@
...
+            )["resources"]["gcp_gcs"]["config"]
+        ),
+        "s3": s3_resource,
```
blarghmatey, line 96: Please update to use the new Pythonic [`S3Resource`](https://docs.dagster.io/_apidocs/libraries/dagster-aws#dagster_aws.s3.S3Resource)

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -0,0 +1,168 @@
...
+    export_type: str = Field(
+        description="The subset of archive files to process", default=None
+    )
+
+
+class UploadConfig(Config):
+    edx_irx_exports_bucket: Optional[str] = Field(
+        description="The S3 bucket where files will be staged",
+        default="ol-devops-sandbox",
```
blarghmatey, line 36: The actual bucket that we will be using is `ol-data-lake-landing-zone-<env_stage>` where `env_stage` is either `qa` or `production`

src/ol_orchestrate/ops/retrieve_edx_exports.py
blarghmatey: For a first pass we can skip on that.

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -0,0 +1,168 @@
...
+    exports_path = f"{context.resources.exports_dir.path}/{config.export_type}"
+    context.log.info(exports_path)
+    for file in config.files_to_sync or []:
+        fname = file.removeprefix("COLD/")
+        Path(exports_path).joinpath(Path(fname).parent).mkdir(
+            parents=True, exist_ok=True
+        )
+        export_type = {
+            "courses": r"internal-\d{4}-\d{2}-\d{2}.zip$",
+            "logs": r"mitx-edx-events-\d{4}-\d{2}-\d{2}.log.gz$",
```
blarghmatey, line 66: As I was thinking through running the historical sync I believe it would be useful to extract the date for the given file so that we can use that as a path component for the upload target. This would be as simple as adding a capture group to the regex used here:
```suggestion
            "courses": r"internal-(\d{4}-\d{2}-\d{2}).zip$",
            "logs": r"mitx-edx-events-(\d{4}-\d{2}-\d{2}).log.gz$",
```

src/ol_orchestrate/ops/retrieve_edx_exports.py
quazi-h: Issue created to tackle this later

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -45,124 +46,200 @@ class UploadConfig(Config):
...
-        if re.match(file_match, fname):
+        file_path = f"{exports_path}/{file_name}"
+        if match := re.match(file_match, file_name):
             blob = bucket.get_blob(file)
-            blob.download_to_filename(f"{exports_path}/{fname}")
             context.log.info(blob.name)
             context.log.info(blob.size)
+            blob.download_to_filename(file_path)
+            context.log.info(file_path)
+            downloaded_files.append(
```
quazi-h, line 92: Now passing a list of files that were downloaded on to extract / upload ops.

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -45,124 +46,200 @@ class UploadConfig(Config):
...
-    if export_type == "courses":
-        exports_path = Path(f"{context.resources.exports_dir.path}/{export_type}")
-        context.log.info(exports_path)
-        zip_files = exports_path.glob("*.zip")
-        bad_files = []
-        for file in zip_files:
-            context.log.info(file)
+    extracted_files = []
+    bad_zip_files = []
+    for export_type, exports_path, file_name, file_date in file_list:
```
quazi-h, line 133: Instead of using glob to find zip files, I am just working off of the list provided by the download op.
I did this for two reasons - the [upstream changes to DailyResultsDir](https://github.com/mitodl/ol-data-platform/blob/7e2f9db4d4bd040096df1748abc1ae5f7b6012d8/src/ol_orchestrate/resources/outputs.py#L25) results in an extra subdirectory as the top level folder, and when testing, every invocation of "context.resources.exports_dir.path" between different ops resulted in a new root level directory so files were not being found, even when using glob or rglob. To solve this, I am maintaining a list of files that are currently being processed. Each item in this list contains 4 details about each file: export_type, export_path, file_name, and file_date. A new list is created within each op that only includes files that were successfully processed.

The second reason is that concurrent/parallel runs may collide with each other if rglob picks up matching files with the right extension but from a different run. In practice, I do not think we will run into this issue as we are running the pipeline on a list of files returned by a sensor or provided in the config. In the off chance that we ever have more than one run for the same type of export, I think it is better to be safe.

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -45,124 +46,200 @@ class UploadConfig(Config):
...
             try:
-                with zipfile.ZipFile(file, "r") as zippedFile:
+                with zipfile.ZipFile(f"{exports_path}/{file_name}", "r") as zippedFile:
                     zippedFile.extractall(path=f"{exports_path}")
+                    # Append the extracted archive's root to the exports_path
+                    (relative_path,) = zipfile.Path(zippedFile).iterdir()
+                    extracted_files.append(
+                        [
+                            export_type,
+                            f"{exports_path}/{relative_path.name}",
```
quazi-h, line 145: This appends the top level directory found in the archive of the zipped course files.
This is not necessary since we are using rglob in the upload op to search for files recursively, but I think it is useful for logging and tracking runs to be more specific about where files are being staged.

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -45,124 +46,200 @@ class UploadConfig(Config):
...
+                        ]
+                    )
             except zipfile.BadZipfile:
                 context.log.exception(
-                    "zipfile.BadZipfile: %s is not a zip file", file.name
+                    "zipfile.BadZipfile: %s is not a zip file", file_name
                 )
-                bad_files.append(file.name)
-            context.log.info("Bad Zip Files: %s", bad_files)
+                bad_zip_files.append([export_type, exports_path, file_name, file_date])
```
quazi-h, line 154: Stubbed a ticket for this:
https://github.com/mitodl/ol-data-platform/issues/858

src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -45,124 +46,200 @@ class UploadConfig(Config):
...
-                Key=s3_key,
-            )
-            Path(file).unlink()
-            yield AssetMaterialization(
-                asset_key=f"edxorg_{file_type}/",
-                description=f"S3 URI for IRx edx {file_type} upload",
-                metadata={"S3_URI": MetadataValue.path(s3_path)},
-            )
+    uploaded_files = []
+    for export_type, exports_path, file_name, file_date in file_list:
```
quazi-h, line 200: Updated the upload op to iterate over the downloaded file_list or extracted file_list. Most of the logic below this is the same.

src/ol_orchestrate/ops/retrieve_edx_exports.py
quazi-h: One question I do have is whether the DailyResultsDir change would affect us being able to pick up a run from a failed step and bring it to completion.

Issue #729, opened 2023-06-13
stage MITx Online tracking logs


rachellougee:
# Description/Context

MITx Online tracking log `raw__mitxonline__openedx__tracking_logs` is loaded into our data lake. It contains:

- [event](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#event-field) - all blank at the moment
- [event_source](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#event-field) e.g. browser
- [event_type](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#event-type-field) 
- [name](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#event-type-field) - it supersedes the `event_type` field if present
- source_type - file
- [username](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#username-field)
- [page](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#page-field)
- [context](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#context-member-fields-common-to-all-events) - all null at the moment
- [ip](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#ip-field) - blank for mobile events
- [host](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#host-field)
- environment
- [agent](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#agent-field) - e.g. Mozilla/5.0
- [accept_language](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#accept-language-field) - e.g. es-xl
- labels (null)
- [time](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#time-field)
- [referer](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/common_fields.html#referer-field)
- log_host  - ip-10-xx-xx

[...]

PR #792, opened 2023-08-08
stage MITx Online tracking log table, update to xPro staging tracking log

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/729

This adds `stg__mitxonline__openedx__tracking_logs__user_activity` as an incremental model for performance. It has dedupe logic because raw data contains duplicates. It also updates to `stg__mitxpro__openedx__tracking_logs__user_activity` to be consistent with mitxonline. 

# How can this be tested?

To test against production data, you will need to update your [profile.xml](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/profiles.yml#L46) with production host `mitol-ol-data-lake-production.trino.galaxy.starburst.io` due to resource limitation of our QA cluster to run large model (https://github.com/mitodl/ol-data-platform/issues/793 is created to solve this problem without updating profiles.yml)
 then run 
```
 dbt build --select stg__mitxonline__openedx__tracking_logs__user_activity --vars 'schema_suffix: YOURSCHEMA' --target dev_production 

 dbt build --select stg__mitxpro__openedx__tracking_logs__user_activity --vars 'schema_suffix: YOURSCHEMA' --target dev_production --full-refresh

```
Tests should all pass

### DIFF
--- a/src/ol_dbt/macros/extract_course_id_from_tracking_log.sql
+++ b/src/ol_dbt/macros/extract_course_id_from_tracking_log.sql
@@ -1,6 +1,7 @@
 {% macro extract_course_id_from_tracking_log() %}
     ---course ID format: course-v1:{org}+{course number}+{run_tag}
-    {% set course_id_regex = 'course-v(\d{1}):([^\/]+)\+([^\/]+)\+([^\/\]]+)' %}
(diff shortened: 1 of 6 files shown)


abeglova (approved):
LGTM

PR #794, opened 2023-08-11
updates to int combined users table to add new fields and new source

KatelynGit:
https://github.com/mitodl/hq/issues/1644

Added Micromasters as a source for the int combined users table and added 3 new fields to the table user_job_title, user_company, and user_industry.

# How can this be tested?
Run the following command against your schema or QA

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/_combined_models.yml
+++ b/src/ol_dbt/models/intermediate/combined/_combined_models.yml
@@ -32,6 +32,12 @@ models:
       platform
   - name: user_birth_year
     description: int, user's birth year
+  - name: user_job_title
+    description: str, user's job title
+  - name: user_company
+    description: str, user company
+  - name: user_industry
+    description: str, user job industry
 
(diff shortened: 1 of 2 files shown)


rachellougee (changes requested):
This PR needs some work, It's not just union MicroMasters users table.  MicroMasters is not really a platform, their users are already in edX.org and some are in MITx Online, which is why we didn't add it in `int__combined__users` in the first place.

 For https://github.com/mitodl/hq/issues/1644 request, I think it's looking for adding MicroMasters profile data to edx.org users, not adding MicroMasters as a source @pdpinch can confirm.  Also  `int__combined__users` table contains duplicate, I de-duped in `int__mitx__users`. In my opinion, we should add those MicroMasters fields to `int__mitx__users`.

pdpinch:
Yes I agree with @rachellougee. We should be enriching the `int_mitx_users` table with this additional profile information from MicroMasters.

KatelynGit:
a different PR was opened for this

PR #803, opened 2023-08-17
add 4 new fields to new xpro mart table marts__mitxpro_ecommerce_productlist

KatelynGit:

This change is to modify a new data mart table marts__mitxpro_ecommerce_productlist. Add the product parent run to the product data mart, but only for course-run products.
And Enhance the product mart with CMS data, specifically: duration, time_commitment, topics.

# How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
+++ b/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
@@ -55,6 +55,16 @@ with ecommerce_product as (
     order by product_id
 )
 
+, ecommerce_course_to_topics as (
+    select 
+        course_to_topics.course_id
+        , array_join(array_distinct(array_agg(coursetopic.coursetopic_name)), ', ') as coursetopic_name
+    from course_to_topics
+    left join coursetopic
+        on course_to_topics.coursetopic_id = coursetopic.coursetopic_id
+    group by course_to_topics.course_id
+)
+
 select
     'xPRO' as product_platform
     , ecommerce_product.product_id as eproductid
@@ -75,22 +85,20 @@ select
     , concat(programs.program_readable_id, '+', course_runs.courserun_tag) as product_parent_run_id
     , courses.cms_coursepage_duration as duration
     , courses.cms_coursepage_time_commitment as time_commitment
-    , coursetopic.coursetopic_name
+    , ecommerce_course_to_topics.coursetopic_names
 from ecommerce_product
 left join ecommerce_productversion_latest
     on ecommerce_product.product_id = ecommerce_productversion_latest.product_id
 inner join course_runs
     on ecommerce_product.courserun_id = course_runs.courserun_id
+inner join courses
+    on course_runs.course_id = courses.course_id
 left join coursesinprogram
     on course_runs.course_id = coursesinprogram.course_id
 left join programs
     on coursesinprogram.program_id = programs.program_id
-left join courses
-    on course_runs.course_id = courses.course_id
-left join course_to_topics
-    on course_runs.course_id = course_to_topics.course_id
-left join coursetopic
-    on course_to_topics.coursetopic_id = coursetopic.coursetopic_id
+left join ecommerce_course_to_topics
+    on course_runs.course_id = ecommerce_course_to_topics.course_id
 where ecommerce_product.product_type = 'course run'
 
 union all
@@ -115,7 +123,7 @@ select
     , null as product_parent_run_id
     , programs.cms_programpage_duration as duration
     , programs.cms_programpage_time_commitment as time_commitment
-    , null as coursetopic_name
+    , null as coursetopic_names
 from ecommerce_product
 left join ecommerce_productversion_latest
     on ecommerce_product.product_id = ecommerce_productversion_latest.product_id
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
```diff
@@ -52,11 +72,25 @@ select
...
 left join ecommerce_productversion_latest
     on ecommerce_product.product_id = ecommerce_productversion_latest.product_id
 inner join course_runs
     on ecommerce_product.courserun_id = course_runs.courserun_id
+left join coursesinprogram
+    on course_runs.course_id = coursesinprogram.course_id
+left join programs
+    on coursesinprogram.program_id = programs.program_id
+left join courses
+    on course_runs.course_id = courses.course_id
```
rachellougee, line 89:
```suggestion
inner join courses
    on course_runs.course_id = courses.course_id
left join coursesinprogram
    on course_runs.course_id = coursesinprogram.course_id
left join programs
    on coursesinprogram.program_id = programs.program_id

```

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
```diff
@@ -52,11 +72,25 @@ select
...
+left join course_to_topics
+    on course_runs.course_id = course_to_topics.course_id
+left join coursetopic
```
rachellougee, line 92: We probably should add `coursetopic_name` to course_to_topics so that one less table to join in the mart, but its ok for now

rachellougee (commented):
Just one suggestion, otherwise it looks good

rachellougee:
The dups are because in `int__mitxpro__coursesinprogram` course can be associated with multiple programs, then `product_parent_run_id` would be multiple. Is Mart expected to have multiple `product_parent_run_id`  for a product?

rachellougee:
Maybe this is a question for @cachob and Kate. Since the course can associate with multiple programs, e.g. course-v1:xPRO+LASERx1+R10, are we showing two rows of different product_parent_run_id - `program-v1:xPRO+LASERx+R10` and `program-v1:xPRO+LASER-SL+R10` in BI? 

Also, this course https://xpro.mit.edu/cms/pages/188/edit/ is in draft, https://xpro.mit.edu/checkout/?product=336 link isn't working, is it expected?

KatelynGit:
> The dups are because in `int__mitxpro__coursesinprogram` course can be associated with multiple programs, then `product_parent_run_id` would be multiple. Is Mart expected to have multiple `product_parent_run_id` for a product?

I think that's ok since there are courses associated with multiple programs but will wait for a response

rachellougee:
@KatelynGit I asked @cachob so it's ok to show two `product_parent_run_id` if there are multiple programs. If that's the case, you will need to update the test because it's no longer unique on `product_id` in your test. You might also want to confirm the `link` field with Bon, according to him, the SL course should not have a checkout page.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
KatelynGit: OK we'll keep this in mind

KatelynGit:
I've made the necessary changes to handle the fact that we can have a course associated with multiple programs and multiple topics per course.  I talked to Bon and he said no one should be able to checkout the pages however the data should have checkout links so I left them in there.

rachellougee (commented):
What do you think of grouping multiple topics into one instead of duplicating it for each topic? I am not sure it's interesting to show multiple topics in separate row for the same product, https://github.com/mitodl/hq/issues/1908#issuecomment-1645817579 suggests `topics` if I read it correctly? The same might apply to product_parent_run_id. Just my thoughts, not requested changes

Just curious what `no one should be able to checkout the pages however the data should have checkout links` means. What is the purpose of displaying a checkout link if some no longer work? Just a question, not a change request.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
```diff
@@ -35,6 +55,17 @@ with ecommerce_product as (
...
 )
 
+, ecommerce_course_to_topics as (
+    select
+        course_to_topics.course_id
+        , array_join(array_distinct(array_agg(coursetopic.coursetopic_name)), ', ') as coursetopic_name
+    from course_to_topics
+    left join coursetopic
+        on course_to_topics.coursetopic_id = coursetopic.coursetopic_id
+    group by course_to_topics.course_id
```
rachellougee, line 65:
```suggestion
    select
        course_to_topics.course_id
        , array_join(array_agg(coursetopic.coursetopic_name order by coursetopic.coursetopic_name), ', ') as coursetopic_names
    from course_to_topics
    inner join coursetopic
        on course_to_topics.coursetopic_id = coursetopic.coursetopic_id
    group by course_to_topics.course_id
```

No need to array_distinct as it's unique on coursetopic_name and course_id, and inner join is proper here

rachellougee (approved):
Thanks for the change. Looks good, just one comment per latest change, but no need to re-review.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
KatelynGit: I've removed array_distinct and changed that to an inner join. Thanks for all your help Rachel!

PR #807, opened 2023-08-22
feat: Add logic to re-process normalized tracking logs for IRx

blarghmatey:

Closes https://github.mit.edu/IR/simeon-internal/issues/42

The logic used by IRx expects the tracking logs to still contain JSON data in the context and event fields. This converts the stringified fields into JSON data, exports the converted records and re-uploads to the `logs` prefix that IRx is pulling their data from in S3.

# How can this be tested?
This is largely relying on an existing, functional pipeline with the logic reversed. To test we can deploy and run a backfill in QA before executing in production.

### DIFF
--- a/src/ol_orchestrate/ops/normalize_logs.py
+++ b/src/ol_orchestrate/ops/normalize_logs.py
@@ -28,11 +28,21 @@ class LoadFilesConfig(Config):
     s3_token: str = Field(
         description="STS token indicating the role assumed for the IAM credentials",
     )
+    path_prefix: str = Field(
+        default="logs", description="Path prefix where the target logs are located"
+    )
 
 
 class WriteFilesConfig(Config):
     tracking_log_bucket: str = Field(
         description="S3 bucket where tracking logs are stored",
+    )
+    source_path_prefix: str = Field(
+        default="logs",
+        description="Path prefix where the logs are being loaded from",
+    )
+    destination_path_prefix: str = Field(
+        default="valid", description="Path prefix where the logs are being written to"
     )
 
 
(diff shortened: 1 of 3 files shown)


src/ol_orchestrate/ops/normalize_logs.py
```diff
@@ -204,7 +262,7 @@ def write_file_to_s3(
             context.resources.s3.upload_file(
                 Filename=local_file_name,
                 Bucket=config.tracking_log_bucket,
-                Key=f"valid/{log_date}{local_file_name}",
+                Key=f"{config.destination_pat}/{log_date}{local_file_name}",
```
quazi-h, line 265:
```suggestion
                Key=f"{config.destination_path_prefix}/{log_date}{local_file_name}",
```

src/ol_orchestrate/ops/normalize_logs.py
```diff
@@ -136,6 +147,50 @@ def transform_log_data(context: OpExecutionContext) -> Nothing:
...
+    """
+    with context.resources.duckdb.get_connection() as conn:
+        columns = ["event", "context"]
+        for col in columns:
+            conn.execute(f"ALTER TABLE tracking_logs ALTER {col} TYPE JSON")
+            # extract JSON from VARCHAR for context and event fields
+            update_query = f"UPDATE tracking_logs SET {col} = json({col}) WHERE json_valid({col});"  # noqa: S608, E501
+            conn.execute(update_query)
+        # convert integer timestamps to datetime
         conn.execute(
```
quazi-h, line 194: Do we still want to transform the timestamp data here? Wasn't Jon just using the data as-is, without our normalization.

blarghmatey: It should actually be a no-op, but yeah, I'll remove that line to avoid potential error cases.

PR #808, opened 2023-08-22
Add openedx_user_id xPro users

rachellougee:

N/A
Similar to https://github.com/mitodl/ol-data-platform/pull/781

This PR adds `stg__mitxpro__openedx__mysql__auth_user` and openedx_user_id to int__mitxpro__users to be consistent with MITx Online. It also removes the old model that we get from BigQuery so we don't rely on BigQuery for our xPro open edx users

# How can this be tested?

dbt build on +int__mitxpro__users


abeglova (approved):
👍

PR #810, opened 2023-08-24
fix dbt test

rachellougee:

N/A

This PR fixes dbt test warning/failure on production

```
dbt_expectations_expect_compound_columns_to_be_unique_int__mitxpro__user_courseactivity_problemcheck_user_username__courserun_readable_id__useractivity_problem_id__useractivity_timestamp (models/intermediate/mitxpro/_int_mitxpro__models.yml)
Got 61 results, configured to fail if >10

Warning in test accepted_values_stg__salesforce__opportunity_opportunity_business_type__B2C__B2B__B2X (models/staging/salesforce/_stg_salesforce__models.yml)
Got 2 results, configured to warn if != 0
```
For `int__mitxpro__user_courseactivity_problemcheck`, those are due to some events in 2019 don't have nanoseconds in timestamp, so we don't have precise data to determine which one happen first, we will have to look at other field to figure out the order.

# How can this be tested?

```
 dbt test --select int__mitxpro__user_courseactivity_problemcheck --target production
```
should now pass

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -1373,7 +1373,10 @@ models:
         "useractivity_video_id", "useractivity_video_currenttime", "useractivity_timestamp"]
 
 - name: int__mitxpro__user_courseactivity_problemcheck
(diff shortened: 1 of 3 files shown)


abeglova (approved):
LGTM

PR #813, opened 2023-08-29
xPRO coupon data mart

KatelynGit:

This will add one new table the contains redeemed and unredeemed coupons as well as b2b coupons for xPRO. The new table name is - int__mitxpro__ecommerce_allcoupons

# How can this be tested?


src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,50 @@
...
+    description: number, either dollar amount or percentage (between 0 and 1) discount
+  - name: coupon_type
+    description: string, one of 'single-use' or 'promo'. Promo coupon codes can be
+      used multiple times
+  - name: discount_source
+    description: string, one of "credit_card", "purchase_order", "marketing, "sales",
+      "staff". "staff" coupons are used to allow staff to enroll in courses. "marketing"
+      coupons are promos for discounts generated by marketing. "credit_card", "purchase_order"
+      and "sales" correspond to different ways that companies can pay for vouchers
+      for their employees to enroll in courses or programs
```
pdpinch, line 31: nitpick - can the description of the possible values be in the same order as the list of values? 

I don't really care which one you change, it's just weird to start with "credit_card" in the list and then have it show up in the middle of the paragraph.

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,50 @@
...
+  - name: discount_type
+    description: one of percent-off or dollars-off
```
pdpinch, line 33: I don't know how much it matters, but this field should always be grouped with discount amount, since discount amount is ambiguous without it.

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,50 @@
...
+  - name: activated_on
+    description: timestamp, activation timestamp for the coupon
+  - name: expires_on
+    description: timestamp, expiration timestamp for the coupon
+  - name: coupon_source_table
+    description: string, the table that the coupon data is coming from. Options are
+      "ecommerce_coupon", "b2bcoupon".
```
pdpinch, line 40: I think we should try to explain this a bit more, although I can't write the text off the top of my head.

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,50 @@
...
+  - name: b2bcoupon_id
+    description: int, foreign key to b2becommerce_b2bcoupon. A coupon that was redeemed
+      for the order.
+  - name: coupon_id
+    description: int, foreign key referencing ecommerce_coupon
```
pdpinch, line 45: Am I correct in understanding that this field would only be populated when `coupon_source_table` is set to `ecommerce_coupon` ?

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: I've put these next to each other in the table since they're related

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,50 @@
...
+  - name: activated_on
+    description: timestamp, activation timestamp for the coupon
+  - name: expires_on
+    description: timestamp, expiration timestamp for the coupon
+  - name: coupon_source_table
+    description: string, the table that the coupon data is coming from. Options are
+      "ecommerce_coupon", "b2bcoupon".
+  - name: b2bcoupon_id
+    description: int, foreign key to b2becommerce_b2bcoupon. A coupon that was redeemed
+      for the order.
```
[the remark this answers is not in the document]
KatelynGit: I removed that sentence it didn't quite make sense

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: that's correct

KatelynGit:
> @KatelynGit Is this PR same as #753? This is just adding intermediate?

Yes, that one was a bit of a mess so it was redesigned per these [comments](https://github.com/mitodl/hq/issues/1803#issuecomment-1667825049)

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,52 @@
...
+    - not_null
+  - name: coupon_name
+    description: string, human readable name for the coupon payment
+  - name: coupon_created_on
+    description: timestamp, specifying when the coupon was initially created
+    tests:
+    - not_null
+  - name: payment_transaction
+    description: string, string that identifies the payment invoice for coupon purchases
+      by companies
```
rachellougee, line 20: It might worth adding description to indicate which fields are only apply to ecommerce coupon, I see there are some fields are marked as null from b2b

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,52 @@
...
+  - name: discount_amount
+    description: number, either dollar amount or percentage (between 0 and 1) discount
```
rachellougee, line 22: This is not null field, might worth adding a test for that

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allcoupons.sql
```diff
@@ -0,0 +1,106 @@
...
+, b2b_coupon_fields as (
+    select
+        b2bcoupon_coupon_code as coupon_code
+        , b2bcoupon_name as coupon_name
+        , b2bcoupon_created_on as coupon_created_on
+        , null as payment_transaction
+        , b2bcoupon_discount_percent as discount_amount
+        , null as coupon_type
+        , null as discount_source
+        , null as discount_type
```
rachellougee, line 64: This might not be clear with discount_amount for b2bcoupon. If we rename `b2bcoupon_discount_percent` to  `discount_amount`, then we should indicate it's percent-off here. As Peter pointed out, discount_amount +discount_type should go together, it might worth combining two together with proper formatting, that's how we do for mitxonline. e.g. $750.00 off, 10.00% off.

rachellougee (commented):
Left small comments. Do we need a field to indicate if it's redeemed or not? Otherwise it looks good

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
pdpinch: I think that's worth mentioning in the description.

pdpinch:
Will there be a separate mart or report with the status of coupons, i.e., whether they have been redeemed, when, and by whom?

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: done

KatelynGit:
> Left small comments. Do we need a field to indicate if it's redeemed or not? Otherwise it looks good
> Will there be a separate mart or report with the status of coupons, i.e., whether they have been redeemed, when, and by whom?

Redeemed status will be in the orders mart table which can easily be joined to here if needed.   I left it out of here so that we could make this table a little simpler at a coupon level at least.

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: added this

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: added this

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allcoupons.sql
KatelynGit: made the suggested changed to the discount_amount field

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allcoupons.sql
```diff
@@ -0,0 +1,123 @@
...
+        , b2bcoupon_created_on as coupon_created_on
+        , null as payment_transaction
+        , null as coupon_type
+        , null as discount_source
+        , b2bcoupon_activated_on as activated_on
+        , b2bcoupon_expires_on as expires_on
+        , 'b2bcoupon' as coupon_source_table
+        , b2bcoupon_id
+        , null as coupon_id
+        , cast((b2bcoupon_discount_percent * 100) as varchar (9)) || '% off' as discount_amount
```
rachellougee, line 88: Same comment as before. I am wondering why choosing varchar (9) here as both these fields are decimal(38,4)

rachellougee (approved):
LGTM. Just small comments about the casting, but its not blocker

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allcoupons.sql
```diff
@@ -0,0 +1,123 @@
...
+        , null as b2bcoupon_id
+        , coupon.coupon_id
+        , case
+            when ecommerce_couponpaymentversion.couponpaymentversion_discount_type = 'dollars-off'
+                then
+                    (
+                        '$'
+                        || cast(
+                            ecommerce_couponpaymentversion.couponpaymentversion_discount_amount
+                            as varchar (100)
```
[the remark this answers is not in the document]
KatelynGit: I've changed this is just varchar

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allcoupons.sql
KatelynGit: I've also changed this to varchar. It can't be a decimal because I've added text to the field.

PR #814, opened 2023-08-29
Fix for issue 1908 to populate parent run id's more consistently

KatelynGit:

changed the join to remove int__mitxpro__coursesinprogram and join directly to programs to pull the product parent run id field

# Screenshots (if appropriate):
see relevant ticket

# How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
+++ b/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
@@ -78,7 +78,7 @@ select
         , cast(ecommerce_product.product_id as varchar (50))
         , '">', course_runs.courserun_readable_id, '</a>'
     ) as link
-    , programs.program_readable_id as product_parent_run_id
+    , concat(programs.program_readable_id, '+', course_runs.courserun_tag) as product_parent_run_id
     , courses.cms_coursepage_duration as duration
     , courses.cms_coursepage_time_commitment as time_commitment
     , ecommerce_course_to_topics.coursetopic_name as coursetopic_names
--- a/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
+++ b/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
@@ -23,11 +23,6 @@ with ecommerce_product as (
     from {{ ref('int__mitxpro__program_runs') }}
 )
 
-, coursesinprogram as (
-    select *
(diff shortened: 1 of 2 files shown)


rachellougee (commented):
Query looks fine, but we need to update the test https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml#L50 since now there won't be multiple `product_parent_run_id` if we are using program in courses model

rachellougee:
Also can you confirm if this is changed? because the new code does not have run id on `program-v1:xPRO+SysEngx`

> Product Parent Run: ID which includes the Run #. For Example, the Product Parent Run of course-v1:xPRO+SysEngx4+R18 is program-v1:xPRO+SysEngx+R18.

KatelynGit:
> Query looks fine, but we need to update the test https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml#L50 since now there won't be multiple `product_parent_run_id` if we are using program in courses model

Good catch I've changed the tests to remove product_parent_run_id from the unique test

KatelynGit:
> Also can you confirm if this is changed? because the new code does not have run id on `program-v1:xPRO+SysEngx`
> 
> > Product Parent Run: ID which includes the Run #. For Example, the Product Parent Run of course-v1:xPRO+SysEngx4+R18 is program-v1:xPRO+SysEngx+R18.

Per Bon's screenshot it looks like he wasn't anticipating run id's as part of that field so I left if off.

src/ol_dbt/models/marts/mitxpro/marts__mitxpro_ecommerce_productlist.sql
```diff
@@ -83,7 +78,7 @@ select
         , cast(ecommerce_product.product_id as varchar (50))
         , '">', course_runs.courserun_readable_id, '</a>'
     ) as link
-    , concat(programs.program_readable_id, '+', course_runs.courserun_tag) as product_parent_run_id
+    , programs.program_readable_id as product_parent_run_id
```
rachellougee, line 81: We still need courserun_tag to product_parent_run_id, the requirement for this is still the same

rachellougee (changes requested):
I would suggest double-checking the requirement for `product_parent_run_id` before implementing the changes, so there is no confusing here. Per @cachob , this is still the same

> Product Parent Run: ID which includes the Run #. For Example, the Product Parent Run of course-v1:xPRO+SysEngx4+R18 is program-v1:xPRO+SysEngx+R18.

rachellougee (approved):

PR #825, opened 2023-09-05
feat: add platform model and update course/program models

asadali145:

Fixes the last part of https://github.com/mitodl/mitxpro/issues/2698

Adds staging and intermediate models for courses_platform in xPro. Also, it updates the staging and intermediate models course/program.

# How can this be tested?

Run the following commands against QA | Production:
`dbt build -s +int__mitxpro__platforms`
`dbt build -s +int__mitxpro__courses`
`dbt build -s +int__mitxpro__programs`
`dbt build -s +stg__mitxpro__app__postgres__courses_platform`
`dbt build -s +stg__mitxpro__app__postgres__courses_course`
`dbt build -s +stg__mitxpro__app__postgres__courses_program`

I have used schema_suffix: `add_platform_model`

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__programs.sql
+++ b/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__programs.sql
@@ -14,6 +14,7 @@ select
     , programs.program_is_live
     , programs.program_readable_id
     , programs.program_is_external
+    , programs.platform_id
     , cms_programs.cms_programpage_description
     , cms_programs.cms_programpage_subhead
     , cms_programs.cms_programpage_catalog_details
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -844,6 +844,20 @@ models:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_id", "program_id", "ecommerce_order_id"]
 
+- name: int__mitxpro__platforms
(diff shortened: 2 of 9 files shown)


src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__programs.sql
```diff
@@ -14,6 +14,7 @@ select
     , programs.program_is_live
     , programs.program_readable_id
     , programs.program_is_external
+    , programs.platform_id
```
rachellougee, line 17: Ideally we should add platform_name to `int__mitxpro__programs` and `int__mitxpro__courses` since courses_platform is just a lookup table. Otherwise we would need join courses_platform when using courses/programs intermediate models.

rachellougee (approved):
LGTM

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__programs.sql
asadali145: Please help me understand this one. Should we add the platform name in the course/program tables? The issue explicitly mentions adding a new table for the platforms just for reusability purposes.

rachellougee: I am not sure what reusability the issue is referring to. For intermediate model, it's mostly denormalized tables where we join multiple tables together so it can be used by self or by mart.
https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/docs/staging_guidelines.md#what-is-the-staging-layer

I don't think it matters that much, but just one less table to join in the mart when using courses/programs as end user would want the name not internal id.

PR #834, opened 2023-09-14
add platform name to tables

KatelynGit:
part of https://github.com/mitodl/hq/issues/1908

adding platform_name to int__mitxpro__programs and int__mitxpro__courses since it's more useful than the id

# How can this be tested?


rachellougee (approved):
LGTM

PR #836, opened 2023-09-14
added mitxpro coupon redemption table to staging

KatelynGit:

Creates the new stg__mitxpro__app__postgres__b2becommerce_b2bcouponredemption table in staging

# How can this be tested?
Run dbt build --select staging.mitxpro if you have all models, otherwise add + in front of staging.mitxpro to run upstream models


abeglova (changes requested):
There is a uniqueness constrain on coupon and order together. Can you add it to the test?
https://github.com/mitodl/mitxpro/blob/master/b2b_ecommerce/models.py#L253

Otherwise lgtm

abeglova (approved):

PR #837, opened 2023-09-14
refactor: Upgrade Healthchecks resource to use Pythonic interface

blarghmatey:
N/A

Migrate the Healthchecks IO resource in Dagster to use their new Pydantic-based configuration interface.

# How can this be tested?
Launch the Dagster dev server targeting the open edX ops graph to ensure that the resources are loaded and configured.


quazi-h (approved):
LGTM

PR #844, opened 2023-09-19
add mitxpro b2b coupon redemption to intermediate

KatelynGit:

This adds the mitxpro b2b coupon redemption table to intermediate

# How can this be tested?

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -10,17 +10,19 @@ models:
     description: int, primary key representing a b2b coupon redemption
     tests:
     - not_null
-    - unique 
+    - unique
   - name: b2border_id
     description: int, foreign key to b2becommerce_b2border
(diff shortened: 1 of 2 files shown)


abeglova (approved):

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -2,6 +2,32 @@
...
+  columns:
+  - name: b2bcouponredemption_id
+    description: int, primary key representing a b2b coupon redemption
+    tests:
+    - not_null
+    - unique
+  - name: b2border_id
+    description: int, foreign key to b2becommerce_b2border
+  - name: b2bcoupon_id
+    description: int, foreign key representing a b2b coupon
```
rachellougee, line 17: `b2border_id` and `b2bcoupon_id` should both be not_null since its foreign key. I would add that to staging as well to be clear

KatelynGit: added not null to both fields

PR #845, opened 2023-09-22
Change format for five columns xpro prodline mart

KatelynGit:

All timestamps on the table were changed to dates and list price was changed to have 2 decimal places instead of 4.

# How can this be tested?


abeglova (changes requested):
As discussed on slack:
This will not fix the timestamp representation in bi - you should use from_iso8601_timestamp( in bi to fix this

If the time part of the timestamps are not useful than we should truncate them in the staging layer but it seems like the timestamp for the enrollment window should be useful. You should check with Bon

abeglova (approved):
Approved

KatelynGit:
I checked with Bon he mentioned the admin users are concerned about the dates not really the times so I think we are ok here

PR #849, opened 2023-09-27
S3 object sync

blarghmatey:
https://github.com/mitodl/hq/issues/1592

- Creates a pipeline for keeping the edx.org program credential reports synchronized to our own S3 bucket so that they can be ingested to the data lake via Airbyte.
- Creates a generic S3 sensor for identifying new files to be retrieved
- Creates a set of generic `op`s for copying, downloading, or uploading objects between S3 buckets or local storage.

# How can this be tested?
This has been tested by running the pipeline locally with the command `dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/sync_program_credential_reports.py` from the root of the repository.

# Additional Context
There is a corresponding change in ol-infrastructure that is needed to support this workflow defined in https://github.com/mitodl/ol-infrastructure/pull/1784

### DIFF
--- a/src/ol_orchestrate/definitions/edx/normalize_tracking_logs.py
+++ b/src/ol_orchestrate/definitions/edx/normalize_tracking_logs.py
@@ -1,4 +1,4 @@
-import os  # noqa: INP001
+import os
 from datetime import UTC, datetime  # type: ignore  # noqa: PGH003
 from functools import partial
 from typing import Literal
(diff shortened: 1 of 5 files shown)


quazi-h (approved):
New pipeline looks good and I was able to launch it locally and it was able to successfully upload and download a small subset of files. The sensor does pick up all the files when first launching it and the download step was running smoothly with a longer list of files to process. Other changes here to the sensor and refactor of how the sensor is called from a pipeline also looks good. With minor changes to the local config in `edxorg_gcp.yaml`, I was able to kick off the edx_gcs_courses pipeline locally.

PR #886, opened 2023-10-20
fix: Ensure that all edX files get uploaded

blarghmatey:
Addresses https://github.mit.edu/IR/simeon-internal/issues/44

Because the logic is iterating over the ResultsDir resource, there is no guarantee that the files are actually present in the resource as created in that step. This changes to accept a list of DagsterPath objects that get iterated over and recursively uploaded to be more explicit about passing of files.

# How can this be tested?
Run the open edX pipelines and ensure that all of the generated CSVs get properly uploaded to S3

# Additional Context
The results_dir resource definition was recently refactored to use the new Pythonic interface. That likely led to some implicit changes in how it functioned, leading to the implicit passing of files via the resource not functioning properly anymore.

### DIFF
--- a/src/ol_orchestrate/jobs/open_edx.py
+++ b/src/ol_orchestrate/jobs/open_edx.py
@@ -44,13 +44,15 @@ from ol_orchestrate.resources.sqlite_db import sqlite_db_resource
 def edx_course_pipeline():
     course_list = list_courses()
     extracts_upload = upload_extracted_data(
-        write_course_list_csv(edx_course_ids=course_list),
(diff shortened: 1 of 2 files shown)


quazi-h (approved):
Changes look good to me, I didn't realize the unintended consequences of refactoring the results_dir resource definition. Should I stub out a ticket to add smoke tests or integration tests? It should be straightforward to mock out our resources and some test data.

blarghmatey:
Yes, a ticket to capture the scope of some tests would be useful.

PR #894, opened 2023-10-24
Refactor file path for edx.org logs staged on S3

quazi-h:
Bugfix related to #1309 [Ingest CSV and tracking log data from edx.org via Airbyte](https://github.com/mitodl/hq/issues/1309)

Ideally, the retrieve_edx_exports upload op code should be refactored to include a dated subdirectory, which ensures we are matching the same structure of the other tracking log buckets. This will reduce the number of conditional statements needed and improve readability. When that change happens, this will need to be refactored to update the glob pattern for edx.org logs.

Most of our tracking log sources have a dated directory with a set of log files under the `logs` directory matching this glob pattern: `bucket/logs/date/logfile.gz` (mitxonline, mitxonline, xpro : qa + prod)
However, the edx.org logs are currently stored at this path: `bucket/edxorg-raw-data/logs/logfile.gz` (edxorg: qa + prod)

This PR updates the retrieve_irx_edx_exports pipeline to reproduce the same directory structure with a subdirectory for each unique date. The s3_paths and references to the file path have been updated where relevant.

# How can this be tested?
This was tested by running the pipeline locally via:
`dagster dev -d src/ol_orchestrate -f src/ol_orchestrate/definitions/edx/normalize_tracking_logs.py`
I ran this locally against QA files for each deployment: mitx, mitxonline, xpro, edxorg to confirm that the download steps were completing successfully. I commented out the SQL executions in the transformation code in order to speed up testing.


src/ol_orchestrate/ops/retrieve_edx_exports.py
```diff
@@ -238,11 +236,7 @@ def upload_files(
...
-        f"s3://{config.edx_irx_exports_bucket}/{config.bucket_prefix}/",
-        "s3_upload_bucket",
-    )
-    yield Output(
-        uploaded_files,
-        "file_list",
-    )
+                yield DynamicOutput(
+                    file_date,
+                    mapping_key="file_dates",
```
blarghmatey, line 241: The mapping key should be the actual value. With a fixed string it will treat all of the outputs as the same thing.

PR #895, opened 2023-10-25
add two new fields for tax amounts to int__mitxpro__ecommerce_order

KatelynGit:
https://github.com/mitodl/mitxpro/issues/2771

add order_tax_amount and order_total_price_paid_plus_tax to the int__mitxpro__ecommerce_order table

# How can this be tested?

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
+++ b/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
@@ -31,7 +31,7 @@ select
     , orders.order_tax_rate
     , orders.order_tax_rate_name
     , (orders.order_total_price_paid * (orders.order_tax_rate / 100)) as order_tax_amount
-    , (orders.order_total_price_paid * (orders.order_tax_rate / 100))
+    , (orders.order_total_price_paid * (orders.order_tax_rate / 100)) 
     + orders.order_total_price_paid as order_total_price_paid_plus_tax
 from orders
 left join couponredemption on orders.order_id = couponredemption.order_id
--- a/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
+++ b/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
@@ -13,11 +13,6 @@ with orders as (
     from {{ ref('stg__mitxpro__app__postgres__ecommerce_couponversion') }}
 )
 
-, couponpaymentversion as (
-    select *
-    from {{ ref('stg__mitxpro__app__postgres__ecommerce_couponpaymentversion') }}
-)
-
 select
     orders.order_id
     , orders.order_state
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
```diff
@@ -25,6 +30,10 @@ select
     , orders.order_tax_country_code
     , orders.order_tax_rate
     , orders.order_tax_rate_name
+    , (orders.order_total_price_paid * (orders.order_tax_rate / 100)) as order_tax_amount
+    , (orders.order_total_price_paid * (orders.order_tax_rate / 100))
+    + orders.order_total_price_paid as order_total_price_paid_plus_tax
 from orders
 left join couponredemption on orders.order_id = couponredemption.order_id
 left join couponversion on couponredemption.couponversion_id = couponversion.couponversion_id
+left join couponpaymentversion on couponversion.couponpaymentversion_id = couponpaymentversion.couponpaymentversion_id
```
rachellougee, line 39: Is this join used here? I have added this table in separate PR to add a few fields

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
```diff
@@ -25,6 +30,10 @@ select
     , orders.order_tax_country_code
     , orders.order_tax_rate
     , orders.order_tax_rate_name
+    , (orders.order_total_price_paid * (orders.order_tax_rate / 100)) as order_tax_amount
+    , (orders.order_total_price_paid * (orders.order_tax_rate / 100))
+    + orders.order_total_price_paid as order_total_price_paid_plus_tax
```
rachellougee, line 35: Should this be added to `stg__mitxpro__app__postgres__ecommerce_order` and then add to intermediate so they are available in staging as well?

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
KatelynGit: I removed it. good catch I think this should be ok now.

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
KatelynGit: I was thinking we probably only need it in intermediate since we probably wouldn't direct users to staging ever for these columns

rachellougee: Sorry that I didn't mean that for end users, these fields are directly calculated from raw order table, not something that can only be created during intermediate so In my opinion it should belong to staging. But that said, up to you.

rachellougee (approved):

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_order.sql
KatelynGit: If we end up needing it i'll add it there too

Issue #885, opened 2023-10-19
Create combined enrollments mart


rachellougee:
# Description/Context

Create a combined enrollments mart from different platforms - MITx, xPro, bootcamps, including common fields for enrollments, orders, coupons and certificates. So we can have a unified mart for enrollment details from our OL apps, and eventually replace query like https://bi.odl.mit.edu/queries/1142/source

rachellougee:
In testing https://github.com/mitodl/ol-data-platform/pull/897, I realized that enrollments to orders are not always 1 to 1, meaning that there could be multiple fulfilled orders for the same enrollment, e.g. MicroMasters and Bootcamp, so we will need to figure out how to present the order information in enrollments if we have to fit them into one table

- platform (MITx, xPro, Bootcamp)
- courserunenrollment_is_active
- courserunenrollment_created_on
- courserunenrollment_enrollment_mode
- courserunenrollment_enrollment_status
-  user_id
-  courserun_id
-  courserun_title
-  courserun_readable_id
-  user_username
-  user_email
-  user_full_name

These below fields could be multiple fulfilled orders
-   order_reference_number
-   order_total_price_paid
-   order_created_on
-   order_line_price
-   coupon_code
-   coupon_redeemed_on
-   coupon_discount_amount_text
-   payment_transaction_id
-   payment_transaction_bill_to_address_state
-   payment_transaction_bill_to_address_country

For Bootcamp, the payment can be split into multiple orders (installments), order_total_price_paid would be SUM of orders for the same application here, i am not entirely sure the order_reference_number and payment_* would be useful for Bootcamp

@pdpinch @KatelynGit Any idea how we should handle these orders in enrollment table?

rachellougee:
Per discussion with @pdpinch @KatelynGit  in data platform meeting, we decided to 
- For MicroMasters orders, pick the most recent order/payment if users paid for same course run multiple times
- For xPro/MITx Online, there should always be one order per enrollment 
- For Bootcamps, sum of all the orders (installments) paid for the same application/enrollment, but we need to wait for @KatelynGit to communicate with Finance team https://github.mit.edu/bootcamps/bootcamps-issues/issues/31 on what to do with order_reference_number and cybersource authorization code payments made from multiple orders

KatelynGit:
I've followed up with Bryan will let you know when I hear back thanks

rachellougee:
[...]

PR #897, opened 2023-10-27
Create combined enrollment mart and upstream changes

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/885

- Creates `marts__combined_course_enrollment_detail` with enrollments, certificates, fulfilled/refunded orders, coupons and cybersource payment details from MITx, xPro and Bootcamps. It contains the following fields
    platform 
    courserunenrollment_is_active
    courserunenrollment_created_on
    courserunenrollment_enrollment_mode
    courserunenrollment_enrollment_status
    user_id
    courserun_id
    courserun_title
    courserun_readable_id
    user_username
    user_email
    user_full_name
    courseruncertificate_created_on
    courseruncertificate_url
    order_state
    order_reference_number
    order_total_price_paid
    order_created_on
    line_price
    coupon_code
    coupon_redeemed_on
    coupon_discount_amount
    payment_transaction_id
    payment_authorization_code
    payment_method
    payment_bill_to_address_state
    payment_bill_to_address_country
- Creates `int__micromasters__orders` for legacy DEDP orders
- A few upstream changes

# How can this be tested?

dbt build --select +int__bootcamps__ecommerce_receipt 
dbt build --select +int__bootcamps__ecommerce_order 
dbt build --select +int__micromasters__orders 
dbt build --select +int__mitxonline__ecommerce_transaction 
dbt build --select +int__mitxonline__ecommerce_order
dbt build --select +int__mitxpro__ecommerce_couponpaymentversion 
dbt build --select +int__mitxpro__ecommerce_coupon
dbt build --select +int__mitxpro__ecommerce_receipt 
dbt build --select +int__mitxpro__ecommerce_order 
dbt build --select +int__mitxpro__ecommerce_line 
dbt build --select +int__mitxpro__ecommerce_allcoupons 

or 

just run `dbt build --select +marts__combined_course_enrollment_detail` its going to take a while

Note if you see warning about something like `dbt_expectations_expect_table_row_count_to_equal_other_table_int__mitxpro__ecommerce_productversion`, that's because the above commands don't run the sibiling of these intermediate models,  you can just run that warned intermediate model, the test should pass

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
@@ -32,6 +32,27 @@ with lines as (
     select * from {{ ref('int__mitxonline__ecommerce_product') }}
 )
 
+, discounts as (
+    select * from {{ ref('stg__mitxonline__app__postgres__ecommerce_discount') }}
+)
+
+, discountredemptions as (
+    select
+        *
+        , row_number() over (partition by order_id order by discountredemption_timestamp desc) as row_num
+    from {{ ref('stg__mitxonline__app__postgres__ecommerce_discountredemption') }}
+
+)
+
+---- small amounts of duplicated payments for the same order. For those, we pick the most recent one
+, payments as (
+    select
+        *
+        , row_number() over (partition by order_id order by transaction_created_on desc) as row_num
+    from {{ ref('stg__mitxonline__app__postgres__ecommerce_transaction') }}
+    where transaction_type = 'payment'
+)
+
 select
     orders.order_id
     , orders.order_state
@@ -44,12 +65,26 @@ select
     , users.user_email
     , lines.line_id
     , lines.product_version_id
+    , intermediate_products_view.product_price
     , intermediate_products_view.product_type
     , intermediate_products_view.product_id
     , intermediate_products_view.courserun_id
     , intermediate_products_view.programrun_id
+    , discounts.discount_code
+    , discounts.discount_amount_text
+    , discountredemptions.discountredemption_timestamp
+    , payments.transaction_authorization_code as payment_authorization_code
+    , payments.transaction_payment_method as payment_method
+    , payments.transaction_readable_identifier as payment_transaction_id
+    , payments.transaction_bill_to_address_state as payment_bill_to_address_state
+    , payments.transaction_bill_to_address_country as payment_bill_to_address_country
 from lines
 inner join orders on lines.order_id = orders.order_id
 inner join users on orders.order_purchaser_user_id = users.user_id
 inner join versions on lines.product_version_id = versions.version_id
 inner join intermediate_products_view on versions.version_object_id = intermediate_products_view.product_id
+left join payments
+    on orders.order_id = payments.order_id and payments.row_num = 1
+left join discountredemptions
+    on orders.order_id = discountredemptions.order_id and discountredemptions.row_num = 1
+left join discounts on discountredemptions.discount_id = discounts.discount_id
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
@@ -36,6 +36,9 @@ with lines as (
     select * from {{ ref('stg__mitxonline__app__postgres__ecommerce_discount') }}
 )
 
+---- this table doesn't have constraint so apply additional logic here as there should be only one discount
+---- that's actually applied to an order
+
 , discountredemptions as (
     select
         *
--- a/src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
@@ -1,0 +1,237 @@
+with mitx_enrollments as (
+    select * from {{ ref('int__mitx__courserun_enrollments') }}
+)
+
+, mitxpro_enrollments as (
+    select * from {{ ref('int__mitxpro__courserunenrollments') }}
+)
+
+, bootcamps_enrollments as (
+    select * from {{ ref('int__bootcamps__courserunenrollments') }}
+)
+
+, mitx_certificates as (
+    select * from {{ ref('int__mitx__courserun_certificates') }}
+)
+
+, mitxpro_certificates as (
+    select * from {{ ref('int__mitxpro__courserun_certificates') }}
+)
+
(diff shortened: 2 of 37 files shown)


rachellougee:
I mark it 'needs review' but I will need to run a final test with xPro orders. I don't know how long its going to take to fix airbyte issue, but code can be reviewed

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
```diff
@@ -0,0 +1,237 @@
...
+        , mitx_certificates.courseruncertificate_created_on
+        , mitx_certificates.courseruncertificate_url
+        , mitxonline_completed_orders.order_state
+        , mitxonline_completed_orders.order_reference_number
+        , mitxonline_completed_orders.order_total_price_paid
+        , mitxonline_completed_orders.order_created_on
+        , mitxonline_completed_orders.product_price as line_price
+        , mitxonline_completed_orders.discount_code as coupon_code
+        , mitxonline_completed_orders.discountredemption_timestamp as coupon_redeemed_on
+        , mitxonline_completed_orders.discount_amount_text as coupon_discount_amount
```
KatelynGit, line 88: for discount percentages and dollar amounts I think we should be rounding to 2 decimal places

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
```diff
@@ -0,0 +1,237 @@
...
+        , mitx_enrollments.courserun_title
+        , mitx_enrollments.courserun_readable_id
+        , mitx_enrollments.user_username
+        , mitx_enrollments.user_email
+        , mitx_enrollments.user_full_name
+        , mitx_certificates.courseruncertificate_created_on
+        , mitx_certificates.courseruncertificate_url
+        , mitxonline_completed_orders.order_state
+        , mitxonline_completed_orders.order_reference_number
+        , mitxonline_completed_orders.order_total_price_paid
```
KatelynGit, line 83: same thing for dollar amounts- ideally we'd round to 2 decimal places

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
```diff
@@ -0,0 +1,237 @@
...
+)
+
+, micromasters_users as (
+    select * from {{ ref('int__micromasters__users') }}
+)
+
+, mitxpro_completed_orders as (
+    select * from {{ ref('int__mitxpro__ecommerce_order') }}
+    where order_state in ('fulfilled', 'refunded')
+)
```
KatelynGit, line 52: For some sources we're including 'partially refunded' order states and then for other sources that status is being excluded. Just a question really why is that? why isn't it the same across sources?

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
```diff
@@ -32,6 +32,27 @@ with lines as (
...
+, discounts as (
+    select * from {{ ref('stg__mitxonline__app__postgres__ecommerce_discount') }}
+)
+
+, discountredemptions as (
+    select
+        *
+        , row_number() over (partition by order_id order by discountredemption_timestamp desc) as row_num
+    from {{ ref('stg__mitxonline__app__postgres__ecommerce_discountredemption') }}
+
```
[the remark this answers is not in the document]
rachellougee: This is necessary as there used to be a bug on MITx Online https://github.com/mitodl/hq/issues/1765. This has been fixed, but the underly table doesn't have any constraint, it relies on application to prevent it. So I added here to prevent any future bug just in case.

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
rachellougee: If we need to rounding to 2 decimal, it should happen in staging, not mart. But any reason why we need to round this? Is it for display?

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
rachellougee: The only difference is from MicroMasters as that's the only place where we have partially refunded state. There are 3 orders ('MM-prod-4979', 'MM-prod-1651', 'MM-prod-8794') are 'partially refunded' in MicroMasters and have payment receipt. So I think we should include these in mart unless you disgree?

I don't know why there is different in MicroMasters order, @pdpinch probably can tell you more. But as part of creating this mart, we should review and treat each application differently and bring together data that makes sense for this mart. Just my opinion

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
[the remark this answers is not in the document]
KatelynGit: That makes sense glad you knew about the bug

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
KatelynGit: I believe it'd be ideal for display in Tableau

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
KatelynGit: I agree we can include them in the mart. Thanks for the info!

KatelynGit:
So far looks good but i assume I'll need to review/test again after the airbyte issue is resolved

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
rachellougee: We talked about it yesterday, so I made changes to round discount amount, price and total_price_paid to be 2 decimal places in staging. There are probably other fields should be rounded to be consistent but that's out of scope for this PR. We will need to make those changes when we add fields to mart.

rachellougee:
@KatelynGit I've rounded discount_amount, order_total_price_paid, line_price to 2 decimal places on staging. I've also ran full test for this PR. Please take a look, thank you

KatelynGit:
The changes to the decimal places look good. I did discover an issue while testing your table that was un-related to your code.  There is an issue with the courserun_title field on stg__bootcamps__app__postgres__courses_courserun it contains dates and countries and even some html tags and a lot of really messy data in that field. I'm not sure what's causing it but we may need to open a separate ticket regarding that issue.

PR #900, opened 2023-10-27
fix dbt test error for certificate_url

rachellougee:

N/A

This fixes the following dbt test error noted in https://pipelines.odl.mit.edu/runs/170b1684-c3d4-4d07-b028-3ae8c1de38d3
```
Failure in test not_null_int__mitxonline__courserun_certificates_courseruncertificate_url (models/intermediate/mitxonline/_int_mitxonline__models.yml)
Got 66 results, configured to fail if >10
```
The `courseruncertificate_url` in `int__mitxonline__courserun_certificates` could be null for revoked certificates, thus need to remove not_null test

# How can this be tested?

Run ` dbt test --select int__mitxonline__courserun_certificates --target production` should now be no error


KatelynGit (approved):
LGTM no issues

PR #909, opened 2023-10-31
add email optout table to int

KatelynGit:
https://github.com/mitodl/hq/issues/1545

adds table int__mitxonline__openedx__mysql__bulk_email_optout to intermediate from staging

# How can this be tested?
Run dbt build --select intermediate.mitxonline if you have all models, otherwise add + in front of intermediate.mitxonline to run upstream models

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
@@ -2,6 +2,25 @@
 version: 2
 
 models:
+- name: int__mitxonline__openedx__mysql__bulk_email_optout
+  columns:
+  - name: email_optout_id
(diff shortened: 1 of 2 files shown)


rachellougee (commented):
I have question related to this PR
- Do we need to bring it to intermediate if it's select * from staging? It seems like staging should be enough for Jon according to Tobias's comment in https://github.com/mitodl/hq/issues/1545#issuecomment-1756108758
- If we do need to bring it to intermediate, should we add username and email to this table? Otherwise, this table just contains `openedx_user_id` and `courserun_readable_id`, it's not super useful as intermediate IMO. 

@KatelynGit @pdpinch

KatelynGit:
@rachellougee @pdpinch I hadn't seen that comment from Tobias but it does look like we only needed this table in stage so I'm going to close this

Issue #917, opened 2023-11-06
Stage raw__irx__edxorg__s3__tracking_logs


rachellougee:
# Description/Context

edX.org tracking log data is loaded into `raw__irx__edxorg__s3__tracking_logs` in https://github.com/mitodl/hq/issues/1309, we should stage it so that we can compare it with `int__edxorg__mitx_user_courseactivities` from IRx data.

The fields in this table should be the same as https://github.com/mitodl/ol-data-platform/issues/729

Issue #793, opened 2023-08-09
Permission issue running dbt code using production schema in QA cluster


rachellougee:
# Description/Context

As data analysts, we would like to run dbt code on the `dev_qa` target but use the production schema because QA data is often not useful

- database: ol_data_lake_qa
- schema: ol_warehouse_production_dev
- host: mitol-ol-data-lake-qa-0.trino.galaxy.starburst.io

This is the error I got when running dbt code against dev_qa
`
  TrinoExternalError(type=EXTERNAL, name=HIVE_METASTORE_ERROR, message="User: arn:aws:sts::610119931565:assumed-role/data-lake-query-engine-role-qa/trino-session is not authorized to perform: glue:CreateDatabase on resource: arn:aws:glue:us-east-1:610119931565:database/ol_warehouse_production_rlougee_staging because no identity-based policy allows the glue:CreateDatabase action (Service: AWSGlue; Status Code: 400; Error Code: AccessDeniedException; Request ID: ee46b622-8933-492b-b914-c0b358b6b3c8; Proxy: null)", query_id=20230809_185544_08799_3cv42)
`

## Expected Behavior

Run dbt code against `dev_qa` using schema `ol_warehouse_production_dev` successfully

rachellougee:
@blarghmatey FYI, this is the issue we talked about.

pdpinch:
@blarghmatey @rachellougee has this been fixed?

rachellougee:
This can be closed. We can now use `dev_production` to create separate dev schema in production database as oppose to `dev_qa`

PR #919, opened 2023-11-08
create stg__edxorg__s3__tracking_logs__user_activity

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/917

Creates staging model stg__edxorg__s3__tracking_logs__user_activity as incremental 

# How can this be tested?

To test against production data, you will need to update your [profile.xml](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/profiles.yml#L46) to `mitol-ol-data-lake-production.trino.galaxy.starburst.io` due to resource limitation of our QA cluster to run large model (https://github.com/mitodl/ol-data-platform/issues/793 is created to solve this problem without updating profiles.yml)

Run  `dbt build --select stg__edxorg__s3__tracking_logs__user_activity`

### DIFF
--- a/src/ol_dbt/models/staging/edxorg/_edxorg_sources.yml
+++ b/src/ol_dbt/models/staging/edxorg/_edxorg_sources.yml
@@ -547,3 +547,53 @@ sources:
     - name: course run start date
       description: str, the start date of the course run in the format 'YYYY-MM-DD
         HH:mm:ss Z' in UTC
+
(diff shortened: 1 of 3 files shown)


quazi-h (approved):
I was able to run the dbt build against the production host and the build completed successfully after 41 minutes. I was able to query the table in starburst and all of the fields have valid data. Yaml and SQL code changes here look good as well.

PR #928, opened 2023-11-15
add new mitxonline email optout table to intermediate

KatelynGit:
https://github.com/mitodl/hq/issues/1545

adds table int__mitxonline__openedx__mysql__bulk_email_optout to intermediate

# How can this be tested?
Run dbt build --select intermediate.mitxonline if you have all models, otherwise add + in front of intermediate.mitxonline to run upstream models

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
@@ -2,7 +2,7 @@
 version: 2
 
 models:
-- name: int__mitxonline__openedx__mysql__bulk_email_optout
+- name: int__mitxonline__bulk_email_optin
   columns:
   - name: openedx_user_id
     description: int, unique ID for each user on the MITx Online open edX platform
@@ -19,9 +19,8 @@ models:
   - name: courserun_title
     description: str, title of the course run
   - name: email_opted_in
-    description: int, if this field is 0 then we've recieved an email optout otherwise
-      it will be 1
-
+    description: int, if this field is 0 then we've recieved an email optout otherwise it will be 1
+  
 - name: int__mitxonline__ecommerce_basketdiscount
   columns:
   - name: basketdiscount_id
--- a/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
@@ -19,8 +19,9 @@ models:
   - name: courserun_title
     description: str, title of the course run
   - name: email_opted_in
-    description: int, if this field is 0 then we've recieved an email optout otherwise it will be 1
-  
+    description: int, if this field is 0 then we've recieved an email optout otherwise
+      it will be 1
+
 - name: int__mitxonline__ecommerce_basketdiscount
   columns:
   - name: basketdiscount_id
--- a/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
@@ -23,7 +23,7 @@ models:
     description: str, name chosen by user and used for login
   - name: user_email
     description: str, user email associated with their account
-
+  
 - name: int__mitxonline__ecommerce_basketdiscount
   columns:
   - name: basketdiscount_id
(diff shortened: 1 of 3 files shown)


KatelynGit:
I will fix the changes unrelated to this ticket before the merge

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__openedx__mysql__bulk_email_optout.sql
```diff
@@ -0,0 +1,30 @@
...
+    from {{ ref('stg__mitxonline__app__postgres__courses_courserun') }}
+)
+
+, users as (
+    select *
+    from {{ ref('int__mitxonline__users') }}
+)
+
+select
+    email_optout.email_optout_id
```
blarghmatey, line 19: I'm curious what the utility of this field is?

blarghmatey (commented):
The semantics of the email opt out table are the opposite of the opt in report that edx.org produces. In that report there is a boolean that indicates that a user has "opted in" to emails. The table that we are using is for when users "opt out" of emails. I think that for this table we may want to pull the full list of users and attach a boolean field that indicates whether the user has (implicitly) agreed to receive emails.

In other words, something like (pseudo-SQL following):
```
SELECT
 users.user_username
 , users.user_email
 , CASE WHEN email_optout.email_optout_id IS NULL THEN 1 ELSE 0 AS email_opted_in
FROM users LEFT JOIN email_optout ON email_optout.user_id = users.user_id
```

blarghmatey:
The query used by edX to generate the opt in table is here: https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/user_api/management/commands/email_opt_in_list.py#L202-L230

src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml
```diff
@@ -1030,6 +1052,7 @@ models:
     description: str, the full URL to the certificate on MITx Online
     tests:
     - unique
+    - not_null
```
rachellougee, line 1055: @KatelynGit Is this added by accident or any reason to change it as part of this PR? This test will fail as for revoked certificates this will be null. I removed the not_null because of that.

KatelynGit: I can remove that. Not sure how it got added

src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__openedx__mysql__bulk_email_optout.sql
KatelynGit: I think the only utility of that column is basically it's use as a primary key of the table. So if we wanted to reference this table in another table it'd be easy and we'd only have to add one column. Or maybe potential use for things like indexing or one column to determine uniqueness if needed instead of multiple which might be simpler for tests.

KatelynGit:
> The semantics of the email opt out table are the opposite of the opt in report that edx.org produces. In that report there is a boolean that indicates that a user has "opted in" to emails. The table that we are using is for when users "opt out" of emails. I think that for this table we may want to pull the full list of users and attach a boolean field that indicates whether the user has (implicitly) agreed to receive emails.
> 
> In other words, something like (pseudo-SQL following):
> 
> ```
> SELECT
>  users.user_username
>  , users.user_email
>  , CASE WHEN email_optout.email_optout_id IS NULL THEN 1 ELSE 0 AS email_opted_in
> FROM users LEFT JOIN email_optout ON email_optout.user_id = users.user_id
> ```

I've made the changes you described above. Hopefully it all looks good.

blarghmatey (approved):

rachellougee:
Sorry I have to comment on this since I got notification. The table name is `int__mitxonline__openedx__mysql__bulk_email_optout` which doesn't seem to in line with our other intermediate tables. Also it's opposite to the column in this table. Am I the only one think this way?

blarghmatey:
Thanks for calling that out Rachel. I agree that it's worth confirming to the model naming that we use for intermediate.

KatelynGit:
> Sorry I have to comment on this since I got notification. The table name is `int__mitxonline__openedx__mysql__bulk_email_optout` which doesn't seem to in line with our other intermediate tables. Also it's opposite to the column in this table. It also doesn't have any test. Am I the only one think this way?
> 
> You can ignore this if you disagree

I've modified the table name to better align with current int naming

PR #931, opened 2023-11-21
Add DBT Grant Configs for Trino Roles and Permissions

### LINKED ISSUE #924: Programatically manage permissions for Trino tables
# User Story

- As a data consumer, I want to be able to access the tables that are shared with me
- As a data platform owner, I want to be able to manage data access via source control

# Description/Context

The primary way that we share data with consumers is by granting permissions to users and roles on different tables or groupings of tables. In order to ensure that those permissions are properly managed, we want them to be defined in code. This enables code review processes, versioning, auditing, etc. By having the roles and grants in code, it also reduces any possible future migration work.

# Acceptance Criteria

- [ ] There is a method for defining roles as code that get populated in Starburst
- [ ] There is a method for defining permissions on those roles that is executed as part of table creation workflows
- [ ] The existing role definitions and permissions are reflected as code.

# Plan/Design

There are two main interfaces that we have for managing these permissions.
- We can implement SQL statements for granting the permissions to be executed, either via the [`grants`](https://docs.getdbt.com/reference/resource-configs/grants) interface or an additional pipeline step.
- We can use the Trino API to programatically set permissions either through explicit role definitions, or by applying tags to tables as part of the pipeline execution.

There may be other methods that we can or should use for managing these permissions, but these are the two I am aware of from an initial investigation. Relevant documentation is available [here](https://docs.starburst.io/starburst-galaxy/security/access-control.html)

quazi-h:
Added grant configs for accountadmin, read_only_production, business_intelligence, production analyst, reverse_etl, and finance_access

Closes #924 

The permissions we grant to different roles allow us to share data with data consumers. Role specific permissions in Trino are currently managed through [Starburst](https://mitol.galaxy.starburst.io/role). We have observed tags and permissions periodically disappearing from Starburst because of how dbt manages table definitions. When a table is deleted and recreated, Starburst loses the reference to the table during the rebuild and there is no reference to maintain the tag relationships.

To address this issue, we are defining roles and database permissions as code within the dbt project config and schema and table configs. This will ensure permissions are properly managed, versioned, and can be audited and reviewed more easily.

# How can this be tested?
I tested this by running dbt build locally against a separate schema and then looking at the roles/permissions granted on those tables. 

```
# build models
dbt run --select open_learning --var 'schema_suffix: qhoque' --target dev_qa
```
In starburst, check privileges on marts__mitxonline_user_profiles, confirm grant statements in dbt logs.

## Checklist:
- [x] There is a method for defining roles as code that get populated in Starburst
- [x] There is a method for defining permissions on those roles that is executed as part of table creation workflows
- [x] The existing role definitions and permissions are reflected as code.

### DIFF
--- a/src/ol_dbt/dbt_project.yml
+++ b/src/ol_dbt/dbt_project.yml
@@ -60,7 +60,7 @@ models:
       format: "'PARQUET'"
     +required_docs: true
     +grants:
-      select: ['accountadmin', 'business_intelligence', 'read_only_production', 'production_analyst']
+      select: ['accountadmin', 'read_only_production', 'production_analyst']
       insert: ['accountadmin', 'production_analyst']
       update: ['accountadmin', 'production_analyst']
       delete: ['accountadmin', 'production_analyst']
@@ -77,13 +77,13 @@ models:
       +materialized: table
       +schema: intermediate
       +grants:
-        select: ['read_only_production']
+        select: ['read_only_production', 'business_intelligence']
         create: ['production_analyst']
     marts:
       +materialized: table
       +schema: mart
       +grants:
-        select: ['read_only_production', 'finance_access', 'reverse_etl']
+        select: ['read_only_production', 'business_intelligence', 'reverse_etl']
         create: ['production_analyst']
     sample:
 
--- a/src/ol_dbt/dbt_project.yml
+++ b/src/ol_dbt/dbt_project.yml
@@ -64,27 +64,22 @@ models:
       insert: ['accountadmin', 'production_analyst']
       update: ['accountadmin', 'production_analyst']
       delete: ['accountadmin', 'production_analyst']
-      create: ['accountadmin']
-      alter: ['accountadmin']
-      drop: ['accountadmin']
+      all privileges: ['accountadmin']
     staging:
       +materialized: table
       +schema: staging
       +grants:
         select: ['read_only_production']
-        create: ['production_analyst']
     intermediate:
       +materialized: table
       +schema: intermediate
       +grants:
         select: ['read_only_production', 'business_intelligence']
-        create: ['production_analyst']
     marts:
       +materialized: table
       +schema: mart
       +grants:
         select: ['read_only_production', 'business_intelligence', 'reverse_etl']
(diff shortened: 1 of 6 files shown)


src/ol_dbt/dbt_project.yml
```diff
@@ -59,15 +59,33 @@ models:
     +properties:
       format: "'PARQUET'"
     +required_docs: true
+    +grants:
+      select: ['accountadmin', 'business_intelligence', 'read_only_production', 'production_analyst']
```
blarghmatey, line 63: I'm not sure that we want to grant select to business_intelligence on _all_ tables. Probably only intermediate and mart?

blarghmatey (commented):
Besides my one comment this should be good to merge. We can extend and tweak as we go.

src/ol_dbt/dbt_project.yml
```diff
@@ -59,15 +59,33 @@ models:
...
       +materialized: table
       +schema: intermediate
+      +grants:
+        select: ['read_only_production']
+        create: ['production_analyst']
     marts:
       +materialized: table
       +schema: mart
+      +grants:
+        select: ['read_only_production', 'finance_access', 'reverse_etl']
```
blarghmatey, line 86: I don't think we want to give finance access to _all_ of the marts, do we? @KatelynGit

KatelynGit: @blarghmatey probably not they only really need access to the following tables: marts__combined_course_enrollment_detail 
marts__mitxpro_ecommerce_productlist

blarghmatey:
As far as conditionally applying grants, it seems like it would need to be set in the model config files using Jinja based on the target

src/ol_dbt/dbt_project.yml
quazi-h: ah thanks, the starburst policies for finance is incorrect then, it has "ol_data_lake_production.ol_warehouse_production_mart.*" under scope.
https://mitol.galaxy.starburst.io/role/r-2991408023/policies

blarghmatey (approved):

PR #935, opened 2023-11-27
updates to prod line mart table and the intermediate courses table

KatelynGit:
https://github.com/mitodl/hq/issues/2425

I've added the short program code field to stg__mitxpro__app__postgres__courses_program, stg__mitxpro__app__postgres__courses_course, int__mitxpro__courses, int__mitxpro__program_runs and added short program code to marts__mitxpro_ecommerce_productlist which is a field that finance currently uses and that can be seen in report 1006 that Jayke utilizes. I've also updated some of the field names in the mart to better match the source.

# How can this be tested?
dbt build --select +marts__mitxpro_ecommerce_productlist


src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
```diff
@@ -1159,6 +1159,9 @@ models:
     description: boolean, whether the course is on an external platform
     tests:
     - not_null
+  - name: position_in_program
```
quazi-h, line 1162: This issue from a few weeks ago mentions that we should be moving away from **position_in_program** and replacing it with `ProgramRequirement`. I'm not sure if that applies here as well since Finance was asking for that field specifically. I was looking for other examples where[ ProgramRequirement was being mentioned](https://github.com/search?q=repo%3Amitodl%2Fol-data-platform+path%3A%2F%5Esrc%5C%2Fol_dbt%5C%2F%2F+programrequirement&type=code) but I wasn't sure how to derive the same position_in_program value.

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -9,19 +9,24 @@ models:
     description: string, defaulted to xPro
     tests:
     - not_null
-  - name: eproduct_id
+  - name: product_id
```
quazi-h, line 12: I was wondering if these fields belong in staging or intermediate rather than in marts. Since Finance only has access to marts and they require the use of these fields, they have to go into the marts table, right?

quazi-h (approved):
Overall changes look good and I was able to run dbt locally and confirmed the new position_in_program field in int__mitxpro__courses as well as the product_id and product_readable_id in marts__mitxpro_ecommerce_productlist. I added two different comments about whether the fields are in the correct tables but Rachel may chime in with more feedback.

rachellougee (commented):
I left a few questions and comments, but feel free to ignore if you think otherwise.

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: It's one of the fields they use in the where clause for this report https://bi.odl.mit.edu/queries/1006/source but honestly I'll reach out to them and ask about it. It seems like it shouldn't be used.

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
KatelynGit: yes since finance uses it for their reporting and they are currently only using mart tables we've added it there too but it's definitely useful in staging and intermediate as well

KatelynGit:
> Overall changes look good and I was able to run dbt locally and confirmed the new position_in_program field in int__mitxpro__courses as well as the product_id and product_readable_id in marts__mitxpro_ecommerce_productlist. I added two different comments about whether the fields are in the correct tables but Rachel may chime in with more feedback.

Thanks Quazi!

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -9,19 +9,24 @@ models:
...
     tests:
     - not_null
     - unique
+  - name: productversion_readable_id
+    description: string, the readable_id field from the product object
   - name: product_type
     description: string, readable product type
+  - name: department_code
+    description: str, the portion of the readable product id that contains the department
+      code
```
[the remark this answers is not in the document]
KatelynGit: good catch I'll double check and see if we need a code modification here as well

src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
KatelynGit: I'm going to remove that field from the int table it seems like we shouldn't be using it

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -50,3 +55,9 @@ models:
     description: boolean, Public product is purchasable through the bulk form at /ecommerce/bulk
   - name: platform_name
     description: str, name of the platform
+  - name: b2c_price
+    description: numeric, the total product price for Boeing products and the price
+      discounted by ten percent for non-Boeing products
```
pdpinch, line 60: This is pretty confusing. Maybe it needs some more context about how it's used for forecasting and why it's different from other prices, and why there is a calculated discount.

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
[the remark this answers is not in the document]
KatelynGit: I've changed the name and description of this field to "short program code" to better capture what it represents to the finance team. I've also added examples to the description. And I've added the field to staging and int layers as well as the mart.

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
KatelynGit: I've removed this field after talking to finance it appears this is no longer needed for forecasting.

src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -50,3 +55,9 @@ models:
...
+  - name: b2b_price
+    description: numeric, the total product price for Boeing products and the price
```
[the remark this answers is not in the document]
KatelynGit: I've removed this field after talking to finance it appears this is no longer needed for forecasting.

PR #952, opened 2023-12-20
create mart for mitxonline video engagement

rachellougee:

https://github.com/mitodl/hq/issues/3222

Creating `marts__mitxonline_video_engagements` for learners video events in a course, so that we can use this mart to create the video view reports indicated in the issue
Adding two fields to `int__mitxonline__course_structure`

# How can this be tested?

 dbt build --select int__mitxonline__course_structure
 dbt build --select marts__mitxonline_video_engagements

### DIFF
--- a/src/ol_dbt/models/marts/mitxonline/_marts__mitxonline__models.yml
+++ b/src/ol_dbt/models/marts/mitxonline/_marts__mitxonline__models.yml
@@ -189,3 +189,64 @@ models:
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_username", "courserun_readable_id", "courseactivity_date"]
+
+- name: marts__mitxonline_video_engagements
+  description: MITx Online learners video engagements - play, pause, seek or stop
+    video action
+  columns:
+  - name: courserun_readable_id
+    description: str, Open edX Course ID formatted as course-v1:{org}+{course code}+{run_tag}.
+    tests:
+    - not_null
+  - name: user_username
+    description: str, username of the open edX user
+    tests:
+    - not_null
+  - name: user_email
+    description: str, user email on MITx Online
+  - name: user_full_name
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/marts/mitxonline/_marts__mitxonline__models.yml
KatelynGit, line 1: Which columns create a unique record for this new table?  I didn't see a unique test

rachellougee: So it would be `user_username`+ `courserun_readable_id`, `useractivity_video_id`+ `video_event_type`+`video_currenttime`+`video_old_time`+`video_new_time` to consider seek_video events. It's almost every field in this new model to make it unique. Do you think its useful to add? I can add that

KatelynGit (approved):
I ran the dbt builds and there were no errors or warnings everything looked good. The table looked good as well when I ran queries against it.  I mostly just had a question- do you think an event like load_video would be useful to include? I assume not but I was thinking about whether people would want to see videos loaded but not played or something like that

src/ol_dbt/models/marts/mitxonline/_marts__mitxonline__models.yml
KatelynGit: that's a lot of different fields/complexity maybe we can leave it off

rachellougee:
>  I mostly just had a question- do you think an event like load_video would be useful to include? I assume not but I was thinking about whether people would want to see videos loaded but not played or something like that

@KatelynGit Thanks for the review. To answer your question about load_video, I looked over insights documentation about video engagement, it doesn't seem like they consider solely `load_video` as engagement or view, so I didn't add that. I don't know if there is use case for that, but I can add it. For load_video, most fields will be null except video_id.

Issue #782, opened 2023-07-31
Unable to identify `openedx_user_id` for MITx Online users


rachellougee:
# Description/Context

There are some MITx Online app users that we can't match with open edx users by username in https://github.com/mitodl/ol-data-platform/pull/781
```
select app_users.username,app_users.email, app_users.is_active, app_users.created_on
from  "ol_data_lake_production"."ol_warehouse_production_raw"."raw__mitxonline__app__postgres__users_user" as app_users
left join "ol_data_lake_production"."ol_warehouse_production_raw"."raw__mitxonline__openedx__mysql__auth_user" as openedx_users  on app_users.username = openedx_users.username
where openedx_users.id is null
order by app_users.created_on 
```
The mismatched username between the app and openedx seems to be either one of these cases 
- usernames contain non-ASCII characters (e.g. ë) 
   - e.g https://mitxonline.mit.edu/admin/users/user/11212/change/ and https://courses.mitxonline.mit.edu/admin/auth/user/584/change/
- mitxonline username contains an underscore, but the open edx username contains a dot
   - e.g. https://courses.mitxonline.mit.edu/admin/auth/user/279/change/ vs https://mitxonline.mit.edu/admin/users/user/281/change/
- No open edx user record 
   - e.g. https://mitxonline.mit.edu/admin/users/user/60986/change/

Looking into how we should handle these cases

pdpinch:
@rachellougee, please update this list, and we'll try to get it fixed once and for all.

rachellougee:
@pdpinch I've updated the spreadsheet with two lists: 

-  54 app users can't match based on either username or email https://docs.google.com/spreadsheets/d/1QaJqTHK_Qugsuu2puG6MFKT1XvijYQJR/edit#gid=1908621716
    - While generating this list, I noted that there could be two app users linked to the same open edx user based on either username or email. e.g. https://mitxonline.mit.edu/admin/users/user/17078/change/, https://mitxonline.mit.edu/admin/users/user/9560/change/, https://courses.mitxonline.mit.edu/admin/auth/user/9491/change/
- 97 app users can't be matched base on username https://docs.google.com/spreadsheets/d/1QaJqTHK_Qugsuu2puG6MFKT1XvijYQJR/edit#gid=1608772514

I checked a couple of these users. One user has never logged and there no account on open edx. The other one seems to be retired user who email is obscured on open edx. Not sure how we should handle these retired user cases

pdpinch:
[...]

PR #975, opened 2024-01-16
populate openedx_user_id for mitxonline app users

rachellougee:

Part of https://github.com/mitodl/ol-data-platform/issues/782

populate `openedx_user_id` in `int__mitxonline__users` and `marts__mitxonline_course_certificates` based on matching username or email between app and open edx. 

There are sill some app users who don't have `openedx_user_id` in `int__mitxonline__users`, but there should be none flowing in `marts__mitxonline_course_certificates` since the only user in this table where openedx_user_id is null can be populated based on email match.

### How can this be tested?

```
dbt build --select +marts__mitxonline_course_certificates
```


KatelynGit (approved):
dbt build looked good no issues. and descriptions and sql looked good signing off

PR #977, opened 2024-01-17
add receipt_authorization_code and receipt_transaction_id to the marts__combined__orders table

KatelynGit:
https://github.com/mitodl/hq/issues/634

This ticket adds Cybersource Auth codes and Transaction IDs to the Ecommerce order mart.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
@@ -46,6 +46,8 @@ with bootcamps__ecommerce_order as (
         , mitxpro__ecommerce_order.order_purchaser_user_id
         , mitxpro__ecommerce_order.order_total_price_paid
         , mitxpro__b2becommerce_b2border.b2border_total_price
+        , mitxpro__ecommerce_order.receipt_authorization_code
+        , mitxpro__ecommerce_order.receipt_transaction_id
         , coalesce(mitxpro__ecommerce_allorders.coupon_id, mitxpro__ecommerce_allorders.b2bcoupon_id) as coupon_id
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -46,6 +46,8 @@ with bootcamps__ecommerce_order as (
         , mitxpro__ecommerce_order.order_purchaser_user_id
         , mitxpro__ecommerce_order.order_total_price_paid
         , mitxpro__b2becommerce_b2border.b2border_total_price
+        , mitxpro__ecommerce_order.receipt_authorization_code
+        , mitxpro__ecommerce_order.receipt_transaction_id
```
rachellougee, line 50: Is there reason why we don't add these two fields for b2b orders? I see that `req_reference_number` is extracted in [int__mitxpro__ecommerce_allorders](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql#L80), so there should be transaction_id and auth_code as well

e.g.
```
SELECT json_query(b2breceipt_data, 'lax $.transaction_id' omit quotes) , json_query(b2breceipt_data, 'lax $.auth_code' omit quotes)
FROM "ol_data_lake_production"."ol_warehouse_production_intermediate"."int__mitxpro__b2becommerce_b2breceipt" 
where b2border_id = 253
```

rachellougee (commented):
The code and test ran fine. But looking at data, it seems that we don't populate these two fields for b2b orders in xPro, see comment below. Is that intended?

PR #978, opened 2024-01-17
add receipt_authorization_code and receipt_transaction_id to the marts__combined__orders table and intermediate sourcing it

KatelynGit:
https://github.com/mitodl/hq/issues/634

This ticket adds Cybersource Auth codes and Transaction IDs to the Ecommerce order mart.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
+++ b/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
@@ -130,8 +130,8 @@ with b2becommerce_b2border as (
         , null as b2bcoupon_id
         , null as b2border_contract_number
         , null as req_reference_number
-        , ecommerce_order.receipt_authorization_code 
-        , ecommerce_order.receipt_transaction_id 
+        , ecommerce_order.receipt_authorization_code
+        , ecommerce_order.receipt_transaction_id
         , case when ecommerce_couponredemption.couponredemption_id is not null then true end as redeemed
     from ecommerce_order
     inner join ecommerce_line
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
```diff
@@ -78,6 +78,8 @@ with b2becommerce_b2border as (
         , b2becommerce_b2border.b2border_contract_number
         , productversion.productversion_readable_id
         , json_extract_scalar(b2becommerce_b2breceipt.b2breceipt_data, '$.req_reference_number') as req_reference_number
+        , json_extract_scalar(b2becommerce_b2breceipt.b2breceipt_data, '$.auth_code') as receipt_authorization_code
+        , json_extract_scalar(b2becommerce_b2breceipt.b2breceipt_data, '$.transaction_id') as receipt_transaction_id
```
rachellougee, line 82: Not super important in this PR, but I think we should move these extraction code to `stg__mitxpro__app__postgres__b2becommerce_b2breceipt` like what we did for `stg__mitxpro__app__postgres__ecommerce_receipt` to be consistent with other code. I can do that later if I have time.

rachellougee (approved):
The new code looks good, It fixed the issue for b2b orders I mentioned in https://github.com/mitodl/ol-data-platform/pull/977.

src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
KatelynGit: I was going to open up another PR for another change and I'll add it to that

PR #979, opened 2024-01-19
add ref field to orders int and mart tables

KatelynGit:
https://github.com/mitodl/hq/issues/657

add the req_reference_number column to the combined orders mart table(marts__combined__orders). And add it to the intermediate mixonline orders table as a prerequisite. 

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
+++ b/src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
@@ -81,6 +81,7 @@ select
     , payments.transaction_readable_identifier as payment_transaction_id
     , payments.transaction_bill_to_address_state as payment_bill_to_address_state
     , payments.transaction_bill_to_address_country as payment_bill_to_address_country
+    , json_extract_scalar(payments.transaction_data, '$.req_reference_number') as req_reference_number
 from lines
 inner join orders on lines.order_id = orders.order_id
 inner join users on orders.order_purchaser_user_id = users.user_id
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/intermediate/mitxonline/int__mitxonline__ecommerce_order.sql
```diff
@@ -81,6 +81,7 @@ select
     , payments.transaction_readable_identifier as payment_transaction_id
     , payments.transaction_bill_to_address_state as payment_bill_to_address_state
     , payments.transaction_bill_to_address_country as payment_bill_to_address_country
+    , json_extract_scalar(payments.transaction_data, '$.req_reference_number') as req_reference_number
```
rachellougee, line 84: Can we move this to https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__ecommerce_transaction.sql#L20?  

All these fields extracted from transaction should go to staging. I don't remember why I skipped req_reference_number for mitxonline, but it should be done the same way as xPro https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxpro/stg__mitxpro__app__postgres__ecommerce_receipt.sql#L17

rachellougee: Ok, I remember why I didn't extract `req_reference_number` for mitxonline now. So for MITx Online, we record `order_reference_number` in application database, it is the same as `req_reference_number` so that should already exist in `int_mitxonline_ecommerce_order. For xPro and Bootcamps, we don't record `order_reference_number` which is why I had to extract from transaction

KatelynGit: Hi Rachel, I made the update to use order_reference_number from int_mitxonline_ecommerce_order and i removed the changes to the intermediate layer. Let me know if it looks good. Thanks!

rachellougee (approved):
LGTM

PR #984, opened 2024-01-23
Process course xml to extract video metadata 

rachellougee:

NA
Part of https://github.com/mitodl/hq/issues/3227

Adding lib functions to process course XML and extract video metadata to json file

### How can this be tested?

Manually test it by passing `archive_path` and `json_file_path`(where the json file is written to)
```
video_data=process_video_xml('archive_path/course.6khpa7z8.tar.gz')
write_video_json_file(video_data, Path('archive_path/video_metadata.json'))
```

The output `video_metadata.json` looks like
```
{
"video_Video_021422093b2a": {"video_block_id": "video_Video_021422093b2a", "edx_video_id": "9c4be346-66f9-479c-b963-a53de9da0635", "duration": "0.0", "course_id": "course-v1:MITxT+10.MBCx+2T2023"}, 
"video_Video_1379713cc292": {"video_block_id": "video_Video_1379713cc292", "edx_video_id": "15503316-0007-4e4a-baef-d4d3a15ca257", "duration": "0.0", "course_id": "course-v1:MITxT+10.MBCx+2T2023"},
}
```

### DIFF
--- a/src/ol_orchestrate/lib/openedx.py
+++ b/src/ol_orchestrate/lib/openedx.py
@@ -1,7 +1,10 @@
 import hashlib
 import json
+import tarfile
 from datetime import UTC, datetime
+from pathlib import Path
 from typing import Any, Optional
+from xml.etree.ElementTree import ElementTree
 
 
 def generate_block_indexes(
@@ -83,3 +86,51 @@ def un_nest_course_structure(
             }
         )
     return course_blocks
+
+
+def process_video_xml(archive_path: Path) -> dict[str, Any]:
+    json_data = {}
+    with tarfile.open(archive_path, "r") as tf:
+        tf.extractall()
+        for member in tf.getmembers():
+            course_id = parse_course_id("course/course.xml")
+            if not member.isdir() and member.path.startswith("course/video/"):
+                video_data = parse_video_xml(member.path)
+                video_data["course_id"] = course_id
+                json_data[video_data["video_block_id"]] = video_data
+    return json_data
+
+
+def parse_course_id(course_xml: str) -> str:
+    with Path(course_xml).open("r") as course:
+        tree = ElementTree()
+        tree.parse(course)
+        course_root = tree.getroot()
+        run_tag = course_root.attrib.get("url_name", None)
+        org = course_root.attrib.get("org", None)
+        course_number = course_root.attrib.get("course", None)
+    return f"course-v1:{org}+{course_number}+{run_tag}"
+
+
+def parse_video_xml(video_file: str) -> dict[str, Any]:
+    with Path(video_file).open("r") as video:
+        tree = ElementTree()
+        tree.parse(video)
+        video_root = tree.getroot()
+        video_block_id = video_root.attrib.get("url_name", None)
+        edx_video_id = video_root.attrib.get("edx_video_id", None)
+        video_asset = video_root.find("video_asset", None)
+        if video_asset:
+            duration = video_asset.attrib.get("duration", None)
+
+    return {
+        "video_block_id": video_block_id,
+        "edx_video_id": edx_video_id,
+        "duration": duration,
+    }
+
+
+def write_video_json_file(video_data: dict[str, Any], video_data_file: Path):
+    with Path(video_data_file).open("w") as video_file:
+        video_file.write(json.dumps(video_data))
+        video_file.write("\n")


src/ol_orchestrate/lib/openedx.py
```diff
@@ -83,3 +86,52 @@ def un_nest_course_structure(
...
         )
     return course_blocks
+
+
+def process_video_xml(archive_path: Path) -> dict[str, Any]:
+    json_data = {}
+    with tarfile.open(archive_path, "r") as tf:
+        tf.extractall()
+        for member in tf.getmembers():
+            course_id = parse_course_id("course/course.xml")
```
rachellougee, line 96: I am not sure if we can get course_id as part of pipeline code, so I extract it from course xml

rachellougee:
@blarghmatey Can you take a look? This is based on what we discussed, but let me know if any feedback

src/ol_orchestrate/lib/openedx.py
blarghmatey: I think getting it from the course XML is a good approach as it ensures data locality and a consistent source of truth.

blarghmatey, line 96: It seems like this should go outside of the `for` loop, otherwise it will run each time through.

src/ol_orchestrate/lib/openedx.py
```diff
@@ -83,3 +86,52 @@ def un_nest_course_structure(
...
+
+    return {
+        "video_block_id": video_block_id,
+        "edx_video_id": edx_video_id,
+        "duration": duration,
+    }
+
+
+def write_video_json_file(video_data: dict[str, Any], video_data_file: Path):
+    with Path(video_data_file).open("w") as video_file:
```
blarghmatey, line 135: The video_data_file should already be a path, so this is a bit redundant. If the expectation is to get a `Path` object then you can just do `video_data_file.write_text(json.dumps(video_data))`

src/ol_orchestrate/lib/openedx.py
```diff
@@ -83,3 +86,52 @@ def un_nest_course_structure(
...
+
+def parse_course_id(course_xml: str) -> str:
+    with Path(course_xml).open("r") as course:
+        tree = ElementTree()
+        tree.parse(course)
+        course_root = tree.getroot()
+        run_tag = course_root.attrib.get("url_name", None)
+        org = course_root.attrib.get("org", None)
+        course_number = course_root.attrib.get("course", None)
+    return f"course-v1:{org}+{course_number}+{run_tag}"
```
blarghmatey, line 112: Should this raise an exception if it's unable to fetch all of the course ID components?

src/ol_orchestrate/lib/openedx.py
rachellougee: Good catch. It was a mistake. It should just run once per `process_video_xml` call

src/ol_orchestrate/lib/openedx.py
rachellougee: Thanks. It reduces to just one line and it worked fine.

src/ol_orchestrate/lib/openedx.py
rachellougee: Per slack conversation, I updated to return None in case we can't fetch course ID here. We should be able to get course ID with video ID, but we can change it to raise exception if we don't want to process video xml from here.

src/ol_orchestrate/lib/openedx.py
```diff
@@ -83,3 +86,56 @@ def un_nest_course_structure(
...
+        tree.parse(video)
+        video_root = tree.getroot()
+        video_block_id = video_root.attrib.get("url_name", None)
+        edx_video_id = video_root.attrib.get("edx_video_id", None)
+        video_asset = video_root.find("video_asset", None)
+        duration = None
+        if video_asset:
+            duration = video_asset.attrib.get("duration")
+
+        retrieved_at = datetime.datetime.now(tz=datetime.UTC).isoformat()
```
rachellougee, line 130: I get `AttributeError: module 'datetime' has no attribute 'UTC'` when running locally with `tz=datetime.UTC`, but I copy it from line49 in the same file. I suppose it should be fine?

rachellougee: I tested it with just `retrieved_at = datetime.datetime.now(tz=datetime.UTC).isoformat()` locally, but pre-commit won't allow me to push code without specifying tz

rachellougee: Just realized that datetime.UTC is new in 3.11. I am running 3.10. I can change it to datetime.timezone.utc if needed

blarghmatey (approved):

PR #986, opened 2024-01-24
Create a micromasters course certificates mart table

KatelynGit:
https://github.com/mitodl/hq/issues/3431

Creates the marts__micromasters_course_certificates table

### How can this be tested?
dbt build --select +marts__micromasters_course_certificates


src/ol_dbt/models/marts/micromasters/_marts_micromasters__models.yml
```diff
@@ -2,6 +2,62 @@
...
+  - name: course_number
+    description: str, unique string for the course e.g. 14.009x
+    tests:
+    - not_null
+  - name: user_edxorg_username
+    description: str, username on edX.org
+  - name: user_mitxonline_username
+    description: str, username on MITxOnline
+  - name: user_email
+    description: str, user email on edX.org or MITxOnline
```
rachellougee, line 42: It may be worth verifying these descriptions are accurate or not. By looking at those old DEDP course certificates, their email , full_name and country are pulling from MicroMasters.

rachellougee (approved):

src/ol_dbt/models/marts/micromasters/_marts_micromasters__models.yml
KatelynGit: I checked and they look good

KatelynGit: I also made a change to add 2 additional columns

rachellougee: FYI, these descriptions don't seem right. See https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_certificates_dedp_from_micromasters.sql#L67-L71. For old DEDP certificate holder, we are getting their profile from MicroMasters, not edX.org or MITxOnline.

PR #988, opened 2024-01-26
fix user profiles in MicroMaster program certificates models

rachellougee:

https://github.com/mitodl/hq/issues/3435

Fixing a few warnings generated https://pipelines.odl.mit.edu/runs/6c4e62bf-1d18-472d-9f20-48d7c5fd48ba
from `int__micromasters__program_certificates`. 
The warns are due to some of new DEDP certificate holders' social auth mitxonline_username on MicroMasters don't match with username on MITx Online, so we couldn't match their email and profile on MicroMasters.
For DEDP certificates on MITx Online, we are now pulling email and profile data from MITx Online directly. If profile data is blank, then fill in with MicroMasters'

### How can this be tested?

Run the following command, data should be fixed now with no warnings

```
dbt run +int__micromasters__program_certificates
dbt test int__micromasters__program_certificates
dbt build marts__micromasters_course_certificates
```

### DIFF
--- a/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
@@ -147,36 +147,37 @@ models:
     description: timestamp, timestamp of the course certificate that completed the
       program
   - name: user_gender
-    description: str, user's gender from edxorg - blank means user did not specify
-      a gender. Null means this student signed up before this information was collected
+    description: str, user's gender from user's profile on MITx Online or edX.org.
+      blank means user did not specify a gender. Null means this student signed up
+      before this information was collected
     tests:
     - accepted_values:
-        values: ['Male', 'Female', 'Other/Prefer Not to Say', '', null]
+        values: '{{ var("gender_values") }}'
   - name: user_address_city
-    description: str, city where user lives in, collected from the micromasters database
+    description: str, city where user lives in from user's profile on MicroMasters.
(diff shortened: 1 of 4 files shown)


KatelynGit (approved):
The builds/dbt tests ran clean.  One thing I noticed while testing is that user gender is populated less often with this change (null on about 106 rows vs 70 before the change).  Is that ok?

rachellougee:
> The builds/dbt tests ran clean. One thing I noticed while testing is that user gender is populated less often with this change (null on about 106 rows vs 70 before the change). Is that ok?

@KatelynGit Good call. Looks like these learners didn't provide profile data on MITx Online yet, so I fill in the blank with their MicroMasters profile data. `user_gender` , `user_address_state_or_territory` and `user_year_of_birth` are populated more than production. 
I believe  `user_gender` has over 250 null or empty in `__micromasters_program_certificates_dedp_from_mitxonline` on production. With this change, there are only 2.

KatelynGit:
> > The builds/dbt tests ran clean. One thing I noticed while testing is that user gender is populated less often with this change (null on about 106 rows vs 70 before the change). Is that ok?
> 
> @KatelynGit Good call. Looks like these learners didn't provide profile data on MITx Online yet, so I fill in the blank with their MicroMasters profile data. `user_gender` , `user_address_state_or_territory` and `user_year_of_birth` are populated more than production. I believe `user_gender` has over 250 null or empty in `__micromasters_program_certificates_dedp_from_mitxonline` on production. With this change, there are only 2.

It looks good now thanks!

PR #989, opened 2024-01-26
fix MicroMasters DEDP grades and certificates

rachellougee:

NA
I noted this issue after reviewing https://github.com/mitodl/ol-data-platform/pull/986. There are null grades for DEDP certificates in the new mart, which doesn't seem right. Despite a few certificates for `MITx/6.041x_3/2T2016` (for SDS program) are missing grades (source data from BigQuery), DEDP certificates should have grades.

```
select * from marts__micromasters_course_certificates where grade is null
```

This fixes DEDP data in `marts__micromasters_course_certificates` and its upstream models `int__micromasters__course_grades` , `int__micromasters__course_certificates`
- For old DEDP courses on edX.org, use MicroMasters data
- For new DEDP courses on MITx Online, use MITx Online data (deduped the data migrated from MicroMasters)

### How can this be tested?

```
dbt build --select +marts__micromasters_course_certificates

```
then 
```
select * from marts__micromasters_course_certificates where grade is null
```
Only a small amount of certificates with no grades from course `MITx/6.041x_3/2T2016`
DEDP certificates should have grade now

### DIFF
--- a/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
+++ b/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
@@ -15,7 +15,7 @@ with course_certificates_dedp_from_micromasters as (
 
 -- DEDP course certificates come from MicroMasters and MITxOnline. We've migrated some learners data from
 -- MicroMasters to MITxOnline around Oct 2022, but only for those users who have MITxOnline account.
--- To avoid data overlapping, we dedup based on their social auth account linked on MicroMasters.
+-- To avoid data overlapping, we deduplicate based on their social auth account linked on MicroMasters.
 -- for old DEDP courses on edx.org, then we use certificates from MicroMasters
 -- for new DEDP course on MITx Online, then we use certificates from MITx Online
 
--- a/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
+++ b/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
@@ -15,8 +15,9 @@ with course_certificates_dedp_from_micromasters as (
 
 -- DEDP course certificates come from MicroMasters and MITxOnline. We've migrated some learners data from
 -- MicroMasters to MITxOnline around Oct 2022, but only for those users who have MITxOnline account.
--- To avoid data overlapping, we use the cut-off date 2022-10-01 to reduce duplication, but there are still
--- a small number dups so need to apply additional logic to dedup based on their linked MITxOnline account
+-- To avoid data overlapping, we dedup based on their social auth account linked on MicroMasters.
+-- for old DEDP courses on edx.org, then we use certificates from MicroMasters
+-- for new DEDP course on MITx Online, then we use certificates from MITx Online
 
 
 , dedp_course_certificates_combined as (
@@ -36,7 +37,7 @@ with course_certificates_dedp_from_micromasters as (
         , coursecertificate_url as courseruncertificate_url
         , coursecertificate_created_on as courseruncertificate_created_on
     from course_certificates_dedp_from_micromasters
-    where coursecertificate_created_on < '2022-10-01'
+    where courserun_platform = '{{ var("edxorg") }}'
 
     union all
 
@@ -56,7 +57,7 @@ with course_certificates_dedp_from_micromasters as (
         , courseruncertificate_url
         , courseruncertificate_created_on
     from course_certificates_dedp_from_mitxonline
-    where courseruncertificate_created_on >= '2022-10-01'
+    where courserun_platform = '{{ var("mitxonline") }}'
 
 )
 
--- a/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_grades.sql
+++ b/src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_grades.sql
@@ -13,9 +13,11 @@ with course_grades_dedp_from_micromasters as (
     from {{ ref('__micromasters_course_grades_non_dedp_from_edxorg') }}
 )
 
--- DEDP course certificates come from MicroMasters and MITxOnline. We've migrated some learners data from
+-- DEDP course grades come from MicroMasters and MITxOnline. We've migrated some learners data from
 -- MicroMasters to MITxOnline around Oct 2022, but only for those users who have MITxOnline account.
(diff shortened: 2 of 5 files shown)


rachellougee:
@pdpinch @KatelynGit This is the PR to fix grade in `marts__micromasters_course_certificates` for DEDP.

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
```diff
@@ -15,8 +15,9 @@ with course_certificates_dedp_from_micromasters as (
 
 -- DEDP course certificates come from MicroMasters and MITxOnline. We've migrated some learners data from
 -- MicroMasters to MITxOnline around Oct 2022, but only for those users who have MITxOnline account.
--- To avoid data overlapping, we use the cut-off date 2022-10-01 to reduce duplication, but there are still
--- a small number dups so need to apply additional logic to dedup based on their linked MITxOnline account
+-- To avoid data overlapping, we dedup based on their social auth account linked on MicroMasters.
```
pdpinch, line 18:
```suggestion
-- To avoid data overlapping, we deduplicate based on their social auth account linked on MicroMasters.
```

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_grades.sql
```diff
@@ -13,9 +13,11 @@ with course_grades_dedp_from_micromasters as (
     from {{ ref('__micromasters_course_grades_non_dedp_from_edxorg') }}
 )
 
--- DEDP course certificates come from MicroMasters and MITxOnline. We've migrated some learners data from
+-- DEDP course grades come from MicroMasters and MITxOnline. We've migrated some learners data from
 -- MicroMasters to MITxOnline around Oct 2022, but only for those users who have MITxOnline account.
--- To avoid data overlapping, we use the cut-off date 2022-10-01 and their linked MITxOnline account to dedup
+-- To avoid data overlapping, we dedup based on their social auth account linked on MicroMasters.
```
pdpinch, line 18:
```suggestion
-- To avoid data overlapping, we deduplicate based on their social auth account linked on MicroMasters.
```

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_grades.sql
```diff
@@ -13,9 +13,11 @@ with course_grades_dedp_from_micromasters as (
...
+-- for old DEDP courses on edx.org, then we use the grade from MicroMasters
```
pdpinch, line 19: For these old DEDP courses, learners need to pass the course on edx.org and then also pass a proctored exam (which was offered by Pearson, and later, by MITx Online). Does this table contain the "combined" grade, which is the weighted average of these two grades?

src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_grades_dedp_from_micromasters.sql
```diff
@@ -27,11 +31,15 @@ with dedp_course_grades as (
         ) as row_num
     from courserun_grades
     inner join courseruns on courserun_grades.courserun_id = courseruns.courserun_id
+    inner join dedp_course_certificates
+        on
+            courserun_grades.user_id = dedp_course_certificates.user_id
+            and courseruns.course_id = dedp_course_certificates.course_id
+            and courseruns.courserun_start_on < dedp_course_certificates.coursecertificate_created_on
```
pdpinch, line 38: I don't understand what this last clause is doing. What does the courserun start date and the certificate created date have to do with it?

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_grades.sql
rachellougee: yes, so for old DEDP courses, this is "combined" grade from MicroMasters, which is weighted average grade of proctored exam and course grade

src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_grades_dedp_from_micromasters.sql
rachellougee: So for old DEDP on MicroMasters, we needed to find a specific run for the course in certificate and grade models. There are cases where learners have grades for multiple runs in the same course, so the logic to determine which run: 
   - Pick the highest grade for the course
   - The highest grade could be the later run than the certificate was generated, so we needed to ensure this highest grade we pick matched the timeline of certificates

For example,
This learner has certificate for course `14.740x` https://micromasters.mit.edu/admin/grades/micromasterscoursecertificate/440/, which generated on Sept 19, 2017
The learner's highest grade for this course is [course-v1:MITx+14.740x+1T2018](https://micromasters.mit.edu/admin/grades/finalgrade/66002/change/), so I assume it wouldn't make sense to match this certificate for course-v1:MITx+14.740x+1T2018 since this run hasn't started when certificate was generated.  So this where clause is to make sure we pick `course-v1:MITx+14.740x+2T2017` instead of `course-v1:MITx+14.740x+1T2018`

Does it make sense? or that's not how it supposed to work?

rachellougee:
Thank you for the review. @pdpinch I updated the description per your comment. I also tried to answer your two questions, but let me know. As for the data, I can share it with you in spreadsheet if you prefer since the output of this table is under my schema on Starburst.

KatelynGit:
> @pdpinch @KatelynGit This is the PR to fix grade in `marts__micromasters_course_certificates` and its upstream intermediate models for DEDP.

The build ran clean and it looks like the field is populated more often now. No records were dropped and no duplicates. code LGTM.

KatelynGit (approved):
Approval

src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_grades_dedp_from_micromasters.sql
pdpinch: I thought we had materialized the combined grades in a table in MicroMasters. Does your algorithm match the materialized grades? 

2. TBH, I think we should be combining the best courseware grade with the best proctored exam grade, regardless of timing. However, in practice, I'm not sure if that's what we did, and I don't want to consider changing the behavior of legacy micromasters code.

pdpinch (approved):

PR #991, opened 2024-02-01
Edxorg archive processing

blarghmatey:
Addresses https://github.com/mitodl/hq/issues/1309

This builds an asset graph for processing and decomposing the contents of edx.org data archives into data assets that are consumed by our data platform.

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### How can this be tested?

TBA


src/ol_orchestrate/definitions/edx/retrieve_irx_edx_exports.py
```diff
@@ -38,29 +125,22 @@ def s3_uploads_bucket(
     return bucket_map[dagster_env]
 
 
-def weekly_edx_exports_config(
+def edxorg_data_archive_config(
     irx_edxorg_gcs_bucket, ol_edxorg_raw_data_bucket, new_files_to_sync
 ):
     return {
         "ops": {
-            "download_edx_data_exports": {
```
rachellougee, line 46: Is this removal intended? How does courses archive get downloaded?

blarghmatey: Yes, that's intentional. The archive download happens via the IO manager initialized as the `gcs_input` key in the resources definition further down. There's no longer an explicit op to manage the file downloading, instead allowing that to be part of the implicit Dagster machinery.

Issue #939, opened 2023-12-04
Add fields to the enrollment mart for finance


KatelynGit:
# User Story
As a member of the finance team, I would like to have a single source of all the Engineering data I need for enrollment reporting so that I can regularly pull the data into Tableau.

# Description/Context
Some enrollment related fields are available in BI(https://bi.odl.mit.edu/queries/1142) but not in the marts(surfaced via Tableau) we need to fill in the gaps in the enrollment mart

# Acceptance Criteria
All info in the BI report  1142 should be accessible via the mart/Tableau

# Plan/Design
[Finance_reports_consolidation.xlsx](https://github.com/mitodl/ol-data-platform/files/13545819/Finance_reports_consolidation.xlsx)

PR #992, opened 2024-02-02
add fields to marts__combined_course_enrollment_detail and upstream changes

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/939

- Adding is_edx_enrolled, courserun_start_on, user_country_code, courseruncertificate_uuid, payment_req_reference_number, and tax related fields to `marts__combined_course_enrollment_detail`
- fixing a couple things in `int__mitxpro__ecommerce_allorders`

### How can this be tested?

```
dbt build +marts__combined_course_enrollment_detail
dbt build +int__mitxpro__ecommerce_allorders
```

### DIFF
--- a/src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
+++ b/src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
@@ -51,7 +51,7 @@ models:
     tests:
     - not_null
   - name: receipt_reference_number
-    description: str, transaction reference number from user's cybersource payment
+    description: str, req_reference_number from user's cybersource payment
     tests:
     - not_null
   - name: receipt_transaction_status
@@ -194,6 +194,8 @@ models:
   - name: courserun_readable_id
     description: str, unique string to identify a bootcamp course run (could be blank
       for older runs)
+  - name: courserun_start_on
+    description: timestamp, date and time when the course starts
   - name: user_username
     description: str, name chosen by user
     tests:
@@ -204,6 +206,8 @@ models:
     - not_null
   - name: user_full_name
     description: string, full name on user's profile
+  - name: user_address_country
+    description: str, country code for the user's address
   tests:
   - dbt_expectations.expect_table_row_count_to_equal_other_table:
       compare_model: ref('stg__bootcamps__app__postgres__courserunenrollment')
--- a/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
@@ -223,8 +223,12 @@ models:
     description: str, user full name on edX.org or MITxOnline
   - name: user_country
     description: str, user country on edX.org or MITxOnline
+  - name: courseruncertificate_uuid
+    description: str, unique identifier for the certificate on MITx Online, MicroMasters
+      or edX.org
   - name: courseruncertificate_url
-    description: str, the full URL to this DEDP course certificate on MITx Online
+    description: str, the full URL to this DEDP course certificate on MITx Online,
+      MicroMasters or edX.org
   - name: courseruncertificate_created_on
     description: timestamp, date and time when this course certificate was initially
       created
@@ -436,6 +440,8 @@ models:
   - name: receipt_authorization_code
     description: str, authorization code from most recent cybersource payment for
       the order
+  - name: receipt_reference_number
+    description: str, req_reference_number from cybersource payment
   - name: receipt_bill_to_address_state
     description: str, address state from most recent cybersource payment for the order
   - name: receipt_bill_to_address_country
--- a/src/ol_dbt/models/intermediate/bootcamps/int__bootcamps__courserunenrollments.sql
+++ b/src/ol_dbt/models/intermediate/bootcamps/int__bootcamps__courserunenrollments.sql
@@ -22,9 +22,11 @@ with enrollments as (
         , enrollments.courserunenrollment_enrollment_status
         , runs.courserun_readable_id
         , runs.courserun_title
+        , runs.courserun_start_on
         , users.user_username
         , users.user_email
         , users.user_full_name
+        , users.user_address_country
     from enrollments
     inner join runs on enrollments.courserun_id = runs.courserun_id
     inner join users on enrollments.user_id = users.user_id
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -129,6 +129,8 @@ models:
       that use new format
     tests:
     - not_null
+  - name: courserun_start_on
+    description: timestamp, date and time when the course starts
   - name: user_username
     description: str, username of a learner on the edX platform
     tests:
--- a/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courserun_enrollments.sql
+++ b/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courserun_enrollments.sql
@@ -40,10 +40,7 @@ with person_courses as (
 )
 
 , edxorg_runs as (
-    select
-        courserun_readable_id
-        , course_number
-        , courserun_title
+    select *
     from {{ ref('stg__edxorg__bigquery__mitx_courserun') }}
 )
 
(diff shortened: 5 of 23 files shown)


src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
```diff
@@ -223,8 +223,12 @@ models:
     description: str, user full name on edX.org or MITxOnline
   - name: user_country
     description: str, user country on edX.org or MITxOnline
+  - name: courseruncertificate_uuid
+    description: str, unique identifier for the certificate on MITx Online, MicroMasters
+      or edX.org
   - name: courseruncertificate_url
-    description: str, the full URL to this DEDP course certificate on MITx Online
+    description: str, the full URL to this DEDP course certificate on MITx Online,
+      MicroMasters or edX.org
```
pdpinch, line 231: Why does this say DEDP course certificate? 

If this is a combined enrollment mart, I would expect the certificate URL to be for any of the possible sites, i.e. edx.org, mitxonline.mit.edu, (legacy) micromasters.mit.edu or xpro.mit.edu

src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
```diff
@@ -51,7 +51,7 @@ models:
     tests:
     - not_null
   - name: receipt_reference_number
-    description: str, transaction reference number from user's cybersource payment
+    description: str, req_reference_number from user's cybersource payment
```
pdpinch, line 54: I'm a little confused by "cybersource payment" -- would it be accurate to say "cybersource receipt" ? Isn't that the table where the data is captured?

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
```diff
@@ -445,6 +451,12 @@ models:
     description: str, discount amount in readable format. It can be fixed discount
       which is $<dollar amount> off, fixed-price which is $<dollar amount> or percent-discount
       which is <percentage> % off
+  - name: discount_amount
+    description: numeric, actual discount dollar amount. For percent-discount coupon,
+      this is calculated as line_price * percentage off
+  - name: coupon_type
+    description: str, type of the coupon which describes what circumstances the coupon
+      can be redeemed. Possible values are "standard" or "discounted-previous-course".
```
pdpinch, line 459: I think we have multiple different "types" of coupons depending on the application source. I think it would be good to try to simplify and better describe these. 

for example, xPRO has `coupon_type` and `payment_type`. The `payment_type` is most relevant to the finance team.

pdpinch (commented):
I have a couple of questions. Sorry if I'm opening a can of worms.

src/ol_dbt/models/intermediate/bootcamps/_int_bootcamps__models.yml
rachellougee: yes, I think cybersource receipt is more accurate, I will update.

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
rachellougee: This is MicroMasters intermediate model. That DEDP was just a mistake from last time, I will update that.

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
rachellougee: This is specific for `int__micromasters__orders` for MicroMasters legacy orders. The reason I added `coupon_type` is this report from xPro https://bi.odl.mit.edu/queries/703/source, which has coupon_type.

For payment_type in xPro, if I understand correctly, it's order_type in this report https://bi.odl.mit.edu/queries/703/source? It has some logic to convert it `staff`, `Boeing`, `B2B` and B2C. I didn't add this field to combined mart since I am not sure how we should convert it in MITx Online or Bootcamps

rachellougee: I found payment_type in discount model from MITx Online, but it's renamed to discount_source for some reason in staging. I can add payment_type to the combined mart even its not used in 703 report

KatelynGit:
The builds ran well no errors for the altered tables.  

When looking at the tables via test queries just had a question-
- There are a lot of course runs with no orders attached to them and no related order data.  This seems odd. For example courserun_readable_id MITx/6.00.1x/2T2020 no order or related data is attached. Do we need to pull that data from elsewhere?  how are users signing up without an order? 

It also looks like discount amounts are not being calculated properly. We may need a fix there. For example:
select *
from ol_warehouse_production_rlougee_mart.marts__combined_course_enrollment_detail
where discount_amount = 5.833500
--not an actual dollar amount

rachellougee:
> There are a lot of course runs with no orders attached to them and no related order data. This seems odd. For example courserun_readable_id MITx/6.00.1x/2T2020 no order or related data is attached. Do we need to pull that data from elsewhere? how are users signing up without an order?

@KatelynGit There are two things I want to clarify
1. learner can sign up in a course without any order. It will create order when they upgrade on our application, which is verified mode `courserunenrollment_enrollment_mode ='verified'`
2. For edX.org course, which is the example you pointed out `MITx/6.00.1x/2T2020` https://www.edx.org/learn/computer-science/massachusetts-institute-of-technology-introduction-to-computer-science-and-programming-using-python?utm_source=mitopenlearning-mit-open-learning&utm_medium=affiliate_partner. That's paid via edX.org, not in our application, so we don't have orders for these courses. I am not aware of where we can get their orders. Maybe @pdpinch know if its possible

> select *
from ol_warehouse_production_rlougee_mart.marts__combined_course_enrollment_detail
where discount_amount = 5.833500

Good catch on that. I fixed it for xPro

KatelynGit:
> > There are a lot of course runs with no orders attached to them and no related order data. This seems odd. For example courserun_readable_id MITx/6.00.1x/2T2020 no order or related data is attached. Do we need to pull that data from elsewhere? how are users signing up without an order?
> 
> @KatelynGit There are two things I want to clarify
> 
> 1. learner can sign up in a course without any order. It will create order when they upgrade on our application, which is verified mode `courserunenrollment_enrollment_mode ='verified'`
> 2. For edX.org course, which is the example you pointed out `MITx/6.00.1x/2T2020` https://www.edx.org/learn/computer-science/massachusetts-institute-of-technology-introduction-to-computer-science-and-programming-using-python. That's paid via edX.org if learner chose to upgrade, not in our application, so we don't have orders for these courses. I am not aware of where we can get their orders. Maybe @pdpinch know if its possible
> 
> > select *
> > from ol_warehouse_production_rlougee_mart.marts__combined_course_enrollment_detail
> > where discount_amount = 5.833500
> 
> Good catch on that. I fixed it for xPro

Look great now! also thanks for the information on signing up for a course without an order

KatelynGit (approved):

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
pdpinch: I see. When this is ready, I'd like to inspect the data just to understand what we have in micromasters and what we might need to migrate going forward.

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
pdpinch: I'm not sure what the right solution is here, but we should keep an eye on these fields and try to move towards a common set of names, so people know what to expect in them. I'm not sure if that normalization should happen in stage or intermediate. It will have to happen in the combined mart.

pdpinch:
> For edX.org course, which is the example you pointed out MITx/6.00.1x/2T2020 https://www.edx.org/learn/computer-science/massachusetts-institute-of-technology-introduction-to-computer-science-and-programming-using-python. That's paid via edX.org if learner chose to upgrade, not in our application, so we don't have orders for these courses. I am not aware of where we can get their orders. Maybe @pdpinch know if its possible

Indeed, these are courses that were purchased on edx.org and, as far as I know, we don't have any order data for them. Perhaps Jon Daries in IRx would have a data source, but I doubt it. 

Should we mention this deficiency somewhere in the documentation?

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
rachellougee: Sure. The `payment_type` are called `discount_source` in xPro and MITx Online staging discount models, so that's consistent there. Ideally normalization should be done in staging, but in this case I rename it back to `payment_type` in this mart as that's what is called in the source table.  I will make separate PR to normalize this name in staging.

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
rachellougee: I will send you a link once its on production.

PR #997, opened 2024-02-05
add new combined user mart

KatelynGit:
https://github.com/mitodl/hq/issues/1644

creates the marts__combined__users by sourcing de-duplicated mitx data as well as bootcamp and xpro platform data from the intermediate layer

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -6,11 +6,17 @@ models:
   description: Mart model for users from different platforms
   columns:
   - name: user_email
-    description: string, user email on the corresponding platform
+    description: string, user email on the corresponding platform including xpro, 
+      bootcamps, edX.org, or MITx Online. If the user is on edX.org and MITx Online
+      then the email will be chosen from the last platform they joined.
   - name: user_joined_on
-    description: timestamp, user join timestamp
+    description: timestamp, user join timestamp on the corresponding platform including xpro, 
+      bootcamps, edX.org, or MITx Online. If the user is on edX.org and MITx Online
+      then the timsestamp will be chosen from the latest platform they joined.
   - name: user_last_login
-    description: timestamp, user last login
+    description: timestamp, user last login on the corresponding platform including xpro, 
+      bootcamps, edX.org, or MITx Online. If the user is on edX.org and MITx Online
+      then the timsestamp will be chosen from the latest login. 
   - name: platforms
     description: string, application where the data is from
   - name: user_full_name
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -14,9 +14,9 @@ models:
   - name: platforms
     description: string, application where the data is from
   - name: user_full_name
-    description: str, user full name. Very small number of
-      edX.org users have blank full name, their name couldn't be populated from other
-      sources if they don't have their accounts linked on MicroMasters.
+    description: str, user full name. Very small number of edX.org users have blank
+      full name, their name couldn't be populated from other sources if they don't
+      have their accounts linked on MicroMasters.
   - name: user_address_country
     description: str, country code provided by the user on the corresponding platform
   - name: user_highest_education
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -2,6 +2,57 @@
 version: 2
 
 models:
+- name: marts__combined__users
+  description: Mart model for users from different platforms
+  columns:
+  - name: user_email
+    description: string, user email on the corresponding platform
```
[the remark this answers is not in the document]
rachellougee: Same for other fields like user_joined_on, user_last_login

src/ol_dbt/models/marts/combined/marts__combined__users.sql
```diff
@@ -0,0 +1,134 @@
...
+        , null as user_bootcamps_username
+        , case
+            when is_mitxonline_user = false
+                then user_edxorg_email
+            when is_edxorg_user = false
+                then user_mitxonline_email
+            when user_joined_on_mitxonline > user_joined_on_edxorg
+                then user_mitxonline_email
+            else coalesce(user_edxorg_email, user_mitxonline_email)
+        end as user_email
```
rachellougee, line 33: If I read it correctly, this can be written to:
```suggestion
        , case
            when user_joined_on_mitxonline > user_joined_on_edxorg
                then user_mitxonline_email
            else coalesce(user_edxorg_email, user_mitxonline_email)
        end as user_email
```

src/ol_dbt/models/marts/combined/marts__combined__users.sql
```diff
@@ -0,0 +1,134 @@
...
+        , case
+            when user_joined_on_mitxonline > user_joined_on_edxorg
+                then user_joined_on_edxorg
+            else user_joined_on_mitxonline
+        end as user_joined_on
```
rachellougee, line 38: If a user joined MITx Online recently, who also had an account on edX.org
- `user_email` is **MITx Online** email
- `user_joined_on` is the date joined on  **edX.org**

That seems confusing to me as two fields are pulling from different applications. I am not sure how this will be used.

src/ol_dbt/models/marts/combined/marts__combined__users.sql
```diff
@@ -0,0 +1,134 @@
...
+        , case
+            when user_last_login_on_mitxonline > user_last_login_on_edxorg
+                then user_last_login_on_mitxonline
+            else user_last_login_on_edxorg
+        end as user_last_login
```
rachellougee, line 43: This `user_last_login` makes sense to me

rachellougee (commented):
Left a few comments and questions. We also have `int__combined__users` which is different than this mart. We probably should either just delete that intermediate model, or move the query there. But it can be done later.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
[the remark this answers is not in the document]
KatelynGit: I've added clarity to the descriptions for these fields

src/ol_dbt/models/marts/combined/marts__combined__users.sql
KatelynGit: I changed it for simplicity thanks

src/ol_dbt/models/marts/combined/marts__combined__users.sql
KatelynGit: I changed joined to latest and email to latest so they match

src/ol_dbt/models/marts/combined/marts__combined__users.sql
KatelynGit: ok cool left that one as is

rachellougee (commented):
@KatelynGit Can you take another look? It looked like `user_joined_on` and `user_last_login` are all NULL for MITx Online users, which doesn't seem right.

```
SELECT * FROM "ol_data_lake_production"."ol_warehouse_production_rlougee_mart"."marts__combined__users"
where platforms='mitxonline'
```

You may want to change that ELSE to coalesce since they don't have data on edX.org.
That's the only thing I noticed

KatelynGit:
> @KatelynGit Can you take another look? It looked like `user_joined_on` and `user_last_login` are all NULL for MITx Online users, which doesn't seem right.
> 
> ```
> SELECT * FROM "ol_data_lake_production"."ol_warehouse_production_rlougee_mart"."marts__combined__users"
> where platforms='mitxonline'
> ```
> 
> You may want to change that ELSE to coalesce since they don't have data on edX.org. That's the only thing I noticed

@rachellougee I changed them to coalesce statements to fix it

rachellougee (approved):

PR #1001, opened 2024-02-08
Create a combined user course roles from MITx Online, xPro and Residential

rachellougee:
### What are the relevant tickets?

Part of https://github.com/mitodl/ol-data-platform/issues/561

### Description (What does it do?)

- Creating `stg__mitxonline__openedx__mysql__user_courseaccessrole`, `stg__mitxpro__openedx__mysql__user_courseaccessrole`
- Creating `int__combined__user_course_roles` from MITx Online, xPro and Residential open edx platform
  - Can be used for our Superset course access control
  - Can be used for https://github.com/mitodl/ol-data-platform/issues/561 report

### How can this be tested?

```
dbt build +int__combined__user_course_roles
```

PR #1002, opened 2024-02-08
fix courserun_platform in MITx Online courserun model

rachellougee:

NA. Notice it while working on https://github.com/mitodl/ol-data-platform/pull/1001

Fixing `courserun_platform` in `stg__mitxonline__app__postgres__courses_courserun` to count for runs that don't follow MITx Online course ID schema. e.g. `course-v1:MITxt+3.012Tx+2T2023`. So that all the MITx Online enrollments are included in int__mitx__courserun_enrollments and marts__mitxonline_course_enrollments.

Note that some edX.org courses were migrated from MicroMasters to MITx Online, so we needed `courserun_platform` as a filter for old edX.org courses.

### How can this be tested?

Run `dbt build --select stg__mitxonline__app__postgres__courses_courserun`

Then check `stg__mitxonline__app__postgres__courses_courserun`, all the runs should have the correct courserun_platform.  e.g.
    - `course-v1:MITxt+3.012Tx+2T2023`, course-v1:TestX+Test101+3T2022 should be MITx Online
    -  course-v1:MITx+14.740x+3T2017 , MITx/14.73x_1/1T2015 should be edX.org

### DIFF
--- a/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
+++ b/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
@@ -947,7 +947,8 @@ models:
     - not_null
   - name: courserun_platform
     description: str, indicating whether this course runs on MITx Online or edX.org.
-      Some of edx.org data were migrated from MicroMasters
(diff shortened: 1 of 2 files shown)


quazi-h (approved):
Logic looks good, was able to run the dbt build command and confirmed the correct courserun_platform for the examples given.

PR #1016, opened 2024-02-26
add user_is_active to intermediate and mart

rachellougee:

NA
Part of https://github.com/mitodl/hq/issues/3613

This adds user_is_active to intermediate models - int__combined__users, int__edxorg__mitx_users , int__mitx__users and marts__combined__users 
- it doesn't filter out inactive users in intermediate to make sure we don't drop order and enrollment data for any inactive users 
- for marts__combined__users, not sure if it makes sense to filter inactive users. But for now I've added user_is_active and pull the correct email if the account exist on both edx.org and MITx Online. 

### How can this be tested?

```
dbt build --select int__edxorg__mitx_users
dbt build --select int__mitx__users
dbt build --select int__combined__users
```

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -226,6 +226,9 @@ models:
   - name: user_mitxonline_username
     description: str, username on MITx Online if user's account is linked on MicroMasters
       portal
+  - name: user_is_active
+    description: boolean, indicating if user's account is active on edx.org. A user
+      is inactive when their username or email is hashed.
 
 - name: int__edxorg__mitx_courserun_grades
   description: Intermediate model for course run grades from edx.org
--- a/src/ol_dbt/models/marts/combined/marts__combined__users.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__users.sql
@@ -23,27 +23,32 @@ with mitx__users as (
         , null as user_mitxpro_username
         , null as user_bootcamps_username
         , case
-            when user_joined_on_mitxonline > user_joined_on_edxorg
+            when user_is_active_on_mitxonline and user_joined_on_mitxonline > user_joined_on_edxorg
                 then user_mitxonline_email
             else coalesce(user_edxorg_email, user_mitxonline_email)
         end as user_email
         , case
-            when user_joined_on_mitxonline > user_joined_on_edxorg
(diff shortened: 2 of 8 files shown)


src/ol_dbt/models/marts/combined/marts__combined__users.sql
```diff
@@ -23,27 +23,32 @@ with mitx__users as (
         , null as user_mitxpro_username
         , null as user_bootcamps_username
         , case
-            when user_joined_on_mitxonline > user_joined_on_edxorg
+            when user_is_active_on_mitxonline and user_joined_on_mitxonline > user_joined_on_edxorg
                 then user_mitxonline_email
```
rachellougee, line 27: there are some retired users on MITx Online, this is to make sure we don't use the hashed email for this
```
select * from ol_warehouse_production_mart.marts__combined__users
where user_mitxonline_username is not null and user_edxorg_username is not null
and user_email like 'retired%'
```

src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
KatelynGit, line 1: I'm getting an error when I try to run the build here. We may need to make a change to the code. 

Database Error in test relationships_int__edxorg__mitx_courserun_enrollments_user_id__user_id__ref_int__edxorg__mitx_users_ (models/intermediate/edxorg/_int_edxorg__models.yml)
13:55:01    TrinoQueryError(type=INTERNAL_ERROR, name=GENERIC_INTERNAL_ERROR, message="java.lang.RuntimeException: java.util.concurrent.TimeoutException: Total timeout 20000 ms elapsed", query_id=20240227_135440_00179_aszsu)

KatelynGit (changes requested):

KatelynGit:
Builds ran clean and definitions and tables looked good. Only issue I saw was the error running the int__edxorg__mitx_users build due to a relationship test.

src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
rachellougee: @KatelynGit I don't see any error when running ` dbt build --select int__edxorg__mitx_users --vars 'schema_suffix: rlougee' --target dev_production`. But based on what you posted here, its not related to this change. If you read the error, it points to this line https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml#L124

So I think you need to run `int__edxorg__mitx_courserun_enrollments`. Can you try that?

KatelynGit (approved):

PR #1017, opened 2024-03-01
Giving mit_irx select privileges on program certificates tables

quazi-h:
These changes should give Jon and the IRX team access to the micromasters program credential data so they can include it in reports they provide to OL Finance.

Closes https://github.com/mitodl/hq/issues/3021
and https://github.mit.edu/micromasters/micromasters-issues/issues/40

This is giving the mit_irx role select privileges on the following tables:
int__edxorg__mitx_program_certificates
marts__micromasters_program_certificates

### How can this be tested?
```
dbt build --select +int__edxorg__mitx_program_certificates
dbt build --select +marts__micromasters_program_certificates

```

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -377,9 +377,6 @@ models:
 - name: int__edxorg__mitx_program_certificates
   description: learners who completed MicroMasters and XSeries programs on edX.org.
     program_certificate_awarded_on might be null for some programs such as SCM
-  config:
-    grants:
-      select: ['mit_irx']
   columns:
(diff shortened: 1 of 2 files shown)


rachellougee (commented):
This PR is on hold per @pdpinch. Also it's probably not necessary to give access for both `int__edxorg__mitx_program_certificates` and `marts__micromasters_program_certificates` since these two tables overlap

quazi-h:
> This PR is on hold per @pdpinch. Also it's probably not necessary to give access for both `int__edxorg__mitx_program_certificates` and `marts__micromasters_program_certificates` since these two tables overlap

Okay, should I only give access to `marts__micromasters_program_certificates` as that contains DEDP program certificates?

rachellougee (approved):

PR #1021, opened 2024-03-07
add income fields to mitxonline users mart

KatelynGit:
https://github.com/mitodl/hq/issues/3656

adds income fields(latest_income_usd, latest_original_income, latest_original_currency) to mitxonline users mart (marts__mitxonline_user_profiles)

### How can this be tested?
dbt build --select +marts__mitxonline_user_profiles

### DIFF
--- a/src/ol_dbt/models/marts/mitxonline/marts__mitxonline_user_profiles.sql
+++ b/src/ol_dbt/models/marts/mitxonline/marts__mitxonline_user_profiles.sql
@@ -7,12 +7,12 @@ with users as (
 )
 
 , user_income as (
-    select 
+    select
         user_id
         , flexiblepriceapplication_income_usd
         , flexiblepriceapplication_original_income
         , flexiblepriceapplication_original_currency
-        , rank() over (partition by user_id order by flexiblepriceapplication_updated_on desc) as rnk  
+        , rank() over (partition by user_id order by flexiblepriceapplication_updated_on desc) as rnk
     from income
 )
 
@@ -31,6 +31,6 @@ select
     , user_income.flexiblepriceapplication_original_currency as latest_original_currency
 from users
 left join user_income
-    on 
+    on
         users.user_id = user_income.user_id
         and user_income.rnk = 1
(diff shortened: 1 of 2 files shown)


quazi-h (approved):
Changes here look good to me, I was able to run this against my own dev_production schema and the updated table contains the 3 extra columns. 13.8% of the records had values for these three new fields.

src/ol_dbt/models/marts/mitxonline/marts__mitxonline_user_profiles.sql
```diff
@@ -2,13 +2,35 @@ with users as (
...
+    select * from {{ ref('int__mitxonline__flexiblepricing_flexiblepriceapplication') }}
+)
+
+, user_income as (
+    select
+        user_id
+        , flexiblepriceapplication_income_usd
+        , flexiblepriceapplication_original_income
+        , flexiblepriceapplication_original_currency
+        , rank() over (partition by user_id order by flexiblepriceapplication_updated_on desc) as rnk
```
rachellougee, line 15: @KatelynGit Do we not consider the `flexiblepriceapplication_status` to pull their most recent application?  I thought that in the past we have use the approved flexible price application since their most recent application could be denied.

rachellougee: Looking at Bryan's comment https://github.com/mitodl/hq/issues/3656#issuecomment-1973799214 `need the most recent/final reported income.`

rachellougee: I can understand pulling from most recent application regardless of status. But I think you may want to clarify with Bryan or whoever is familiar with flexible pricing application since `most recent/final` is a bit ambiguous.

rachellougee: I just raised the question, but I am not certain either. Up to you if pulling most recent one makes sense.

KatelynGit: pulling the most recent definitely make sense and we are.  But I'll definitely clarify on status and if they want us to not show income in those cases.

PR #1027, opened 2024-03-11
updates to edxorg tracking log staging model

rachellougee:

Part of https://github.com/mitodl/hq/issues/3713
The change to `int__edxorg__mitx_user_courseactivities` will need to wait until we have `courseware_studentmodule` in place for edx.org

Updates `stg__edxorg__s3__tracking_logs__user_activity` to use our own edX.org tracking log table raw__edxorg__s3__tracking_logs

### How can this be tested?

```
dbt build --select stg__edxorg__s3__tracking_logs__user_activity --full-refresh
```

### DIFF
--- a/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__tracking_logs__user_activity.sql
+++ b/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__tracking_logs__user_activity.sql
@@ -8,7 +8,7 @@
 }}
 
 with source as (
-    select * from {{ source('ol_warehouse_raw_data','raw__irx__edxorg__s3__tracking_logs') }}
+    select * from {{ source('ol_warehouse_raw_data','raw__edxorg__s3__tracking_logs') }}
     where
         username != ''
         and json_query(context, 'lax $.user_id' omit quotes) is not null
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__tracking_logs__user_activity.sql
```diff
@@ -49,7 +49,7 @@ with source as (
         , name as useractivity_event_name
         , event_type as useractivity_event_type
         , {{ extract_course_id_from_tracking_log() }} as courserun_readable_id
-        , json_query(context, 'lax $.user_id' omit quotes) as user_id
+        , cast(json_query(context, 'lax $.user_id' omit quotes) as integer) as user_id
```
quazi-h, line 52: should this column be named "openedx_user_id" like in [stg__edxorg__s3__tracking_logs__user_activity](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxonline/stg__mitxonline__openedx__tracking_logs__user_activity.sql#L57)? Ticket doesn't mention that column names should be the same, just data types and filtering logic but I wanted to double check.

rachellougee: Good question. I think it should be named user_id for edx.org as there is only one user_id that is referenced in stg__edxorg__bigquery__mitx_person_course. For MITx Online and xPro, there are user_id and openedx_user_id from app and open edx, so we need that distinction for referencing in other models

quazi-h (approved):
Changes look good, I was able to build against my own dev_production db and see the correct table structure and data in stg__edxorg__s3__tracking_logs__user_activity.

PR #1028, opened 2024-03-14
Create OCW course mart

rachellougee:

mart for https://github.com/mitodl/hq/issues/3609

Create `marts__ocw_courses` to replace the query in https://bi.odl.mit.edu/queries/1353 with mart model. 
OCW course metadata are different from MITx Online or xPro platform, it makes sense to have it in a separate mart since most of the fields are unique to OCW.

### How can this be tested?

Run the following command, tests should all pass.
```
dbt build --select +marts__ocw_courses  --vars 'schema_suffix: <YOURNAME>' --target dev_production
```

### DIFF
--- a/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
+++ b/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
@@ -3,7 +3,7 @@ version: 2
 
 models:
 - name: marts__ocw_courses
-  description: OpenCourseWare courses metadata
+  description: OCW courses managed in OCW Studio
   columns:
   - name: course_short_id
     description: str, short unique identifier for the OCW course. e.g. 18.05-spring-2022
--- a/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
+++ b/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
@@ -5,9 +5,8 @@ models:
 - name: marts__ocw_courses
   description: OpenCourseWare courses metadata
   columns:
-  - name: course_readable_id
+  - name: course_short_id
     description: str, short unique identifier for the OCW course. e.g. 18.05-spring-2022
-      The source of this field is short ID of the website.
     tests:
     - not_null
     - unique
--- a/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
+++ b/src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
@@ -1,0 +1,73 @@
+---
+version: 2
+
+models:
+- name: marts__ocw_courses
+  description: OpenCourseWare courses metadata
+  columns:
+  - name: course_readable_id
+    description: str, short unique identifier for the OCW course. e.g. 18.05-spring-2022
+      The source of this field is short ID of the website.
+    tests:
+    - not_null
+    - unique
+  - name: course_name
+    description: str, unique name of the OCW course. It appears as part of OCW Studio
+      urls
+    tests:
+    - not_null
+    - unique
+  - name: course_title
+    description: str, title of the OCW course
+    tests:
+    - not_null
+  - name: course_live_url
+    description: str, url to the OCW course on production.
+  - name: course_is_live
+    description: boolean, indicating if the course website is currently published
+      on OCW production.
+    tests:
+    - not_null
+  - name: course_is_unpublished
+    description: boolean, indicating if the course website is currently unpublished
+      on OCW production.
+    tests:
+    - not_null
+  - name: course_has_never_published
+    description: boolean, indicating if the course website has never been published
+      on OCW production.
+    tests:
+    - not_null
+  - name: course_first_published_on
+    description: timestamp, date and time when the course website was first published
+      to OCW production
+  - name: course_publish_date_updated_on
+    description: timestamp, date and time when the publish date was most recently
+      updated on OCW production. The publish date is updated by the webhook to Concourse
+      pipeline (it rebuilds every site) whenever we do a release in ocw repos.
+  - name: course_term
+    description: str, course term. Possible values are Spring, Summer, Fall, January
+      IAP, or blank
+  - name: course_year
+    description: int, course year
+  - name: course_level
+    description: str, course level in comma-separated list. Possible values are Undergraduate
+      , Graduate, Non-Credit, High School, or blank
+  - name: course_primary_course_number
+    description: str, the primary course number e.g. 21A.850J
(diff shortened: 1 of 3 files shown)


src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
```diff
@@ -0,0 +1,74 @@
...
+models:
+- name: marts__ocw_courses
+  description: OpenCourseWare courses metadata
+  columns:
+  - name: course_short_id
+    description: str, short unique identifier for the OCW course. e.g. 18.05-spring-2022
+    tests:
+    - not_null
+    - unique
+  - name: course_name
```
KatelynGit, line 13: This doesn't seem like a very useful field.  The info is already present in the URL field on this table.  I would remove it because just looking at the title of this field I was unsure of the difference between this   and "course title". If we do keep it maybe we could re-name it something more descriptive like "course url short name" to differentiate it.

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
```diff
@@ -0,0 +1,74 @@
+---
+version: 2
+
+models:
+- name: marts__ocw_courses
+  description: OpenCourseWare courses metadata
```
KatelynGit, line 6: I'm not sure why the work "metadata" is in the table description?  Usually metadata is data about other data for example like row created and update timestamps etc.  But this table looks like it contains actual data about the course not just metadata.  I'd remove it but definitely up to you.  If we want a longer description we could add something like "OpenCourseWare courses are a series of classes/education materials offered by MIT"

src/ol_dbt/models/marts/ocw/marts__ocw_courses.sql
```diff
@@ -0,0 +1,33 @@
...
+, instructors as (
+    select
+        course_uuid
+        , array_join(array_agg(course_instructor_title), ', ') as course_instructors
+    from {{ ref('int__ocw__course_instructors') }}
+    group by course_uuid
+)
+
+select
+    courses.course_readable_id as course_short_id
```
KatelynGit, line 14: why did we rename this field? I feel like keeping names consistent is good if possible. But no worries if we needed to.

KatelynGit (approved):
I ran the builds and everything worked great there! It looks like we have all the fields they requested in the requirements ticket so that's great too. And I ran test queries against the table and that looked good as well.  I did come across some strange data but it's not at all an issue with this table in particular so it's probably ok.  It was a record with course short id = '22.01-fall-2021'.

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
rachellougee: I can't say if its useful or not as this is subjective. But the difference between `course_name` and `course_title` is course_name is part of OCW Studio urls. I think its mentioned in the description.

I keep it there as its already in https://bi.odl.mit.edu/queries/1353 @pdpinch do we need to remove it?

src/ol_dbt/models/marts/ocw/marts__ocw_courses.sql
rachellougee: So this mart is very specific to OCW, they call it short ID in their UI as they are familiar to. If this is combined mart, then it would make sense to call it course_readable_id.

I also asked Peter about this yesterday, its better to name course_short_id in this mart

rachellougee:
>  I did come across some strange data but it's not at all an issue with this table in particular so it's probably ok. It was a record with course short id = '22.01-fall-2021'.

@KatelynGit Can you say more about this? What's strange about it?

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
KatelynGit: If it's part of the query and they're pulling it separately from the url I guess that makes sense to keep it as is.

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
rachellougee: Oh I don't think I gave too much thought about metadata. I can remove that wording if its confusing. I like the `"OpenCourseWare courses are a series of classes/education materials offered by MIT"` but I am not sure if its describing what this table is. @pdpinch do you have any suggestion or if you like Kate's description, I can update it.

rachellougee: Interesting that that's how they describe these fields in ocw studio UI
image

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
rachellougee: I can see the naming confusion, so maybe we should rename it to something else, but I don't know the right name for it as course_name is used as studio url and also indirectly appear in public url  so i will wait for Peter to weight in before I change it

pdpinch: I agree that it looks repetitive, but it ends up being useful. If there's a better way to communicate that, I'm open to ideas. 

`course_title` is free text and can be edited at any time. 
`course_name` is created when the course (or site) is created in OCW Studio. It's used for constructing URLs in Studio, so if you want a report with clickable links so you can edit a course, you  will need this. Most of the time it will look like a slugified version of the course title, but not always (for example, when the course title has been changed).

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
pdpinch: I can see how "metadata" could be confusing in this context.

How about

> OCW courses published at https://ocw.mit.edu/

pdpinch (approved):
Just a couple of comments on descriptions

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
pdpinch: Let me amend that, since not all courses are published 

> OCW courses managed in OCW Studio

src/ol_dbt/models/marts/ocw/_marts__ocw__models.yml
rachellougee: Whether its called course_name or course_url_short_name, they may still be confusing. In the admin UI, course_name is labeled as Name, so I think I will keep it the way it is for now.

PR #1029, opened 2024-03-14
fix unique test for int__mitxpro__ecommerce_product

rachellougee:

NA
One of warnings in https://pipelines.odl.mit.edu/runs/4f67327e-5c5c-4e64-82c7-c266c48997a4

This fix the [unique test](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml#L971) in `int__mitxpro__ecommerce_product` to count for inactive products. 
There are 3 inactive products (program-v1:xPRO+LASERx | R16/17/18) Bon recently created, which linked to ProgramRun. He wants these 3 inactive products to be deleted but he couldn't do so due to restriction on deleting products. But we should only test the uniqueness on ["product_type", "program_id", "courserun_id"] for `active` products for this reason.

### How can this be tested?

run `dbt test --select int__mitxpro__ecommerce_product --target production`, test should now pass

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
+++ b/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml
@@ -969,6 +969,7 @@ models:
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["product_type", "program_id", "courserun_id"]
+      row_condition: "product_is_active"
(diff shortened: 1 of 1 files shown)


quazi-h (approved):
Fix looks good to me, I ran the dbt test command against main as well as this fix and it did resolve the warning.

PR #1030, opened 2024-03-15
refactor: Avoid processing empty CSV files from edx.org

blarghmatey:
Related to https://github.com/mitodl/hq/issues/1309

Airbyte is unable to process CSV files that consist only of a header. This adds logic to bypass uploading and materializing those files from the edx.org archive processing.

### How can this be tested?
Run the edxorg processing pipeline and verify that only empty CSV files are skipped.


rachellougee (approved):
Looks good. I was only able to test the if statement reading with polars locally, that seems working for empty csv file

PR #1038, opened 2024-03-18
add a new combined programs mart

KatelynGit:

Adds a new marts__combined_program_enrollment_detail table that contains program, enrollment, and certificate data from xpro, mixonline, and edx.

### How can this be tested?


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -263,3 +263,44 @@ models:
...
+  - name: programcertificate_created_on
+    description: timestamp, specifying when a certificate was initially created
+  - name: programcertificate_is_revoked
+    description: boolean, indicating whether the certificate is revoked and invalid
+  - name: programcertificate_uuid
+    description: str, unique identifier for the program certificate
+  tests:
+  - dbt_expectations.expect_compound_columns_to_be_unique:
+      column_list: ["platform_name", "program_id", "user_id", "programcertificate_created_on",
+        "programenrollment_created_on", "programcertificate_uuid"]
```
rachellougee, line 306: The uniqueness for this table shouldn't be all these 6 columns as that indicates we are missing some columns to be included in the mart. 
- For xPro, its unique on program, user and order. So we may want to bring in order here as without order, we can't tell why there are multiple enrollments for the same user and program
- For the rest, its unique on program_title and user_email

So if we want the uniqueness for the whole table, it should be program_title, user_email and order_id.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -263,3 +263,44 @@ models:
...
+    description: str, name of the platform
+    tests:
+    - not_null
+  - name: program_id
+    description: int, id representing a single program on a particular platform
+  - name: program_title
+    description: str, title of the program
+  - name: program_is_live
+    description: boolean, indicating whether the program is available to users on
+      the MITOnline website or MITxPro website
```
rachellougee, line 281: `MITx Online`

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
```diff
@@ -0,0 +1,104 @@
...
+        , null as program_readable_id
+        , edx_program_certificates.user_id
+        , users.user_edxorg_email as user_email
+        , edx_program_certificates.user_username
+        , null as programenrollment_is_active
+        , null as programenrollment_created_on
+        , null as programenrollment_enrollment_status
+        , edx_program_certificates.program_certificate_awarded_on as programcertificate_created_on
+        , null as programcertificate_is_revoked
+        , edx_program_certificates.program_uuid as programcertificate_uuid
```
rachellougee, line 98: `program_uuid` can't be the programcertificate_uuid.  
- program_uuid is unique per program
- programcertificate_uuid is unique per certificate, not per program

For edx.org program certificate, there is no uuid from 2U data. But we can either generate one in intermediate if needed.

rachellougee (changes requested):
I left a few comments below, especially the uniqueness of this table. 

Also, based on the query in the mart, I believe we are missing old DEDP enrollments/certificates that only exist in MicroMasters. This also exposes data incomplete for DEDP enrollments in intermediate, so we may want to address that before creating this mart. 

Lastly, we should discuss whether we want all the program enrollments from edX.org be null. But that may be question for @pdpinch as well.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
KatelynGit: I've simplified this and removed columns but unfortunately I don't have order_id on the table so I had to use the timestamp in it's place but it does make it unique.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
KatelynGit: fixed this thanks!

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
KatelynGit: fixed this as well

KatelynGit:
sorry had to close this... please see new PR https://github.com/mitodl/ol-data-platform/pull/1055

PR #1040, opened 2024-03-19
add primary key to micromasters program certificate models

rachellougee:

https://github.com/mitodl/hq/issues/3806

Adding `program_certificate_hashed_id` to marts__micromasters_program_certificates and its upstream models for primary key
  - For certificates sourced from MITx Online and MicroMasters, it uses the existing hash values from source since these don't change and they are unique
  - For certificates from edx.org (int__edxorg__mitx_program_certificates), it uses hash function to generate unique value from user_id and program_uuid
 
Adding unique test to ensure the new field is indeed unique

### How can this be tested?

Run the following command and tests should pass
```
dbt run --select +marts__micromasters_program_certificates
dbt test --select marts__micromasters_program_certificates
```

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_program_certificates.sql
+++ b/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_program_certificates.sql
@@ -1,9 +1,7 @@
 with completed_program_learners as (
     select
         *
-        , lower(
-           to_hex(md5(cast(cast(user_id as varchar) as varbinary) || cast(program_uuid as varbinary))) --noqa
-        ) as program_certificate_hashed_id
+        , {{ generate_hash_id('cast(user_id as varchar) || program_uuid') }} as program_certificate_hashed_id
     from {{ ref('stg__edxorg__s3__program_learner_report') }}
     where user_has_completed_program = true
 )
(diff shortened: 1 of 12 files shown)


src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_program_certificates.sql
```diff
@@ -1,5 +1,10 @@
 with completed_program_learners as (
-    select * from {{ ref('stg__edxorg__s3__program_learner_report') }}
+    select
+        *
+        , lower(
+           to_hex(md5(cast(cast(user_id as varchar) as varbinary) || cast(program_uuid as varbinary))) --noqa
```
blarghmatey, line 5: We should be using a sha256 or higher hash to avoid potential collisions. The likelihood is low with md5, but it's not guaranteed to not happen.

rachellougee: So I tested it with `sha256` but it generated way too long string e.g. 2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824. The other hash value from MITx Online application is like `a4453798ffc24f5ab98364a228e3a0fd`. When combining into one table, it looked odd, but they are unique though.

If we don't care about the uneven length, I can use `sha256`

rachellougee: Per our chat yesterday, I made the change to use sha256 hash and also applied the same hash to uuid generated from app, so that made the primary key similar length. The tests are passed from my local run

blarghmatey (approved):

PR #1041, opened 2024-03-20
add grade and course fields to marts

KatelynGit:
https://github.com/mitodl/hq/issues/3767

adds course run grade, course run passing indicator, and course title to marts__combined_course_enrollment_detail from the different platforms (bootcamps, xpro, edx, mitxonline)

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
@@ -149,7 +149,7 @@ with mitx_enrollments as (
             and mitx_enrollments.courserun_id = mitxonline_completed_orders.courserun_id
             and mitxonline_completed_orders.row_num = 1
     left join mitx_grades
-        on 
+        on
             mitx_enrollments.courserun_readable_id = mitx_grades.courserun_readable_id
             and mitx_enrollments.user_mitxonline_username = mitx_grades.user_mitxonline_username
     left join mitx_courses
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
```diff
@@ -22,6 +22,34 @@ with mitx_enrollments as (
...
+    select * from {{ ref('int__mitx__courses') }}
+)
+
+, mitxpro_courseruns as (
+    select * from {{ ref('int__mitxpro__course_runs') }}
+)
+
+, mitxpro_courses as (
+    select * from {{ ref('int__mitxpro__courses') }}
+)
```
rachellougee, line 51: Just questions, why are we importing all these courses into the mart? We've already have course run title in each platform, are xpro course run titles any different from the course title?

quazi-h (approved):
Changes lgtm, I was able to build this table against my own dev_production schema and can confirm the three new columns are present with data where available.

src/ol_dbt/models/marts/combined/marts__combined_course_enrollment_detail.sql
KatelynGit: The course titles and course run titles can be different.  For xpro for example the course title(course id=17) is 'Navigating and Leveraging Culture and Networks' but the course run associated to that(course run id =41) is 'Leading Change in Organizations'. sometimes it's not that drastic and the course run title will be the course title but with something like 'run 2' added to it or sometimes it will be the same.  But I figured it made sense to capture both values just in case we need them.

PR #1043, opened 2024-03-22
add a edx program enrollment int table

KatelynGit:

Adds a new int__edxorg__mitx_program_enrollments table that contains program enrollment data from edx.

### How can this be tested?
dbt build --select int__edxorg__mitx_program_enrollments

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -463,4 +463,3 @@ models:
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_id", "program_uuid"]
-
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -378,11 +378,6 @@ models:
   description: learners who completed MicroMasters and XSeries programs on edX.org.
     program_certificate_awarded_on might be null for some programs such as SCM
   columns:
-  - name: program_certificate_hashed_id
-    description: str, unique hash value used to identify the program certificate
-    tests:
-    - not_null
-    - unique
   - name: program_type
     description: str, the type of the program. The value are 'MicroMasters' and 'XSeries'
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
```diff
@@ -378,11 +378,6 @@ models:
   description: learners who completed MicroMasters and XSeries programs on edX.org.
     program_certificate_awarded_on might be null for some programs such as SCM
   columns:
-  - name: program_certificate_hashed_id
```
rachellougee, line 381: Is this by accident? Why are we removing this column we just added to int__edxorg__mitx_program_certificates?

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_program_enrollments.sql
```diff
@@ -0,0 +1,32 @@
...
+    select * from {{ ref('int__mitx__programs') }}
+    where is_micromasters_program = true
+)
+
+, program_learners_sorted as (
+    select
+        *
+        , row_number() over (partition by user_id, program_uuid order by courserunenrollment_created_on desc) as row_num
+    from program_learners
+)
```
rachellougee, line 15: This works. But in this model we can just do a distinct for program related columns from program_learners. I did this sort in int__edxorg__mitx_program_certificates because we want to get the latest program_certificate_awarded_on in case learners awarded same program more than one. Anyway it doesn't hurt, just saying for simplicity.

rachellougee (commented):
This looks good except one removed column. Please review

src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
KatelynGit: this should be fixed now. thanks for catching this!

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_program_enrollments.sql
KatelynGit: I did a group by instead for simplicity

rachellougee (approved):

PR #1056, opened 2024-03-27
add a new staging program enrollment table

KatelynGit:

creates a new staging table stg__micromasters__app__postgres__dashboard_programenrollment sourced from raw__micromasters__app__postgres__dashboard_programenrollment. We need this data to fill in a data gap in the new program data mart.

### How can this be tested?
dbt build --select stg__micromasters__app__postgres__dashboard_programenrollment

### DIFF
--- a/src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
+++ b/src/ol_dbt/models/staging/micromasters/_micromasters__sources.yml
@@ -568,3 +568,6 @@ sources:
       description: int, foreign key for users_user
     - name: program_id
       description: int, foreign key to courses_program
+    - name: share_hash
+      description: str, share hash for the program enrollment 
+        example:72ff85f6-b88d-4636-830a-69cf59f896f5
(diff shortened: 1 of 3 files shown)


quazi-h (approved):
Changes and logic looks are straightforward and look good to me.
I ran dbt build locally and observed the new raw source table (raw__micromasters__app__postgres__dashboard_programenrollment) as well as the new staging table (stg__micromasters__app__postgres__dashboard_programenrollment) with both containing valid data. The only thing I wasn't sure about was the "share_hash" column on the new raw table because there was only 1378 records out of 166511 that were not null. Since that column isn't detailed in the yaml file, I assume it is one of the ones airbyte generates automatically during a sync?

KatelynGit:
> Changes and logic looks are straightforward and look good to me. I ran dbt build locally and observed the new raw source table (raw__micromasters__app__postgres__dashboard_programenrollment) as well as the new staging table (stg__micromasters__app__postgres__dashboard_programenrollment) with both containing valid data. The only thing I wasn't sure about was the "share_hash" column on the new raw table because there was only 1378 records out of 166511 that were not null. Since that column isn't detailed in the yaml file, I assume it is one of the ones airbyte generates automatically during a sync?

Thanks @quazi-h.  That field is part of the front end but I think it's mostly used internally only.  I added it but sounds like we won't need it in the intermediate layer if we add this data there in the future because as you mentioned above it's pretty sparsely populated.

PR #1057, opened 2024-03-28
add course readable id to combined enrollment mart and upstream models

rachellougee:

Part of https://github.com/mitodl/hq/issues/3875

Adding course_readable_id to marts__combined_course_enrollment_detail and upstream changes
-  some null for some old edx.org runs due to missing data from BigQuery mitx_course table. This hopefully will get resolved once we use our own edx.org data
- null bootcamps for now until we add it in bootcamps app https://github.com/mitodl/hq/issues/3905

### How can this be tested?

```
dbt build --select +marts__combined_course_enrollment_detail
```


quazi-h (approved):
Changes look good, I ran this locally and did not hit any errors with the stg__edxorg__bigquery__mitx_courserun table. Tests passed without warnings.

PR #1064, opened 2024-04-02
Add user profile fields to enrollment mart and micromasters intermediate

rachellougee:

Part of https://github.com/mitodl/hq/issues/3875

- adding `user_birth_country` to int__micromasters__users and its upstream model (This field is only on MM at the moment)
- Adding a few user profile fields like user_highest_education, user_company and user_address_country to `marts__combined_course_enrollment_detail`

### How can this be tested?

```
dbt run --select int__micromasters__users
dbt run --select +marts__combined_course_enrollment_detail
```


rachellougee:
> The other thing is just has a question- why do we rename user address country to user country code? Thanks

It was renamed when creating this model, not as part of this PR. I don't recall the specific reason, but probably its actually a code.  I think user_address_country_code would make sense but then its inconsistent with other models.

PR #1071, opened 2024-04-10
Update the Micromasters Summary Mart

KatelynGit:
https://github.com/mitodl/hq/issues/3934

Update the logic for the Micromasters Summary Mart to remove revoked certificates, update logic, and combine programs that are reported on the same way. Also made updates to filter revoked certificates from int__micromasters__program_certificates and int__micromasters__course_certificates and add a new dedp indicator to int__micromasters__program_certificates, int__micromasters__course_certificates, and int__micromasters__course_enrollments.

### How can this be tested?
dbt build --select +marts__micromasters_summary


src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
```diff
@@ -139,7 +145,17 @@ with course_certificates_dedp_from_micromasters as (
         , courseruncertificate_download_uuid as courseruncertificate_uuid
         , courseruncertificate_download_url as courseruncertificate_url
         , courseruncertificate_created_on
+        , if(mitxonline_program_id in (1, 2, 3), true, false) as program_is_dedp
     from course_certificates_non_dedp_program
 )
 
-select * from program_course_certificates
+select course_certificates.*
```
rachellougee, line 152: Just wanted to point out these changes are not necessary as `int__micromasters__course_certificates` has already filtered out courseruncertificate_is_revoked in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_certificates_dedp_from_mitxonline.sql#L4.  Can you take a look?

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
```diff
@@ -139,7 +145,17 @@ with course_certificates_dedp_from_micromasters as (
         , courseruncertificate_download_uuid as courseruncertificate_uuid
         , courseruncertificate_download_url as courseruncertificate_url
         , courseruncertificate_created_on
+        , if(mitxonline_program_id in (1, 2, 3), true, false) as program_is_dedp
```
rachellougee, line 148: If I am reading correctly, this is `course_certificates_non_dedp_program` subquery, so this won't have DEDP program. It can just be true

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
```diff
@@ -119,6 +124,7 @@ with course_certificates_dedp_from_micromasters as (
         , courseruncertificate_uuid
         , courseruncertificate_url
         , courseruncertificate_created_on
+        , if(mitxonline_program_id in (1, 2, 3), true, false) as program_is_dedp
```
rachellougee, line 127: Same here. This is `dedp_course_certificates` so all of the programs here are DEDP. Also, there is already a field `is_dedp_program` in int__mitx__programs.

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
```diff
@@ -95,24 +100,33 @@ with program_certificates_dedp_from_micromasters as (
...
+    , report.user_full_name
+    , report.user_year_of_birth
+    , report.user_country
+    , report.user_address_postal_code
+    , report.user_street_address
+    , report.user_address_state_or_territory
+    , if(report.mitxonline_program_id in (1, 2, 3), true, false) as program_is_dedp
 from report
-order by program_completion_timestamp desc
+left join mitxonline_program_certificates
```
rachellougee, line 124: These changes are also not necessary either. I believe all we need to do is add to where clause in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_program_certificates_dedp_from_mitxonline.sql#L60. That would filter these revoked certificates.

rachellougee:
To make it simple for this PR,
 - Add `where mitxonline_program_certificates.programcertificate_is_revoked` to https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_program_certificates_dedp_from_mitxonline.sql#L60. That will also solve https://github.com/mitodl/hq/issues/3821.
- Import  `int__mitx__programs` into `marts__micromasters_summary` it should have the fields we need for simplifying the queries around DEDP,  'Statistics and Data Science',  'Statistics and Data Science (General track)' and 'MIT Finance' values in the PR. 
Let me know if that makes sense

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
KatelynGit: i removed those. thanks

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
KatelynGit: i think you meant false since it's not dedp so i updated it like that

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
KatelynGit: i updated that as well

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
KatelynGit: I know we could filter in the table sql or the subquery sql. Either would work using table filter for now.

KatelynGit:
> To make it simple for this PR,
> 
> * Add `where mitxonline_program_certificates.programcertificate_is_revoked=false` to https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_program_certificates_dedp_from_mitxonline.sql#L60. Other changes are not necessary unless I misread it.  That will also solve [Revoked MicroMasters program certificates  hq#3821](https://github.com/mitodl/hq/issues/3821).
> * Import  `int__mitx__programs` into `marts__micromasters_summary` it should have the fields we need for simplifying the queries around `program_is_dedp`,  'Statistics and Data Science',  'Statistics and Data Science (General track)' and 'MIT Finance' values in the PR.
>   Let me know if that makes sense

I added int mix programs instead of the hard coding. thanks!

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
rachellougee: My reason is that it defeats the purpose of `__micromasters_program_certificates_dedp_from_micromasters` if we are adding the same logics in two places. I believe the subqueries folder is created for simplifying query. But if you think adding it here makes sense, I am fine too

rachellougee: One line change in `__micromasters_program_certificates_dedp_from_mitxonline` vs adding all these joins just to filter `mitxonline_program_certificates.programcertificate_is_revoked=false` which could be done in __micromasters_program_certificates_dedp_from_mitxonline

KatelynGit: To me it was confusing since there are a bunch of different subqueries it looked simplier to add the one join in the main table SQL

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
rachellougee: yes, I meant false. Can you point me where you update it? I am not seeing it?

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
rachellougee: I don't mean to be nitpicking. I think this creates more confusion. Since `int__micromasters__course_certificates` handles revoke certificates in subquery with just a filter https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__micromasters_course_certificates_dedp_from_mitxonline.sql#L4

 But we handle revoked program certificates in `int__micromasters__program_certificates` with duplicate logic. Can I ask what is the confusion about adding `__micromasters_program_certificates_dedp_from_mitxonline`?

src/ol_dbt/models/marts/micromasters/_marts_micromasters__models.yml
```diff
@@ -164,7 +164,7 @@ models:
   - name: program_certificates
     description: int, count of program certificates earned for the the program
 
-- name: marts__micromasters_summary
+- name: marts__micromasters__summary
```
rachellougee, line 167: Can we put back `marts__micromasters_summary` instead of deleting it? dbt doesn't delete the model, so we might end up having two marts unless someone deletes it manully

rachellougee (approved):
I will approve it for now. If you address these comments, feel free to merge it.

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
KatelynGit: the subquery didn't have the table you mentioned so i added the join there to the table that contains the revoked flag and i added the flag as a filter. I removed the join and filter from the main table sql.

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_certificates.sql
KatelynGit: line 143 -         , false as program_is_dedp

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_enrollments.sql
```diff
@@ -24,6 +24,7 @@ select
     , mitx_enrollments.courserunenrollment_is_active
     , mitx_enrollments.courserun_title
     , mitx_enrollments.courserunenrollment_enrollment_mode
+    , if(mitx_enrollments.mitxonline_program_id in (1, 2, 3), true, false) as program_is_dedp
```
rachellougee, line 27: @KatelynGit Another FYI I forgot to comment, this line can be replaced with `programs.is_dedp_program`. I believe I mentioned it in other comments that `int__mitx__programs` has is_dedp_program already

src/ol_dbt/models/intermediate/micromasters/int__micromasters__program_certificates.sql
KatelynGit: I think it's the correct one now.  The subqueries are adding some complexity

src/ol_dbt/models/marts/micromasters/_marts_micromasters__models.yml
KatelynGit: I'ver changed this

src/ol_dbt/models/intermediate/micromasters/int__micromasters__course_enrollments.sql
KatelynGit: this code is working well so I'm going to leave it as is

PR #1072, opened 2024-04-10
update user_company for xpro users to use ecommerce_company

rachellougee:

NA. 

Peter mentioned that xPro has a standardized [ecommerce_company](https://xpro.mit.edu/admin/ecommerce/company/ ) so I use that to populate user_company for xPro users. I created https://github.com/mitodl/hq/issues/3987 for the messy data.

Update user_company in `int__mitxpro__users` to use the standardized ecommerce_company if it is populated in their enrollments. This will flow into any upstream intermediate and marts.

### How can this be tested?

```
dbt build --select int__mitxpro__users
```


KatelynGit (approved):
I ran the build it was clean, code looked good, and it looks like it did in fact improve the data quality in a bunch of instances which is great.  We still may have some bad data because of the text field but this definitely improved it.

PR #1074, opened 2024-04-11
add readable_id and make it unique in bootcamps course staging model

rachellougee:

Close https://github.com/mitodl/hq/issues/3905

Updating `readable_id` with source data to stg__bootcamps__app__postgres__courses_course, and making it unique field in dbt test
Note that there is currently no data for this field until back-populating is done in bootcamps repo. But data will just flow through at that point

### How can this be tested?

dbt build --select stg__bootcamps__app__postgres__courses_course


KatelynGit (approved):
the build ran clean.  pretty straight forward change and code LGTM

PR #1075, opened 2024-04-12
Add courserun_is_current to combined course runs and enrollment mart

rachellougee:

The last code change for https://github.com/mitodl/hq/issues/3875

Adding `courserun_is_current` and the relevant date fields to int__combined__course_runs and marts__combined_course_enrollment_detail
  - `courserun_is_current` is true if courserun_start_on is in the past and blank courserun_end_on, OR courserun_start_on
      is in the past and courserun_end_on is in the future.
  -  there are some old edx.org runs e.g.  MITx/8.02.2x/2T2019 where blank courserun_end_on with courserun_start_on is in the past, but that is data issue we get from BigQuery

### How can this be tested?

dbt build --select int__combined__course_runs

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -162,7 +162,13 @@ models:
     description: str, unique string to identify a course run on the corresponding
       platform
   - name: courserun_start_on
-    description: timestamp, date and time when the course starts
+    description: timestamp, date and time when the course run starts
+  - name: courserun_end_on
+    description: timestamp, date and time when the course run ends
+  - name: courserun_is_current
+    description: boolean, indicating if the course run is currently running. True
+      if courserun_start_on is in the past and blank courserun_end_on, or courserun_start_on
+      is in the past and courserun_end_on is in the future.
   - name: user_username
     description: string, username to identify a user on the corresponding platform
     tests:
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -162,7 +162,13 @@ models:
...
       platform
   - name: courserun_start_on
-    description: timestamp, date and time when the course starts
+    description: timestamp, date and time when the course run starts
+  - name: courserun_end_on
+    description: timestamp, date and time when the course run ends
+  - name: courserun_is_current
+    description: boolean, indicating if the course run is currently running. True
+      if courserun_start_on is in the past and blank courserun_end_on, or courserun_start_on
+      is in the past and courserun_end_on is in the future.
```
quazi-h, line 171: Would these columns benefit from any tests? I noticed you added a not_null test on the courserun_is_current on the upstream intermediate table but I suppose that adding the same test here would be redundant since this is populating from that same upstream table, right?

rachellougee: yes, it's the same test. But I can add to mart as well

rachellougee: Added. I think it is a good call to add any test to the mart, so mart users don't have to dig into upstream models to figure out

quazi-h (approved):
New columns look good to me, I was able to run this locally with no issues. I confirmed the new columns are populated as expected in my test schema.

PR #1083, opened 2024-04-18
simplify Micromasters dedp indicator in intermediate and mart tables

KatelynGit:
https://github.com/mitodl/hq/issues/3934

Eliminate dedp indicator field from int__micromasters__course_certificates it's only used in the marts__micromasters_summary and there we added code to pull it from int__mitx__programs there. Replaced complicated if statement on int__micromasters__course_enrollments with the is_dedp_program field on int__mitx__programs.  These changes are based on feedback from Rachel as well.

### How can this be tested?
dbt build --select int__micromasters__course_certificates
dbt build --select int__micromasters__course_enrollments
dbt build --select marts__micromasters_summary

### DIFF
--- a/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
@@ -297,9 +297,6 @@ models:
       created
     tests:
     - not_null
-  - name: program_is_dedp
-    description: boolean, specifying if the program is DEDP from the mixonline program
-      id
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_email", "courserun_readable_id", "micromasters_program_id",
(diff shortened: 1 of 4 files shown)


quazi-h (approved):
Changes here look good to me.
I was able to run the dbt build commands with no errors.
I confirmed that the 3 separate DEDP categories were consolidated into one in the final mart tables (see attached screenshots). The only thing I wasn't sure about is that I was expecting the total counts to be the same but looking at the changes in logic, we are pulling from int__mitx__programs instead of int__micromasters__course_certificates, which may be why?

DEDP Counts before fix
DEDP counts consolidated

KatelynGit:
I took a look the counts and I have purposely improved them in the latest version.  in the old version this single enrollment was counting twice when it should have only been counted once:
select *
from int__micromasters__course_enrollments
where courserun_readable_id= 'MITx/14.73x/1T2021'
and user_email= 'mohitoshk@gmail.com'

PR #1084, opened 2024-04-18
remove salesforce dbt models

KatelynGit:
Salesforce passwords have expired https://airbyte.odl.mit.edu/workspaces/3f0ccc59-baa3-4567-92e5-c61fb737467b/connections/ad3f786c-54b7-4aa9-a29e-521c3a47af72/job-history#8746::0
We need to pause that pipeline. This ticket is to remove all the staging and intermediate salesforce dbt models we currently have. This is an urgent need.

### How can this be tested?
Check to see all salesforce dbt models are now missing (see ol_warehouse_production_kdelaney_staging and ol_warehouse_production_kdelaney_intermediate). And run:

dbt build --select int__mitxpro__b2becommerce_b2border


blarghmatey (changes requested):
Rather than deleting these models entirely, it might be better to set them up in a `.dbtignore` so that they get skipped for the time being while maintaining the code in the repo. https://docs.getdbt.com/reference/dbtignore

KatelynGit:
Tobias I remove the models but all the code is saved. I'll put that elsewhere

KatelynGit:
> Rather than deleting these models entirely, it might be better to set them up in a `.dbtignore` so that they get skipped for the time being while maintaining the code in the repo. https://docs.getdbt.com/reference/dbtignore

o ok I can do that instead

KatelynGit:
@blarghmatey I've added a dbtignore file but since I've never done that before let me know if it looks ok to you

KatelynGit:
@blarghmatey I've removed the salesforce field from int__mitxpro__b2becommerce_b2border. I'll open a new issue with the exact code we previous had included so when we get sales force back it'll be an easy change just to add it back.

src/ol_dbt/.dbtignore
```diff
@@ -0,0 +1,10 @@
+# .dbtignore
+
+# ignore individual files
+int__salesforce__opportunitylineitem.sql
+int__salesforce__opportunity.sql
+stg__salesforce__opportunitylineitem.sql
+stg__salesforce__opportunity.sql
+_salesforce__sources.yml
+_stg_salesforce__models.yml
+_int_salesforce__models.yml
```
rachellougee, line 10: It can be directories instead of listing all the files. Either way would work.
```
models/staging/salesforce
models/intermediate/salesforce
```

rachellougee (approved):

PR #1095, opened 2024-04-23
remove test for micromasters users without profiles

KatelynGit:
https://github.com/mitodl/hq/issues/4062

There is currently a test to check if user_profile_id on __micromasters__users is null. However, users can exist without profiles so we should remove the test.

### How can this be tested?
dbt build --select __micromasters__users


rachellougee:
@KatelynGit Does this still need a review? Per our discussion in the office, removing the test is probably not the solution to the airbyte sync issues. But let me know.

PR #1107, opened 2024-05-02
staging edxorg course enrollment and grade

rachellougee:

Part of https://github.com/mitodl/hq/issues/4067, the rest of the raw tables mentioned in the issue are not ready yet

Adding `stg__edxorg__s3__courserun_grade` and `stg__edxorg__s3__courserun_enrollment` 
  - fields are converted to the proper data types as raw fields are all ingested as string
  - data is deduplicated

### How can this be tested?

 dbt build --select stg__edxorg__s3__courserun_enrollment
 dbt build --select stg__edxorg__s3__courserun_grade

### DIFF
--- a/src/ol_dbt/macros/apply_deduplication_query.sql
+++ b/src/ol_dbt/macros/apply_deduplication_query.sql
@@ -1,15 +1,15 @@
-{% macro deduplicate_query(cte_name1='source', cte_name2='most_recent_source') %}
+{% macro deduplicate_query(cte_name1='source', cte_name2='most_recent_source', partition_columns='id') %}
     /*
         Add additional queries to handle duplicated data introduced by airbyte sync mode "Incremental Sync - Append"
         where cursor is set to a timestamp field. The deduplication logic works like this:
-        - If there are multiple copies of the same record based on primary key id, we will use the record from most
+        - If there are multiple copies of the same record based on partition_columns, we will use the record from most
           recent sync in the source table.
-        - If there is only one record for the same primary key id, it will just load that record as it is.
+        - If there is only one record for the same partition_columns, it will just load that record as it is.
     */
    , source_sorted as (
         select
             *
-            , row_number() over ( partition by id order by _airbyte_emitted_at desc) as row_num
+            , row_number() over ( partition by {{ partition_columns }} order by _airbyte_emitted_at desc) as row_num
         from {{ cte_name1 }}
     )
 
--- a/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
+++ b/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
@@ -516,3 +516,79 @@ models:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_username", "useractivity_context_object", "useractivity_event_source",
         "useractivity_event_type", "useractivity_event_object", "useractivity_timestamp"]
+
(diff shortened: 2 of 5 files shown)


src/ol_dbt/macros/apply_deduplication_query.sql
```diff
@@ -1,15 +1,15 @@
-{% macro deduplicate_query(cte_name1='source', cte_name2='most_recent_source') %}
+{% macro deduplicate_query(cte_name1='source', cte_name2='most_recent_source', partition_columns='id') %}
```
rachellougee, line 1: The change here doesn't affect the existing models that are currently using this as `id` is the default option

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
```diff
@@ -516,3 +516,79 @@ models:
...
+  - name: user_id
+    description: int, user ID on edX.org
+    tests:
+    - not_null
+  - name: courserunenrollment_mode
+    description: str, indicating what kind of enrollment it is. Possible values are
+      honor, audit, credit, verified, unpaid-executive-education, professional, no-id-professional.
+    tests:
+    - not_null
+  - name: courserunenrollment_is_active
```
KatelynGit, line 541: It looks like this is showing up as an integer in the table when created instead of a string. I'd also specify in the description that this will be a 1 or a 0 value.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
```diff
@@ -516,3 +516,79 @@ models:
...
+    description: str, indicating whether this enrollment is active
+    tests:
+    - not_null
+  - name: courserunenrollment_created_on
+    description: timestamp, time when this enrollment was created
```
KatelynGit, line 546: I'd change it to: 
 description: timestamp, time and date when this enrollment was created

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
```diff
@@ -516,3 +516,79 @@ models:
...
+  - name: courserungrade_letter_grade
+    description: str, calculated course grade as a letter value (e.g., A-D, Pass),
+      per grading policy If the learner's grade is Fail or F, this value is empty.
+  - name: courserungrade_first_passed_on
+    description: timestamp, date and time when the learner first passed the course.
+      If this value is empty, the learner never passed the course. If this value is
+      non-empty but the letter_grade value is empty, the learner transitioned from
+      passing to not passing.
+  - name: courserungrade_created_on
+    description: timestamp, time when the course grade was first calculated for this
```
KatelynGit, line 583: I'd change it from "time" to "time and date" for timestamp descriptions on this table

KatelynGit (commented):
The builds all looked good I just left some comments on the descriptions

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
rachellougee: I think this should be `boolean` to be consistent with other data sources like mitxonline. So I cast to boolean instead of integer.

rachellougee:
@KatelynGit Thanks for the review. It should be all set. But let me know anything else for the description

KatelynGit (changes requested):
Looks good

PR #1115, opened 2024-05-07
add missing mart program enroll records

KatelynGit:
https://github.com/mitodl/hq/issues/4061

Adds old DEDP enrollment data to the marts__combined_program_enrollment_detail. And adds a new platform name field to  the int__micromasters__program_enrollments table.

### How can this be tested?
dbt build --select +marts__combined_program_enrollment_detail

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
@@ -118,7 +118,7 @@ with mitxpro__programenrollments as (
 )
 
 , mm_no_dups as (
-    select
+    select 
         micromasters__program_enrollments.platform_name
         , micromasters__program_enrollments.micromasters_program_id as program_id
         , micromasters__program_enrollments.program_title
@@ -137,16 +137,16 @@ with mitxpro__programenrollments as (
     inner join micromasters__users
         on micromasters__program_enrollments.micromasters_user_id = micromasters__users.user_id
     left join combined_programs
-        on
+        on 
             micromasters__program_enrollments.user_email = combined_programs.user_email
             and micromasters__program_enrollments.program_title = combined_programs.program_title
-    where
+    where 
         combined_programs.user_email is null
         and combined_programs.program_title is null
         and micromasters__program_enrollments.platform_name = 'micromasters'
 )
 
-select
+select 
     combined_programs.platform_name
     , combined_programs.program_id
     , combined_programs.program_title
@@ -165,7 +165,7 @@ from combined_programs
 
 union all
 
-select
+select 
     mm_no_dups.platform_name
     , mm_no_dups.program_id
     , mm_no_dups.program_title
--- a/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
@@ -118,7 +118,7 @@ with mitxpro__programenrollments as (
 )
 
 , mm_no_dups as (
-    select 
+    select
         micromasters__program_enrollments.platform_name
         , micromasters__program_enrollments.micromasters_program_id as program_id
         , micromasters__program_enrollments.program_title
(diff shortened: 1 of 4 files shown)


rachellougee (commented):
Leaving notes from yesterday.  

I ran the dbt code and reviewed the data, here are some issues I found :
- around 45k null user_emails in `int__micromasters__program_enrollments`, which has warned in dbt build 
```
select * from  "ol_data_lake_production"."ol_warehouse_production_rlougee_intermediate"."int__micromasters__program_enrollments"
where user_email is null
```
These seem to be MITx Online users who don't exist in MicroMasters. Their email should be pulled from MITx Online.

- `marts__combined_program_enrollment_detail` is missing certificates for those MicroMasters learners who don't have accounts in MITx Online
```
select * from  "ol_data_lake_production"."ol_warehouse_production_rlougee_intermediate".int__micromasters__program_certificates certificates
left join "ol_data_lake_production"."ol_warehouse_production_rlougee_mart".marts__combined_program_enrollment_detail enrollments  
on certificates.user_email = enrollments.user_email and certificates.micromasters_program_id = enrollments.program_id
where certificates.micromasters_program_id=2
and enrollments.programcertificate_created_on is null

```
The following comment is what we talked yesterday, but note here so you can review them all
- `micromasters` is not a platform, it's a program type. So we probably want to use either `'{{ var("mitxonline") }}'` or `'{{ var("edxorg") }}'` to refer those micromasters data you are trying to pull in and decide which source of data to represent their user_id, user_email and username in the mart

KatelynGit:
@rachellougee I think I fixed the null email issue in int__micromasters__program_enrollments. I also removed the staging table as a source for the mart since we discussed that yesterday. I also fixed the platform issue to remove 'micromasters'.  Let me know if you have any questions or concerns. Thanks!

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
```diff
@@ -109,4 +117,67 @@ with mitxpro__programenrollments as (
...
+    inner join micromasters__users
+        on micromasters__program_enrollments.micromasters_user_id = micromasters__users.user_id
+    left join combined_programs
+        on
+            micromasters__program_enrollments.user_email = combined_programs.user_email
+            and micromasters__program_enrollments.program_title = combined_programs.program_title
+    where
+        combined_programs.user_email is null
+        and combined_programs.program_title is null
+        and micromasters__program_enrollments.platform_name = 'micromasters'
```
rachellougee, line 146: Should this be '{{ var("edxorg") }}' instead of 'micromasters'?

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
```diff
@@ -109,4 +117,67 @@ with mitxpro__programenrollments as (
...
+, mm_no_dups as (
+    select
+        micromasters__program_enrollments.platform_name
+        , micromasters__program_enrollments.micromasters_program_id as program_id
+        , micromasters__program_enrollments.program_title
+        , null as program_is_live
+        , null as program_readable_id
+        , micromasters__users.user_id
+        , micromasters__program_enrollments.user_email
+        , micromasters__users.user_username
```
rachellougee, line 129: Should user_id, and user_username be pulled from their edx.org accounts instead of micromasters?

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
KatelynGit: I changed it this should be good now

src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
KatelynGit: I changed both to edx

rachellougee (approved):
I will approve for now. But we should fix the unique test failure on the mart due to the platform and profile mismatch. Otherwise, it looks good.

KatelynGit:
> I will approve for now. But we should fix the unique test failure on the mart due to the platform and profile mismatch. Otherwise, it looks good.

I fixed the unique test failure

PR #1119, opened 2024-05-08
add number_of_passed_courses to user mart and upstream

rachellougee:

NA but related to https://bi.ol.mit.edu/superset/dashboard/11/ 

Adding  `num_of_course_passed` and `num_of_course_enrolled` to `marts__combined__users` and its upstream changes so that the filter # of courses passed used on https://bi.ol.mit.edu/superset/dashboard/11/ can directly come from dbt model rather than calculating it on the superset

### How can this be tested?

dbt build int__combined__courserun_enrollments
dbt build marts__combined__users


KatelynGit (approved):
looks good to me and builds ran clean

PR #1121, opened 2024-05-10
Update enrollment mart to use combined intermediate

rachellougee:

NA

This moves some of the joins in marts__combined_course_enrollment_detail to intermediate combined models, simplifying the mart. 

### How can this be tested?

Run the following commands, tests should all pass
dbt build --select int__combined__courserun_certificates
dbt build --select int__combined__courserun_enrollments

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/_combined_models.yml
+++ b/src/ol_dbt/models/intermediate/combined/_combined_models.yml
@@ -57,6 +57,9 @@ models:
     - not_null
     - accepted_values:
         values: '{{ var("platforms") }}'
+  - name: courserunenrollment_id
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/intermediate/combined/_combined_models.yml
```diff
@@ -95,11 +102,17 @@ models:
     - not_null
   - name: user_full_name
     description: str, user full name on user's profile on the corresponding platform
-  - name: user_has_certificate
```
rachellougee, line 98: Currently, this field is not used by any model, so this is replaced by courseruncertificate_is_earned

quazi-h (approved):
Most recent fixes have addressed the failing test.
Everything is passing now.

PR #1130, opened 2024-05-14
Staging edxorg users, profiles, courseaccessroles and certificates

rachellougee:

remaining https://github.com/mitodl/hq/issues/4067

Creating `stg__edxorg__s3__user` and `stg__edxorg__s3__user_profile`, `stg__edxorg__s3__user_courseaccessrole` and `stg__edxorg__s3__courserun_certificate`
- fields are converted to the proper data types and cleaned
- data is deduplicated

### How can this be tested?

Run these commands and there should be no errors

dbt build --select stg__edxorg__s3__user
dbt build --select stg__edxorg__s3__user_profile
dbt build --select stg__edxorg__s3__user_courseaccessrole
dbt build --select stg__edxorg__s3__courserun_certificate

### DIFF
--- a/src/ol_dbt/dbt_project.yml
+++ b/src/ol_dbt/dbt_project.yml
@@ -46,9 +46,10 @@ vars:
     highest_education_values: ['Doctorate', "Master''s or professional degree", "Bachelor''s\
         \ degree", 'Associate degree', 'Secondary/high school', 'Junior secondary/junior
         high/middle school', 'Elementary/primary school', 'No formal education', 'Other
-        education', '']
+        education', "Doctorate in science or engineering", "Doctorate in another field",
+      '']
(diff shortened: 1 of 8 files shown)


KatelynGit (approved):
The builds all looked good no errors or warnings. The definitions and data looked good to me too.  I guess just a question- do you think we should have relationship data checks in the dbt?

rachellougee:
> The builds all looked good no errors or warnings. The definitions and data looked good to me too. I guess just a question- do you think we should have relationship data checks in the dbt?

It's a good idea to add relationship checks but given the nature of this dataset and we are still loading missing data, we can add that later if needed once data are completely loaded.  This PR is for staging and cleaning the data. There will be additional work in https://github.com/mitodl/hq/issues/4263 to validate this data

PR #1138, opened 2024-05-22
update design of enrollment and order marts

KatelynGit:

Changed order of enrollment mart to alphabetical for readability. Moved order columns to orders mart from enrollment mart it makes more sense there design wise. Improved definition.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
@@ -49,6 +49,11 @@ with bootcamps__ecommerce_order as (
         , mitxpro__ecommerce_allorders.receipt_authorization_code
         , mitxpro__ecommerce_allorders.receipt_transaction_id
         , mitxpro__ecommerce_allorders.req_reference_number
+        , mitxpro__ecommerce_order.order_tax_country_code
+        , mitxpro__ecommerce_order.order_tax_rate
+        , mitxpro__ecommerce_order.order_tax_rate_name
+        , mitxpro__ecommerce_order.order_tax_amount
+        , mitxpro__ecommerce_order.order_total_price_paid_plus_tax
         , coalesce(mitxpro__ecommerce_allorders.coupon_id, mitxpro__ecommerce_allorders.b2bcoupon_id) as coupon_id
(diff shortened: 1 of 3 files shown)


src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -103,6 +108,11 @@ with bootcamps__ecommerce_order as (
         , payment_authorization_code as receipt_authorization_code
         , payment_transaction_id as receipt_transaction_id
         , order_reference_number as req_reference_number
+        , null as order_tax_country_code
+        , null as order_tax_rate
+        , null as order_tax_rate_name
+        , null as order_tax_amount
+        , null as order_total_price_paid_plus_tax
```
rachellougee, line 115: Per the updated description `For other platform, this is the same as order_total_price_paid`, this shouldn't be null, right?

rachellougee (commented):
For re-arranging the field order in marts__combined_course_enrollment_detail, I think we should also consider the primary key for the table, not just sort alphabetically? These coupon-related fields are now placed first, but these are not primary keys for the table and most are NULL. I don't have a strong opinion but from a user standpoint I would expect `platform` and `courserunenrollment_id` to be the first columns.

Just a question, why don't we sort `marts__combined__orders` alphabetically for consistency?

src/ol_dbt/models/marts/combined/marts__combined__orders.sql
KatelynGit: I've changed this tp pull the order_total_price_paid for the other platforms should be fixed now

KatelynGit:
> For re-arranging the field order in marts__combined_course_enrollment_detail, I think we should also consider the primary key for the table, not just sort alphabetically? These coupon-related fields are now placed first, but these are not primary keys for the table and most are NULL. I don't have a strong opinion but from a user standpoint I would expect `platform` and `courserunenrollment_id` to be the first columns.
> 
> Just a question, why don't we sort `marts__combined__orders` alphabetically for consistency?

I've also changed the order mart so it's alphabetical order.  But I've also changed both to add a few key joining columns at the top.

rachellougee (approved):

PR #1139, opened 2024-05-22
Create problem and video activities intermediate models for edX.org

rachellougee:

https://github.com/mitodl/hq/issues/3713

- This adds a few user course activity models sourced from edx.org tracking logs. These are similar to MITx Online.
- Update `int__edxorg__mitx_user_courseactivities` to use our edx.org tracking logs. No upstream model uses it, so there should be no impact.


quazi-h (approved):
Changes look good to me. Sorry for the delay in the review, I had to reset my python environments and am now using virtual env for testing dbt. All the commands ran successfully with no issues.

PR #1151, opened 2024-05-30
Create intermediate OVS videos to populate video duration

rachellougee:

Close https://github.com/mitodl/hq/issues/4234

Creating `int__ovs__videos` so that it can be used to pull video duration from videos used on MITx Online and xPro

### How can this be tested?

dbt build --select int__ovs__videos


rachellougee:
> This is more just an observation but I wasn't sure what OVS stood for maybe we can add that in the table description instead of just the acronym? up to you though

I updated OVS to ODL Video Service to make it clear

PR #1159, opened 2024-06-05
add columns to edx learners

KatelynGit:
https://github.com/mitodl/hq/issues/4460
https://github.com/mitodl/hq/issues/4425

This adds the columns latest_program_cert_award_on and ever_completed_program to the stg__edxorg__s3__program_learner_report table. Currently only 4 records are making it into the intermediate program cert table (see https://github.com/mitodl/hq/issues/4460) because the current record program completion boolean is just set to 'false' when the person actually completed the program. We need a way to identify if they ever completed a program.

there will also be other intermediate changes following this PR

### How can this be tested?
dbt build --select stg__edxorg__s3__program_learner_report

### DIFF
--- a/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
+++ b/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
@@ -438,6 +438,12 @@ models:
   - name: courserunenrollment_upgraded_on
     description: timestamp, the date in ISO-8601 format that the learner first upgraded
       to the verified track course mode.
+  - name: latest_program_cert_award_on
+    description: timestamp, the latest date and time that the learner has passed
+      each of the courses in the program and was awarded a program certificate.
+  - name: ever_completed_program
+    description: boolean, whether the learner at any point has passed each of the 
+      courses in the program with a grade that qualifies them for a verified certificate.
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
       column_list: ["user_id", "courserun_readable_id", "program_uuid"]
--- a/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
+++ b/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
@@ -439,10 +439,10 @@ models:
     description: timestamp, the date in ISO-8601 format that the learner first upgraded
       to the verified track course mode.
   - name: latest_program_cert_award_on
-    description: timestamp, the latest date and time that the learner has passed
-      each of the courses in the program and was awarded a program certificate.
+    description: timestamp, the latest date and time that the learner has passed each
+      of the courses in the program and was awarded a program certificate.
   - name: ever_completed_program
-    description: boolean, whether the learner at any point has passed each of the 
+    description: boolean, whether the learner at any point has passed each of the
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql
```diff
@@ -5,13 +5,27 @@ with source as (
 , source_sorted as (
     select
         *
+        , case
+            when "date program certificate awarded" = 'null' then null
+            else to_iso8601(date_parse("date program certificate awarded", '%Y-%m-%dT%H:%i:%sZ'))
+        end as program_certificate_awarded_dt
```
rachellougee, line 11: Since we are casting this column here, we probably don't need to cast it again in the `cleaned` subquery. But it's probably not a big deal.

rachellougee (commented):
The code works except ever_completed_program is not a boolean. 

Also just a question, do we need the existing columns `user_has_completed_program` and `program_certificate_awarded_on` if the downstream intermediate model is going to use these two new columns instead?

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
```diff
@@ -438,6 +438,12 @@ models:
   - name: courserunenrollment_upgraded_on
     description: timestamp, the date in ISO-8601 format that the learner first upgraded
       to the verified track course mode.
+  - name: latest_program_cert_award_on
+    description: timestamp, the latest date and time that the learner has passed each
+      of the courses in the program and was awarded a program certificate.
+  - name: ever_completed_program
+    description: boolean, whether the learner at any point has passed each of the
```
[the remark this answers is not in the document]
KatelynGit: good catch. I changes this to cast to boolean

src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql
KatelynGit: I removed the extra code like you mentioned to simplify this

KatelynGit:
> The code works except ever_completed_program is not a boolean.
> 
> Also just a question, do we need the existing columns `user_has_completed_program` and `program_certificate_awarded_on` if the downstream intermediate model is going to use these two new columns instead?

I'm going to leave those columns for now. They have a lot of downstream dependencies in redash reports and intermediate tables but as we switch things over we can come back and remove them i think.

rachellougee (approved):
The change looks good. 

Related to my question, I wasn't clear but what I meant is instead of creating two new columns, we could update the source and definition of `user_has_completed_program` and `program_certificate_awarded_on`. So there is no need for upstream changes as everything will flow through.

KatelynGit:
> The change looks good.
> 
> Related to my question, I wasn't clear but what I meant is instead of creating two new columns, we could update the source and definition of `user_has_completed_program` and `program_certificate_awarded_on`. So there is no need for upstream changes as everything will flow through.



> The change looks good.
> 
> Related to my question, I wasn't clear but what I meant is instead of creating two new columns, we could update the source and definition of `user_has_completed_program` and `program_certificate_awarded_on`. So there is no need for upstream changes as everything will flow through.

o good point I think I'll change it to that before I merge

PR #1160, opened 2024-06-07
Changes to int__micromasters__dedp_proctored_exam_grades and upstream models

rachellougee:

Part of https://github.com/mitodl/hq/issues/4481 

This PR includes the changes for MicroMasters' ProctoredExamGrade data.
- Adding stg__micromasters__app__postgres__exams_examrun
- Updating stg__micromasters__app__postgres__grades_proctoredexamgrade - field definitions and tests
- Adding `examrun_readable_id` and `examrun_semester` to int__micromasters__dedp_proctored_exam_grades -field definitions and tests

 I will have a separate PR to add the exam grades from MITx Online and update the `marts__micromasters_dedp_exam_grades` to have the completed data.

### How can this be tested?

dbt build --select stg__micromasters__app__postgres__exams_examrun
dbt build --select stg__micromasters__app__postgres__grades_proctoredexamgrade
dbt build --select int__micromasters__dedp_proctored_exam_grades

### DIFF
--- a/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
+++ b/src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
@@ -21,6 +21,9 @@ models:
   - name: examrun_readable_id
     description: str, edX exam run ID formatted as course-v1:{org}+{exam code}+{semester},
       e.g., course-v1:MITxT+14.100PEx+2T2022. May be null for exam runs before 2022.
+  - name: examrun_semester
+    description: str, semester for the exam run, e.g., 2T2022. May be null for exam
+      runs before 2022.
   - name: user_edxorg_username
     description: str, username on edX.org
   - name: user_mitxonline_username
@@ -33,20 +36,23 @@ models:
     - not_null
   - name: user_mitxonline_email
     description: str, current user email on mitxonline
+  - name: proctoredexamgrade_passing_score
+    description: float, passing score set for the proctored exam, range between 0
+      to 100
+    tests:
+    - not_null
   - name: proctoredexamgrade_score
     description: float, user score for the proctored exam, range between 0 to 100
+    tests:
+    - not_null
+  - name: proctoredexamgrade_percentage_grade
+    description: float, user percentage grade for the proctored exam, range between
+      0 to 1
     tests:
     - not_null
   - name: proctoredexamgrade_is_passing
     description: boolean, indicating whether the user has passed the passing score
       set for this proctored exam
-    tests:
-    - not_null
-  - name: proctoredexamgrade_passing_score
-    description: float, passing score for the proctored exam, range between 0 to 100
-  - name: proctoredexamgrade_percentage_grade
-    description: float, user percentage grade for the proctored exam, range between
-      0 to 1
     tests:
     - not_null
   - name: proctoredexamgrade_exam_on
@@ -69,7 +75,7 @@ models:
     - unique
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
-      column_list: ["user_micromasters_email", "examrun_readable_id"]
+      column_list: ["user_micromasters_email", "proctoredexamgrade_exam_on"]
 
 - name: int__micromasters__course_enrollments
   description: MicroMasters enrollments. Enrollments in courses that satisfy the requirements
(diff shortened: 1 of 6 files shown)


src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__grades_proctoredexamgrade.sql
```diff
@@ -10,7 +10,7 @@ with source as (
         id as proctoredexamgrade_id
         , user_id
         , course_id
-        , passing_score as proctoredexamgrade_passing_score
+        , if(passing_score < 1, passing_score * 100, passing_score) as proctoredexamgrade_passing_score
```
KatelynGit, line 13: If the learner got a '1' indicating a possible 100 would it be represented correctly?

KatelynGit, line 13: If the learner got a '1' indicating a possible 100 would it be represented correctly?

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
```diff
@@ -31,28 +36,46 @@ models:
...
   - name: proctoredexamgrade_id
-    description: int, sequential ID representing a DEDP proctor exam grade
+    description: int, sequential ID representing a DEDP proctored exam grade in MicroMasters
+      database
     tests:
     - not_null
     - unique
+  tests:
+  - dbt_expectations.expect_compound_columns_to_be_unique:
+      column_list: ["user_micromasters_email", "proctoredexamgrade_exam_on"]
```
KatelynGit, line 78: Should this check be removed?  the unique column seems to be proctoredexamgrade_id but do we need another check in addition to that one?

src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
```diff
@@ -880,7 +909,55 @@ models:
...
+  - name: examrun_semester
+    description: str, semester for the exam run, e.g., 2T2022. May be null for exam
+      runs before 2022.
+  - name: examrun_description
+    description: str, description for the exam run
+  - name: examrun_is_authorized
+    description: boolean, indicating if the exam run has authorizations from Pearson
+    tests:
+    - not_null
+  - name: examrun_passing_score
```
KatelynGit, line 952: right now the exam grade fields are showing 4 digits after the decimal it seems like we'd never use all 4 right? and we'd want to reformat the field for reporting eventually in actual report to remove extra unnecessary precision.  Should we do that in the tables or wait to reformat and do it in the report layers?

KatelynGit (approved):
Left a few questions/comments.  The builds ran great and it looks great overall!

src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
rachellougee: This field is not used in any upstream, but it is the same field as `proctoredexamgrade_passing_score` in stg__micromasters__app__postgres__grades_proctoredexamgrade. The 4 digits are for any float fields sourced from raw tables. It would be a lot if we had to convert them all. I will handle that in the later PR if we show them in mart.

src/ol_dbt/models/staging/micromasters/stg__micromasters__app__postgres__grades_proctoredexamgrade.sql
KatelynGit: That works thanks for clarifying!

src/ol_dbt/models/intermediate/micromasters/_int_micromasters__models.yml
KatelynGit: I wasn't sure if we needed another uniqueness check here since we already had one(on proctoredexamgrade_id the pk) but we can leave both.

src/ol_dbt/models/staging/micromasters/_stg_micromasters__models.yml
KatelynGit: handling it later or downstream works for me

PR #1168, opened 2024-06-14
Update combined orders mart and upstream changes

rachellougee:

Part of https://github.com/mitodl/hq/issues/4532

A few changes to marts__combined__orders

- Added the MicroMasters orders for legacy edx.org courses
- Added order_reference_number and courserun_readable_id
- Removed some of the joins to intermediate to simplify the mart . The query for xPro orders should probably be moved to mitxpro__ecommerce_allorders but that is out of scope for this PR.

### How can this be tested?

dbt build --select +int__bootcamps__ecommerce_order
dbt build --select int__micromasters__orders

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__orders.sql
@@ -1,5 +1,4 @@
 --- This model combines intermediate orders from different platforms
----{{ config(materialized='view') }}
 
 with bootcamps__ecommerce_order as (
     select * from {{ ref('int__bootcamps__ecommerce_order') }}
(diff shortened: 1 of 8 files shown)


src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -33,36 +24,61 @@ with bootcamps__ecommerce_order as (
...
         , mitxpro__ecommerce_allorders.req_reference_number
         , mitxpro__ecommerce_order.order_tax_country_code
         , mitxpro__ecommerce_order.order_tax_rate
         , mitxpro__ecommerce_order.order_tax_rate_name
         , mitxpro__ecommerce_order.order_tax_amount
         , mitxpro__ecommerce_order.order_total_price_paid_plus_tax
+        , coalesce(
+            mitxpro__ecommerce_order.order_total_price_paid, mitxpro__b2becommerce_b2border.b2border_total_price
+        ) as order_total_price_paid
         , coalesce(mitxpro__ecommerce_allorders.coupon_id, mitxpro__ecommerce_allorders.b2bcoupon_id) as coupon_id
```
rachellougee, line 60: The `coupon_id` is a bit ambiguous here as I noticed that there are b2b orders, e.g., b2border_id = 201, which had an associated `coupon_id`= 14559 with b2b_only_indicator ='Y'. So I am not sure how the end user knows this refers to a regular coupon vs a b2b coupon. I didn't change as part of this PR but could if needed. Or I can create new issue.

KatelynGit: I think we should make it two separate fields probably. I can do that in my next PR.

KatelynGit (approved):
The builds ran clean and the definitions looked good to me. The code also looks good.

PR #1172, opened 2024-06-17
Extract openedx course metadata

quazi-h:
https://github.com/mitodl/hq/issues/4067

Extends the openedx lib in our Dagster pipelines to process the course bundle metadata. The new code will find the course metadata xml file using the run tag parsed from courses.xml From there, the handful of metadata attributes will be parsed from the file and written to a json file, similar to how the video xml data is being handled in the same pipeline.

### How can this be tested?
Once deployed to QA, should test the new pipeline by kicking it off in Dagster.

### Checklist:
- [ ] Confirm outfile generation logic/location
- [ ] Confirm whether or not to include the chapter or wiki elements.
- [ ] Get new code deployed to QA and Test

### DIFF
--- a/src/ol_orchestrate/assets/openedx_course_archives.py
+++ b/src/ol_orchestrate/assets/openedx_course_archives.py
@@ -1,0 +1,54 @@
+import hashlib
+import json
+from pathlib import Path
+
+from dagster import (
+    AssetExecutionContext,
+    AssetIn,
+    AssetKey,
+    AutoMaterializePolicy,
+    DataVersion,
+    Output,
+    asset,
+)
+
+from ol_orchestrate.assets.edxorg_archive import course_and_source_partitions
+from ol_orchestrate.lib.openedx import process_course_xml
+
+
+@asset(
+    key=AssetKey(("edxorg", "raw_data", "course_xml")),
+    partitions_def=course_and_source_partitions,
+    group_name="edxorg",
+)
+def dummy_edxorg_course_xml(): ...
+
+
+@asset(
+    key=AssetKey(("edxorg", "raw_data", "course_metadata")),
+    partitions_def=course_and_source_partitions,
+    group_name="edxorg",
+    ins={"course_archive": AssetIn(key=AssetKey(("edxorg", "raw_data", "course_xml")))},
+    auto_materialize_policy=AutoMaterializePolicy.eager(
+        max_materializations_per_minute=None
+    ),
+)
+def extract_edxorg_courserun_metadata(
+    context: AssetExecutionContext, course_archive: Path
+):
+    course_metadata = process_course_xml(course_archive)
+    course_metadata_file = Path("course_metadata.json")
+    course_metadata_file.write_text(json.dumps(course_metadata))
+    data_version = hashlib.file_digest(
+        course_metadata_file.open("rb"), "sha256"
+    ).hexdigest()
+    course_metadata_object_key = f"edxorg/processed_data/course_metadata/{context.partition_key}/{data_version}.json"  # noqa: E501
+    yield Output(
+        (course_metadata_file, course_metadata_object_key),
+        output_name="flattened_course_structure",
+        data_version=DataVersion(data_version),
+        metadata={
+            "course_id": course_metadata["course_id"],
+            "object_key": course_metadata_object_key,
+        },
+    )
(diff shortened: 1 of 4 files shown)


rachellougee (commented):
Ran `process_course_xml` locally, other than a couple of errors on the variables commented in PR, the functionality looked good.

src/ol_orchestrate/lib/openedx.py
```diff
@@ -134,3 +143,116 @@ def write_video_json_file(video_data: dict[str, Any], video_data_file: Path):
...
+    enrollment_end = metadata_root.attrib.get("enrollment_end", None)
+    enrollment_start = metadata_root.attrib.get("enrollment_start", None)
+    giturl = metadata_root.attrib.get("giturl", None)
+    graceperiod = metadata_root.attrib.get("graceperiod", None)
+    instructor_info = metadata_root.attrib.get("instructor_info", None)
+    language = metadata_root.attrib.get("language", None)
+    learning_info = metadata_root.attrib.get("learning_info", None)
+    minimum_grade_credit = metadata_root.attrib.get("minimum_grade_credit", None)
+    mobile_available = metadata_root.attrib.get("mobile_available", None)
+    self_paced = metadata_root.attrib.get("self_paced", None)
```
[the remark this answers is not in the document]
rachellougee: The output of the json had these as literal 'true' or 'false' `'self_paced': 'true', 'enable_subsection_gating': 'true'`, rather than json booleans

src/ol_orchestrate/lib/openedx.py
```diff
@@ -134,3 +143,104 @@ def write_video_json_file(video_data: dict[str, Any], video_data_file: Path):
...
+    :rtype: Dict[str, Any]
+    """
+    with tarfile.open(archive_path, "r") as tf:
+        tar_info = tf.getmember("course/course.xml")
+        course_xml_file = tf.extract(tar_info)
+        # course xml file is in the root directory
+        course_id, run_tag = parse_course_id(course_xml_file)
+        # use the run_tag to get the course metadata file
+        tar_info = tf.getmember(f"course/course/{run_tag}.xml")
+        course_metadata_file = tf.extract(tar_info)
```
blarghmatey, line 167: So, looking at the documentation again, and the pre-commit errors, the return value of the extract method is `None`. We'll need to change this and the previous `extract` calls to look like:
```suggestion
        course_metadata_file = Path('course_metadata.xml')
        tf.extract(tar_info, path=course_metadata_file)
```

src/ol_orchestrate/lib/openedx.py
```diff
@@ -209,16 +214,18 @@ def parse_course_xml(metadata_file: str) -> dict[str, Any]:
...
-        slug = wiki.attrib.get("slug", None)
+    slug = wiki.attrib.get("slug", None) if wiki else None
     chapters = metadata_root.findall("chapter", None)
     # if there is chapter data
-    if len(chapters) > 0:
-        chapter_ids = [chapter.attrib.get("url_name", None) for chapter in chapters]
+    chapter_ids = (
+        [chapter.attrib.get("url_name", None) for chapter in chapters]
+        if len(chapters) > 0
+        else None
```
blarghmatey, line 227: This should return an empty list of there are no chapters for consistency of data types

src/ol_orchestrate/assets/openedx_course_archives.py
```diff
@@ -0,0 +1,54 @@
...
+@asset(
+    key=AssetKey(("edxorg", "raw_data", "course_xml")),
+    partitions_def=course_and_source_partitions,
+    group_name="edxorg",
+)
+def dummy_edxorg_course_xml(): ...
+
+
+@asset(
+    key=AssetKey(("edxorg", "raw_data", "course_metadata")),
```
blarghmatey, line 28:
```suggestion
    key=AssetKey(("edxorg", "processed_data", "course_metadata")),
```
I think this makes sense to be under the `processed_data` hierarchy since it's not strictly raw information

blarghmatey: That also matches with the key path that is being constructed in the function body

blarghmatey:
Once the above comments are addressed this should be good to merge

blarghmatey (approved):

PR #1187, opened 2024-07-01
Update metadata fields parsed to match int__edxorg__mitx_courseruns

quazi-h:
https://github.com/mitodl/hq/issues/4067

After setting up the new metadata stream _raw__edxorg__s3__course_structure__course_metadata_ in the **edx.org Production Course Structure** Airbyte connection, we were seeing Trino errors when trying to query the table in Starburst>
`Glue table 'ol_warehouse_production_raw.raw__edxorg__s3__course_structure__course_metadata' column 'allow_anonymous' has invalid data type: null`. Rachel mentioned that most of these elements/attributes are not necessary for us right now, so we should just drop it if it's causing issues. I've removed all of the extraneous fields and whittled the selection down to the courserun fields that are being processed in the [int__edxorg__mitx_courseruns dbt model](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_courseruns.sql#L24-L35) as recommended by Rachel (with the exception of the courserun_url).

### How can this be tested?
Once deployed, allow the metadata files to be reprocesses, and then try querying the tables in Starburst again. Confirm that there are no longer any issues/errors that pop up.

### DIFF
--- a/src/ol_orchestrate/lib/openedx.py
+++ b/src/ol_orchestrate/lib/openedx.py
@@ -214,8 +214,8 @@ def parse_course_xml(metadata_file: str) -> dict[str, Any]:
     return {
         "enrollment_start": enrollment_start,
         "enrollment_end": enrollment_end,
-        "start": start,
-        "end": end,
+        "course_start": start,
+        "course_end": end,
         "instructor_info": instructor_info,
         "self_paced": self_paced,
         "title": title,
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
👍

PR #1189, opened 2024-07-02
removing a join that was causing issues in the int__mitxpro__ecommerce_allorders table

KatelynGit:

Removes a join to commerce coupon that was causing issues with duplicating b2b data.  That table wasn't needed for regular orders anyway and coupon data for those orders can be gotten via the other tables. B2B data should be pulled from b2b tables. This will help with duplicated data downstream as well.

### How can this be tested?
dbt build --select int__mitxpro__ecommerce_allorders

### DIFF
--- a/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
+++ b/src/ol_dbt/models/intermediate/mitxpro/int__mitxpro__ecommerce_allorders.sql
@@ -18,11 +18,6 @@ with b2becommerce_b2border as (
     from {{ ref('int__mitxpro__ecommerce_couponpaymentversion') }}
 )
 
-, ecommerce_coupon as (
-    select *
-    from {{ ref('int__mitxpro__ecommerce_coupon') }}
-)
-
 , ecommerce_couponredemption as (
     select *
     from {{ ref('int__mitxpro__ecommerce_couponredemption') }}
@@ -123,7 +118,7 @@ with b2becommerce_b2border as (
         , course_runs.courserun_id
         , course_runs.courserun_readable_id
         , programs.program_readable_id
-        , ecommerce_coupon.coupon_id
+        , null as coupon_id 
         , b2becommerce_b2border.b2bcoupon_id
         , b2becommerce_b2border.b2border_created_on as order_created_on
         , b2becommerce_b2border.b2border_status as order_state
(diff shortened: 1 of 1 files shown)


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -82,6 +82,9 @@ models:
 - name: marts__combined__orders
   description: B2B and regular orders combined into one table
   columns:
+  - name: combined_orders_id
+    description: int, primary key for this table. At the grain of order, platform,
+      and line.
```
rachellougee, line 87:
```suggestion
      tests:
      - not_null
```
Also, can we update the description to match the order of the long cast in the SQL file and give an example of the generated ID?

src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -234,10 +231,23 @@ with bootcamps__ecommerce_order as (
...
+    cast(
+        cast(coalesce(order_id, 9) as varchar)
+        || cast(coalesce(line_id, 9) as varchar)
+        || cast(coalesce(b2border_id, 9) as varchar)
+        ||
+        cast(case
+            when platform = 'xPro' then 1
+            when platform = 'edX.org' then 2
+            when platform = 'MITx Online' then 3
+            when platform = 'Bootcamps' then 4
```
rachellougee, line 243:
```suggestion
            when platform = '{{ var("mitxpro") }}' then 1
            when platform = '{{ var("edxorg") }}' then 2
            when platform =  '{{ var("mitxonline") }}' then 3
            when platform = '{{ var("bootcamps") }}' then 4
```

src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -234,10 +231,23 @@ with bootcamps__ecommerce_order as (
 )
 
 select
-    platform
+    cast(
+        cast(coalesce(order_id, 9) as varchar)
```
rachellougee, line 235: Curious about why we choose 9 as the default?

rachellougee (approved):
I left a few comments. dbt ran fine. It may be worth creating a macro for the combined_orders_id so it can be re-used, but it can be done later when we have to generate the same ID in other marts later.

src/ol_dbt/models/marts/combined/marts__combined__orders.sql
rachellougee: Also, should we give some thoughts about the order of each platform? Not sure if 1-4 is chosen randomly, or it doesn't matter.

rachellougee: Ok. I didn't realize it but we now have two ways to generate the primary key in our code base. We had `generate_hash_id` used for hightouch.

PR #1197, opened 2024-07-08
fix dbt error on mart combined users

rachellougee:

NA

This fixes the dbt warnings and errors in today's run https://pipelines.odl.mit.edu/runs/3a4d8047-b1c8-4523-90b0-16e7b6b896f0

```
Failure in test not_null_marts__combined__users_platforms (models/marts/combined/_marts__combined__models.yml)
4:45:41.689 AM
run_dbt_2db87
ERROR
  Got 360 results, configured to fail if >10

```

There are some micromasters users added to int__mitx__users but they don't have any social auth account to mitxonline or edxorg so the platform is null for these users. I think we should filter them out in the mart until we know what to do with these users. But let me know if you disagree.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -76,8 +76,8 @@ models:
       platforms
   tests:
   - dbt_expectations.expect_compound_columns_to_be_unique:
-      column_list: ["user_mitxonline_id", "user_edxorg_id", "user_mitxpro_id", "user_bootcamps_id",
-        "platforms"]
+      column_list: ["user_mitxonline_username", "user_edxorg_username", "user_mitxpro_username",
+        "user_bootcamps_username", "platforms"]
 
 - name: marts__combined__orders
   description: B2B and regular orders combined into one table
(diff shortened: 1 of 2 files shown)


pdpinch:
I don't understand -- did you say we have an email for them, but no socialauth for edx.org or for mitxonline?

rachellougee:
> So It looks like a lot of these folks do have a micromasters email address even though they don't have a mitxonline or edxorg one.

Yes, these forks do have MicroMasters email. But MicroMasters is not a platform and currently their profile data is either pulled from MITx Online or edxorg in the combined users mart, see https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/combined/marts__combined__users.sql#L43-L70.

> I don't understand -- did you say we have an email for them, but no socialauth for edx.org or for mitxonline?

As mentioned above, these users do have MicroMasters email but they either don't have any socialauth at all or we can't link them to MITx Online or edx.org based on their socialauth, so their profile fields are blank, which is why we got these warning and error because I recently added these MicroMasters users to `int__mitx__users` https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql#L164-L198 which then flowed into this mart. 

If we do want to include these users in the mart, then we need to fix these profile fields to pull from MicroMasters. But I am unsure what platform these users belongs to for those who have no socialauth at all

rachellougee:
I updated only to filter out roughly 360 micromasters users who don't have any social auth accounts. That's much less than what was before.

PR #1199, opened 2024-07-11
remove b2b and fix pk of orders mart

KatelynGit:

This changes the code so that the primary key of the marts__combined__orders table is generated via a reusable macro that can easily be added to other tables for joining. It also removed the b2b records and columns which will be put in a seperate table since they are not very common and only applicable to one platform.

### How can this be tested?


rachellougee (approved):
Code and test all good

pdpinch:
I wish we had discussed removing the b2b orders before you worked on this. I'm not sure I agree with your reasoning about removing it. 

But now that it's gone, I guess we can leave it out until we can discuss it with the accounts receivable finance team.

PR #1215, opened 2024-07-24
Split the Airbyte syncs into independent jobs

blarghmatey:
N/A

Rather than having the entire Airbyte and dbt graph materialized as a single job, this splits the different Airbyte connections into separate jobs that are independently scheduled, and only materialize the direct descendents of those table assets. This decouples slow and fast jobs, as well as being more resilient to individual connection failures. The downstream dbt models will automatically materialize based on the eager automaterialize policy.

image

### How can this be tested?

Execute the definition locally using `dagster dev -f src/ol_orchestrate/definitions/lakehouse/elt.py` and explore the job and schedule definitions.

### Additional Context

The daily Airbyte sync has grown fragile, and is susceptible to individual connections (or even single table) failures preventing all other downstream assets from being materialized. It also ties the entire execution graph to the slowest connection, preventing any more frequent synchronization for faster connections. This removes that tight coupling, allowing for much more flexible scheduling.

### DIFF
--- a/src/ol_orchestrate/definitions/lakehouse/elt.py
+++ b/src/ol_orchestrate/definitions/lakehouse/elt.py
@@ -1,5 +1,6 @@
 import json  # noqa: INP001
 import os
+import re
 from pathlib import Path
 
 from dagster import (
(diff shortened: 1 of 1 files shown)


quazi-h (approved):
Changes and new logic here looks good to me.
I am currently troubleshooting my local dagster instance as I am hitting errors when I attempt to load the elt.py pipeline, but they are unrelated to these changes - I think I just need to set my dagster env variables.

src/ol_orchestrate/definitions/lakehouse/elt.py
```diff
@@ -65,27 +71,43 @@
...
+    group_names.add(asset.group_name)
+
+airbyte_asset_jobs = []
+airbyte_update_schedules = []
+group_count = len(group_names)
+for count, group_name in enumerate(group_names, start=1):
+    job = define_asset_job(
+        name=f"sync_and_stage_{group_name}",
+        selection=AssetSelection.groups(group_name)
+        .downstream(depth=1, include_self=True)
```
rachellougee, line 89: I will take a look at QA once it's there, but I am wondering when the intermediate models are materialized now with this change. For example, we sometimes get warning about [not_null_int__micromasters__users_user_profile_id](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/micromasters/subqueries/__int_micromasters_subqueries__models.yml#L103) because of two staging/raw tables are out of sync. Will this change have any effect on these test?

PR #1233, opened 2024-08-06
Add Dagster Assets to Extract Open edX Course Structures data

quazi-h:
Refactoring the openedx course structures pipeline code to Dagster assets.
Added a course_list asset and a course_structure multi asset that will produce the course_blocks and course_structures files. Utilizing the existing s3 io_manager to handle the asset oriented files and using data versioning to better organize the uploaded files. Still need to update the actual job to call these new components. Leaving the existing openedx ops alone since the edx_course_pipeline graph job still requires them and we do not want to disrupt downstream consumers of that data like Jon.

Addresses https://github.com/mitodl/hq/issues/4465

We are seeing a mix of data from different openedx deployments in the course_structures and course_blocks data in our data lake. As it is, we are pulling a list of course_ids and using that to query the Open edX REST API to retrieve the course_structure and course_blocks data. It appears that both mitxonline and xpro courses are being pulled when this happens as the files in both mitxonline_openedx_extracts and xpro_openedx_extracts in S3 (ol-data-lake-landing-zone-production) are identical.

Without modifying the existing code that handles generating the list of courses and queries the Open edX instances for the course structures data (because we have other downstream consumers of the data produced by the edx_course_pipeline), we want to put up new Dagster assets to handle generating the list of courses and pulling and flattening the course_structures data. This PR adds a new file that contains openedx Dagster assets.

### How can this be tested?
Try running the pipeline locally.
```
export GITHUB_TOKEN=<your_token>
dagster dev -f src/ol_orchestrate/definitions/edx/openedx_data_extract.py
```
Once the code is merged, test on Dagster QA.

### Checklist:
- [ ] Update the extract_open_edx_data_to_ol_data_platform graph in jobs/open_edx.py

### DIFF
--- a/src/ol_orchestrate/assets/openedx.py
+++ b/src/ol_orchestrate/assets/openedx.py
@@ -1,0 +1,166 @@
+# - Query the openedx api to get course structures and course blocks data
+# - Model the different asset objects according to their type
+
+import hashlib
+import json
+from datetime import UTC, datetime
+from pathlib import Path
+
+import jsonlines
+from dagster import (
+    AssetExecutionContext,
+    AssetIn,
+    AssetKey,
+    AssetOut,
+    AutoMaterializePolicy,
+    Config,
+    DagsterEventType,
+    DataVersion,
+    EventRecordsFilter,
+    Output,
+    asset,
+    multi_asset,
+)
+from flatten_dict import flatten
+from flatten_dict.reducers import make_reducer
+from pydantic import Field
+from upath import UPath
+
+from ol_orchestrate.lib.openedx import un_nest_course_structure
+
+
+@asset(
+    key=AssetKey(("openedx", "raw_data", "course_list_config")),
+    # partitions_def=,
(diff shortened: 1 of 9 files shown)


rachellougee (approved):
The code looks good. But the test instruction `dagit -d src/ol_orchestrate -f src/ol_orchestrate/jobs/open_edx.py` doesn't seem to work for me as I got `dagster._core.errors.DagsterInvalidDefinitionError: resource with key 'openedx' required by op 'list_edx_courses' was not provided. Please provide a ResourceDefinition to key 'openedx', or change the required key to one of the following keys which points to an ResourceDefinition: ['io_manager']` 

Are there any variables that need to be set beside `DAGSTER_ENV` in order to run it locally?

blarghmatey:
There shouldn't be any additional variables needed. The only thing I exported was the `GITHUB_TOKEN` that is used for authenticating to Vault.

PR #1244, opened 2024-08-20
Update dbt execution to be compatible with Dagster 1.8

blarghmatey:
N/A

The 1.8 release of Dagster updates the API for working with dbt projects. This updates the lakehouse elt definition to work with that API.

### How can this be tested?

Run the elt definition locally after exporting the Airbyte environment variables
```
export DAGSTER_AIRBYTE_HOST="config-airbyte-qa.odl.mit.edu"
export DAGSTER_AIRBYTE_PORT=443
export DAGSTER_AIRBYTE_AUTH="dagster:<redacted>"
dagster dev -f src/ol_orchestrate/definitions/lakehouse/elt.py
```


rachellougee (commented):
I don't know the password for "config-airbyte-qa.odl.mit.edu", so I had to test on my local host. I got `dagster._core.definitions.events.Failure: Max retries (3) exceeded with url: https://127.0.0.1:3000/api/v1/workspaces/list.` with dagster 1.8.1 locally. Is that expected? Please let me know if I am missing something.

blarghmatey:
That is expected, yes. That's coming from the Airbyte host.

rachellougee (approved):
Ok. Thank you.

src/ol_orchestrate/definitions/lakehouse/elt.py
```diff
@@ -1,50 +1,63 @@
...
-    else Path("/opt/dbt")
-)
-
 dbt_config = {
-    "project_dir": str(dbt_repo_dir),
-    "profiles_dir": str(dbt_repo_dir),
-    "target": dagster_deployment,
+    "project_dir": str(DBT_REPO_DIR),
+    "profiles_dir": str(DBT_REPO_DIR),
+    "target": DAGSTER_ENV,
```
rachellougee, line 58: If this target refers to the target in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/profiles.yml, it connects duckdb for dev. Should we make it configurable when running locally?

rachellougee (approved):
The code looks good to me, but I couldn't test it as I keep getting `dagster._core.errors.DagsterUserCodeUnreachableError: Could not reach user code server. gRPC Error code: UNAVAILABLE` when running `dagster dev -f src/ol_orchestrate/definitions/lakehouse/elt.py` locally.

src/ol_orchestrate/definitions/lakehouse/elt.py
blarghmatey: Good catch. I had done that in the assets module but forgot to reflect it here.

PR #1271, opened 2024-09-05
Openedx course xml export

blarghmatey:
N/A

Implements an asset-oriented method for course XML export from Open edX instances so that it can be used for downstream processing.

### How can this be tested?

Run the code location:
```
export GITHUB_TOKEN=<token with read:org scope>
dagster dev -f src/ol_orchestrate/definitions/edx/openedx_data_extract.py
```
Then go to the local running Dagster instance, visit the "Deployments" tab and turn on the sensors for one of the target deployments. Once it generates the list of partitions, manually trigger a materialization for one of the partitions in the `openedx/courseware` asset. That will then kick off a materialization of the downstream assets, including the course XML

### Additional Context

Our current means of exporting course XML is on a naive daily trigger. This updates the logic to only export the course data when there is a change in the running course. This also simplifies the work of consuming and processing the course XML data to derive other details, such as video duration data.

### DIFF
--- a/src/ol_orchestrate/assets/openedx.py
+++ b/src/ol_orchestrate/assets/openedx.py
@@ -6,6 +6,7 @@ import json
 import time
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from urllib.parse import urlparse
 
 import jsonlines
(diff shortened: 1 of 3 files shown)


src/ol_orchestrate/assets/openedx.py
```diff
@@ -120,3 +129,120 @@ def course_structure(context: AssetExecutionContext, courseware):  # noqa: ARG00
...
+        time.sleep(timedelta(seconds=5).seconds)
+        for course_id, task_id in tasks.items():
+            task_status = context.resources.openedx.client.check_course_export_status(
+                course_id,
+                task_id,
+            )
+            if task_status["state"] == "Succeeded":
+                successful_exports.add(course_id)
+            if task_status["state"] in {"Failed", "Canceled", "Retrying"}:
+                failed_exports.add(course_id)
```
rachellougee, line 164: It looks like this code is used in edx_export_courses ops. Is it worth writing a lib function for this course export task?

rachellougee (approved):
Not able to test on my local dagster instance, but I ran some manual tests on extract_courserun_details. That functionality works fine. I left one small comment, but up to you.

src/ol_orchestrate/assets/openedx.py
blarghmatey: That's now a legacy implementation that will be deprecated, and it's just different enough that it's not worth extracting to a lib.

PR #1273, opened 2024-09-06
add courserun_upgrade_deadline to product mart and intermediate combined runs

rachellougee:

NA
Adding courserun_upgrade_deadline to https://bi.ol.mit.edu/superset/dashboard/25/

Adding courserun_upgrade_deadline to marts__combined__products and int__combined__course_runs

### How can this be tested?

dbt build --select int__combined__course_runs

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/_combined_models.yml
+++ b/src/ol_dbt/models/intermediate/combined/_combined_models.yml
@@ -245,6 +245,10 @@ models:
     description: timestamp, specifying when the course run begins. May be Null.
   - name: courserun_end_on
     description: timestamp, specifying when the course run ends. May be Null.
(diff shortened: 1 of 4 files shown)


rachellougee:
@KatelynGit Thanks for the review. 

As for [this warning](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitxpro/_int_mitxpro__models.yml#L967) during your test, that's related to how you ran dbt command. If you ran `dbt build --select +marts__combined__products` it would cause this warning because it only refreshed `int__mitxpro__ecommerce_product` but not `int__mitxpro__ecommerce_productversion` because the mart doesn't reference that intermediate. This happens a lot due to how we run dbt code locally. I wanted to mention it in case you wondering why.

PR #1287, opened 2024-09-12
Update combined user roles to associate seed file users based on hashed email

rachellougee:

https://github.com/mitodl/hq/issues/5383

Updating int__combined__user_course_roles to associate these 3 seed file users listed in user_course_roles.csv with their emails in mitxonline since they don't have any course roles in edx.org

### How can this be tested?

Run these commands to refresh the upstream models.

dbt run --select +int__combined__user_course_roles
dbt build --select int__combined__user_course_roles

Then run this you should see there are 453 rows with the user_email populated
```
select * from "ol_data_lake_production"."ol_warehouse_production_rlougee_intermediate"."int__combined__user_course_roles"
where hashed_user_email in ('8531233373138830eeacb853250d724b15fe5f9aac844641b79331c7fa86fe7e',
'bcdacb4d576676bcbf9a2ff16f6c0c1276cb8eae4f9115aaa339b5bf92a8bdfa'
,'4ff9dc0205e3f54985836ef5d3ec2a1f5611a0f0bec06851c00638225ff74965')
```

The row level filter used in https://bi.ol.mit.edu/rowlevelsecurity/list/?pageIndex=0&sortColumn=changed_on_delta_humanized&sortOrder=desc will work for their emails so they can see all DEDP courses.

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/int__combined__user_course_roles.sql
+++ b/src/ol_dbt/models/intermediate/combined/int__combined__user_course_roles.sql
@@ -115,55 +115,54 @@ with mitxonline_courseroles as (
     inner join residential_openedx_users on residential_courseroles.user_id = residential_openedx_users.user_id
 )
 
-, combined_users as (
+, seed_file_users as (
     select * from (
         select
-            platform
-            , user_username
-            , user_email
-            , hashed_user_email
-            , user_full_name
-            , row_number() over (partition by hashed_user_email, platform order by user_email desc) as row_num
+            combined_courseroles.platform
+            , combined_courseroles.user_username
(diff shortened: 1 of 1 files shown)


quazi-h (approved):
Changes look good here. The first time I ran the commands, I only got 303 records when I queried the table in Starburst. I reran the commands with the --full-refresh flag and was able to see 453 records in the table.

rachellougee:
Thanks, Quazi. It shouldn't need a --full-refresh if you run `dbt run --select +int__combined__user_course_roles` first as there are no schema changes to any of these tables. Good that you can see all 453 records.

Issue #1267, opened 2024-09-03
Create a docker-compose dev setup


blarghmatey:
### Description/Context

The local development setup for Dagster can be confusing for newcomers. To simplify the setup and standardize the experience we want to use a docker-compose developer environment that is similar to how we build our other applications.

### Plan/Design

- Create a docker-compose file that mounts the locally cloned repository at the proper locations
- Ensure that the various Dagster CLI commands are available in the path when loading a shell in the container
- Document how to run specific code locations

PR #1294, opened 2024-09-16
Create a docker-compose dev setup

quazi-h:
https://github.com/mitodl/ol-data-platform/issues/1267

Added a new dockerfile, docker-compose file, and new dagster yaml fil…e for local dagster development.

### How can this be tested?
You will need to make sure you have the latest version of Docker Compose.
Start up the local application by running `docker-compose up`.

### Checklist:
- [ ] Finish up with logic for sensor status helper method
- [ ] Work through a few more linter errors and get app running locally

### DIFF
--- a/dagster.yaml
+++ b/dagster.yaml
@@ -1,0 +1,38 @@
+---
+storage:
+  postgres:
+    postgres_db:
+      username:
+        env: DAGSTER_PG_USERNAME
+      password:
+        env: DAGSTER_PG_PASSWORD
+      hostname:
+        env: DAGSTER_PG_HOST
+      db_name:
+        env: DAGSTER_PG_DB
+      port: 5432
+
+compute_logs:
+  module: dagster_aws.s3.compute_log_manager
+  class: S3ComputeLogManager
+  config:
+    bucket: "mycorp-dagster-compute-logs"
+    prefix: "dagster-test-"
+
+local_artifact_storage:
+  module: dagster.core.storage.root
+  class: LocalArtifactStorage
+  config:
+    base_dir: "/opt/dagster/local/"
+
+run_launcher:
+  module: dagster_docker
+  class: DockerRunLauncher
+  config:
+    env_vars:
+    - DAGSTER_PG_USERNAME
+    - DAGSTER_PG_PASSWORD
+    - DAGSTER_PG_DB
+    container_kwargs:
+      volumes:
+      - ~/GIT/ol-data-platform:/opt/dagster/app/
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,36 @@
+# -*- mode: yaml -*-
+
+services:
+  # This service runs dagster-webserver, which loads your user code from the user code container.
+  # Since our instance uses the QueuedRunCoordinator, any runs submitted from the webserver will be put on
+  # a queue and later dequeued and launched by dagster-daemon.
+  dagster-webserver:
+    container_name: dagster_webserver
+    image: dagster_webserver_image
+    restart: unless-stopped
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN pip install poetry
+
+ENV DAGSTER_HOME=/opt/dagster/dagster_home/
+# Copy dagster instance YAML to $DAGSTER_HOME
+COPY dagster.yaml /opt/dagster/dagster_home/
+
+USER dagster
+EXPOSE 3000
+WORKDIR /opt/dagster/code
+RUN poetry install --without=dev --no-cache
```
rachellougee, line 38: This line gave me the error `3.248 Poetry could not find a pyproject.toml file in /opt/dagster/code or its parents`. We are missing a COPY pyproject.toml  to the current WORKDIR.

rachellougee (commented):
I left two comments below related to Dockerfile. 

I am also getting dagster config errors from `dagster_webserver` and `dagster_daemon` because of missing these 4 env.  
https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/dagster.yaml#L20-L26

```
dagster._config.errors.PostProcessingError: You have attempted to fetch the environment variable "DAGSTER_PG_USERNAME" which is not set. In order for this execution to succeed it must be set in this environment.
```

Can we set these environments locally? What values should we set?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    apt install -y mongodb-org-tools && \
+    apt clean && \
+    apt autoremove && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip install poetry
+
+ENV DAGSTER_HOME=/opt/dagster/dagster_home/
+# Copy dagster instance YAML to $DAGSTER_HOME
+COPY dagster.yaml /opt/dagster/dagster_home/
```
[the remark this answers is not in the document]
quazi-h: So I had initially created a separate dagster.yaml with a lot of the config from src/ol_orchestrate/dagster.yaml removed to avoid the errors with the missing environment variables. After pairing with Tobias he mentioned that we could probably do without a yaml for the docker compose file because Dagster should auto generate a yaml file when the docker compose runs.

rachellougee (approved):
`docker compose build` works now. `docker compose run` gave me import errors `ModuleNotFoundError: No module named 'ol_orchestrate'` but UI is up http://localhost:3000/locations.

Approving it but it would be nice to know what I am missing.

```
dagster._core.errors.DagsterImportError: Encountered ImportError: `No module named 'ol_orchestrate'` while importing module ol_orchestrate.definitions.edx.normalize_tracking_logs
```

image

README.md
```diff
@@ -3,3 +3,18 @@
...
 building data applications.
+
+# Running Dagster Locally via Docker
+- Ensure that you have the latest version of Docker installed.
+	https://www.docker.com/products/docker-desktop/
+- Install docker compose. Check the documentation and requirements for your specific machine.
+	https://docs.docker.com/compose/install/
+- Ensure you are able to authenticate into GitHub + Vault
+	https://github.com/mitodl/ol-data-platform/tree/main
+	https://vault-qa.odl.mit.edu/v1/auth/github/login
```
Ardiea, line 14: Why do I need to be able to auth to vault?

blarghmatey: That's because the Dagster pipeline code is using HVAC to connect to Vault for retrieving runtime secrets associated with QA environments

Ardiea: All the talking to vault is done via the github token I presume?

blarghmatey: Correct. https://github.com/mitodl/ol-data-platform/blob/mitlearn_data_flow_rfc/src/ol_orchestrate/resources/secrets/vault.py#L32-L38

PR #1304, opened 2024-09-23
[WIP] Add RFC for MIT Learn data flows

blarghmatey:
N/A

Adds an RFC for migrating data flows from Learn to the data platform

### DIFF
--- a/docs/rfcs/0000-rfc-template.md
+++ b/docs/rfcs/0000-rfc-template.md
@@ -1,0 +1,90 @@
+# [RFC Template] Title
+
+**Authors:**
+
+-
+
+## 1 Executive Summary
+
+<!-- A one or two sentence summary of the purpose for this RFC -->
+
+## 2 Motivation
+
+*What motivates this proposal, and why is it important?*
+
+*Here, we aim to get comfortable articulating the value of our actions.*
+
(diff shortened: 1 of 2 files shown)


mbertrand:
A few things to keep in mind:

 - While most "learning resources" are read-only as far as the end user is concerned, and created/updated/deleted only by the ETL pipelines, there is one exception: learning paths.  A learning path has child learning resources determined by the user (but only admins and a select group of users should have permission to create/update them).  The other resource types are: course, program, video, video_playlist, podcast, podcast_episode.

- Admin users have the ability to edit any learning resource (or news/event object) directly in the django admin interface.  Not typically used except locally, but we should probably make this read-only if we go forward with the data platform approach.
 
 - MIT Learn allows users to subscribe to certain queries (like courses provided by OCW, for example), ie the `PercolateQuery` model in django (`postgres.public.learning_resources_search_percolatequery` tables in the db).  Whenever a learning resource that would be returned by this query is upserted, an email is sent to subscribing users.  Currently this is implemented as a pluggy function that is triggered whenever a resource is upserted.

blarghmatey:
Closing in favor of https://github.com/mitodl/hq/pull/5687

PR #1307, opened 2024-09-26
add mart combined orders table to finance permissions

KatelynGit:
https://github.com/mitodl/hq/issues/5548 bryan requested mart orders in addition to demographics as well for Phoebe and finance

adds combined orders mart to finance permissions per the request of Bryan yesterday

### How can this be tested?
dbt build --select marts__combined__orders and review yaml

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -112,6 +112,9 @@ models:
 
 - name: marts__combined__orders
   description: ecommerce regular b2c orders
+  config:
+    grants:
+      select: ['finance']
   columns:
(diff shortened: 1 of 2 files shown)


rachellougee:
I have a question related to this. It looks like finance has access to `marts__combined_course_enrollment_detail` https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/combined/_marts__combined__models.yml#L259, Is there a reason we need to grant marts__mitxonline_course_enrollments specifically? The combined table should cover the mitxonline enrollments already

rachellougee (approved):

rachellougee (approved):
Looks good. @KatelynGit Thanks for checking in with Bryan

PR #1316, opened 2024-10-01
add is_live and free courses to combined product mart

rachellougee:

https://github.com/mitodl/hq/issues/3992

Adding `is_live` and 'free' courses to marts__combined__products so that it can be used as part of the `Open for enrollment` filter in https://bi.ol.mit.edu/superset/dashboard/25/
   - The calculated field `Open for enrollment` should return 72 results, which should match with https://mitxonline.mit.edu/api/v2/courses/?courserun_is_enrollable=true&live=true&page__live=true
   - `is_live` = true is same as `live=true&page__live=true` filter used on MITx Online api.
   - free course e.g. https://mitxonline.mit.edu/courses/course-v1:MITxT+8.EFTx/ is included now

### How can this be tested?

Then verify this query returns 72 rows
select * from ol_data_lake_production.ol_warehouse_production_YOURNAME_mart.marts__combined__products
where platform='MITx Online'
and from_iso8601_timestamp(enrollment_start_on) <= current_date
and (enrollment_end_on is null or from_iso8601_timestamp(enrollment_end_on) > current_date)
and **is_live = true**

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -511,9 +511,9 @@ models:
     tests:
     - not_null
   - name: is_live
-    description: boolean, indicating whether the course/program is set as live on
-      MITx Online or xPro. Courses/Programs set as is_live=false are not listed on
-      the catalog page.
+    description: boolean, indicating whether the course/program page is set as live
+      on MITx Online or xPro. Courses/Programs set as is_live=false are not listed
+      on the catalog page.
     tests:
     - not_null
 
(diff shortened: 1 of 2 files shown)


rachellougee:
> The build ran clean and the test query results looked good so I'm approving. My main concern with this change is if we change the field to allow non-numeric values like "Free" it's more difficult to do calculations on the list price field. Taking a sum, average, etc. Should this change be made? or maybe should "Free" be 0 instead?

That's a fair point on the price type. Part of this PR is to add the remaining MITx Online courses that have no normalized product price. For these courses, the only place we can get price detail is from CMS but the field is free text. These added courses are not all free courses, e.g. course-v1:MITxT+15.699x+3T2022. In CMS it says `$49` as free text but no normalized price as product in `3T2022` run.  Let me ask about the course.

rachellougee:
@KatelynGit, I updated the list price to a number, which is the normalized price. I left these courses with no normalized product as null unless there is a need to specify 0 for free courses. Can you take another look? thanks

PR #1317, opened 2024-10-03
Update discount field in staging discount models

rachellougee:

Item 2 from https://github.com/mitodl/hq/issues/5705 needs dbt code change as discount fields are formatted in staging.

Removing the text “off” from the ‘discount’ field and keeping $” or “%” signs

### How can this be tested?

dbt run --select +marts__combined__orders

then verify the field
select discount from ol_data_lake_production.ol_warehouse_production_mart.marts__combined__orders
where discount is not null

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -149,8 +149,8 @@ models:
     description: string, unique string to identify a course run on the corresponding
       platform.
   - name: discount
-    description: str, discount applied for the order in readable format, e.g., $100
-      off, 30% off.
+    description: str, discount applied for the order in readable format, e.g., $100,
+      30%.
   - name: order_created_on
     description: timestamp, specifying when the order was initially created. If this
       is a b2b order it will be the timestamp that order was created otherwise it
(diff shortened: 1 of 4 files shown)


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -149,8 +149,8 @@ models:
     description: string, unique string to identify a course run on the corresponding
       platform.
   - name: discount
-    description: str, discount applied for the order in readable format, e.g., $100
-      off, 30% off.
+    description: str, discount applied for the order in readable format, e.g., $100,
+      30%.
```
pdpinch, line 153:
```suggestion
    description: str, discount applied to the order in a readable format. The value is always an amount taken off the list price of the order, e.g., "$100" means $100 was taken off the price. "30%" means the price was reduced by 30%. 
```

I think this is clearer and more grammatical. Do we use "list price" elsewhere, or is there a better phrase for that?

rachellougee: yes, we use `list_price` in marts__mitxpro_ecommerce_productlist. I will update the description. Thanks.

pdpinch:
@KatelynGit did you find out from Bryan why he wants to remove "off" ? Do you think the field's meaning is still reasonably clear, with "discount" instead of "discount off" ?

PR #1318, opened 2024-10-04
add 5 columns to combined users mart in order to move a redash report with these columns to superset

KatelynGit:
https://github.com/mitodl/hq/issues/5610

Currently in Redash we report on the first time a user was enrolled in a course, the last time a user was enrolled in a course, the number of certificates awarded to a user, the total amount a user has paid, the user's income, and  whether they are a program credential holder. In order to continue to report on these values we should add the columns to our marts so that we can surface them more easily in superset.

adds the following columns to the combined users mart:
first_course_start_datetime, last_course_start_datetime, number_of_courserun_certificates, total_amount_paid_orders, latest_income_usd, program_cred_ind

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -110,7 +110,8 @@ models:
   - name: last_course_start_datetime
     description: string, year and term from the users latest course enrollment
   - name: number_of_courserun_certificates
-    description: int, number of course run enrollment certificates associated with the user
+    description: int, number of course run enrollment certificates associated with
+      the user
   - name: total_amount_paid_orders
     description: int, sum of total orders paid associated with the user
   - name: latest_income_usd
--- a/src/ol_dbt/models/marts/combined/marts__combined__users.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__users.sql
@@ -33,31 +33,31 @@ with mitx__users as (
 )
 
 , program_stats as (
-    select 
+    select
         user_email
         , count(distinct programcertificate_uuid) as cert_count
-        , sum(case 
-            when 
+        , sum(case
+            when
                 program_title in (
                     'Data, Economics, and Design of Policy'
                     , 'Data, Economics, and Design of Policy: International Development'
                     , 'Data, Economics, and Design of Policy: Public Policy'
-                ) 
-                and 
+                )
+                and
                 programcertificate_uuid is not null
-                then 1 
+                then 1
             else 0
         end) as dedp_program_cred_count
     from combined_programs
-    group by user_email 
+    group by user_email
 )
 
 , orders_stats as (
-    select 
+    select
         user_email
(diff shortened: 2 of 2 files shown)


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -105,6 +105,19 @@ models:
...
+    description: string, year and term from the users latest course enrollment
+  - name: number_of_courserun_certificates
+    description: int, number of course run enrollment certificates associated with
+      the user
+  - name: total_amount_paid_orders
+    description: int, sum of total orders paid associated with the user
+  - name: latest_income_usd
+    description: int, the users latest income in usd. Only populated for mitxonline.
+  - name: program_cred_ind
+    description: string, Set to Y if the user has recieved a program credential
```
rachellougee, line 120: This column is a bit unambiguous. A learner can receive multiple program certificates, how are we planning to use this column to determine if he/she received DEDP in the report you linked?

Also, can we make this column boolean instead of Y and make the name clearer?

rachellougee (commented):
I will need to review it closely next Monday but I left a question. I am not sure this mart can be used to answer DEDP user questions but I will check.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -105,6 +105,21 @@ models:
...
+    description: int, number of course run enrollment certificates associated with
+      the user
+  - name: total_amount_paid_orders
+    description: int, sum of total orders paid associated with the user
+  - name: latest_income_usd
+    description: int, the users latest income in usd. Only populated for mitxonline.
+  - name: program_cred_ind
+    description: string, Set to true if the user has recieved a program credential
+  - name: dedp_program_cred_ind
+    description: string, Set to true if the user has recieved a dedp program credential
```
rachellougee, line 122:
```suggestion
  - name: has_program_certificate
    description: boolean, Set to true if the user has received a program credential
  - name: has_dedp_program_certificate
    description: boolean, Set to true if the user has received a DEDP program credential
```

The name and data type could be clearer, and also some minor typos.

src/ol_dbt/models/marts/combined/marts__combined__users.sql
```diff
@@ -16,18 +16,68 @@ with mitx__users as (
...
+, program_stats as (
+    select
+        user_email
+        , count(distinct programcertificate_uuid) as cert_count
+        , sum(case
+            when
+                program_title in (
+                    'Data, Economics, and Design of Policy'
+                    , 'Data, Economics, and Design of Policy: International Development'
+                    , 'Data, Economics, and Design of Policy: Public Policy'
```
rachellougee, line 44: I would generally suggest using our existing column `is_dedp_program` so that we don't hardcode this logic in multiple places. But it can be done as improvement in later.

rachellougee (approved):
I left two additional comments. There are debatable arguments about whether these columns belong to this model, but that could be addressed when we re-model our marts. Approving it for now.

PR #1323, opened 2024-10-08
add user_hashed_id to combined marts

rachellougee:

https://github.com/mitodl/hq/issues/5715

Adding `user_hashed_id` and unique test to marts__combined__users and other marts so that it can be used for joining. Without this column, it isn't easy to join marts__combined__users.

### How can this be tested?

dbt build -select 1+marts__combined__users --- the direct upstream model should be enough
dbt build -select marts.combined

You will get a failed [unique test ](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/combined/_marts__combined__models.yml#L435)on marts__combined_program_enrollment_detail, which is unrelated to this PR. https://github.com/mitodl/hq/issues/5756 is created to address that.

### DIFF
--- a/src/ol_dbt/models/intermediate/bootcamps/int__bootcamps__ecommerce_receipt.sql
+++ b/src/ol_dbt/models/intermediate/bootcamps/int__bootcamps__ecommerce_receipt.sql
@@ -9,6 +9,7 @@ select
     , receipt_updated_on
     , order_id
     , receipt_data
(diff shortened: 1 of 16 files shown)


quazi-h (approved):
My dbt builds all completed (finally) and I confirmed the user_hashed_id in all the updated tables.

rachellougee:
The mart test takes a while because we now have so much data. @quazi-h Thank you for taking the time to test.

PR #1357, opened 2024-10-30
add program run readable id to combined orders mart

KatelynGit:
https://github.com/mitodl/hq/issues/2425

But also feedback from the finance team
https://docs.google.com/document/d/12k4cphQVbb9EBRxHZLEYmAqeyfMhwLxL44Lif3JFU9o/edit?tab=t.0

Add the Program Run Description to the combined orders table:         
                                            “program-v1:xPRO+SysEngx+R22”  
                                     **This field is the ‘product_id’ from BI query 1045. 

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
+++ b/src/ol_dbt/models/marts/combined/_marts__combined__models.yml
@@ -214,7 +214,8 @@ models:
         row_condition: "platform != 'Bootcamps' and platform != 'edX.org'"
   - name: product_readable_id
     description: string, the product identifier either referring to courserun_readable_id,
-      programrun_readable_id, or program_readable_id. May be null for bootcamps and edX.org.
+      programrun_readable_id, or program_readable_id. May be null for bootcamps and
+      edX.org.
     tests:
     - dbt_expectations.expect_column_values_to_not_be_null:
         row_condition: "platform != 'Bootcamps' and platform != 'edX.org'"


src/ol_dbt/models/marts/combined/marts__combined__orders.sql
```diff
@@ -168,6 +172,7 @@ with bootcamps__ecommerce_order as (
         , mitxpro__ecommerce_allorders.coupon_id
         , mitxpro__ecommerce_allorders.order_id
         , mitxpro__ecommerce_order.order_total_price_paid
+        , mitxpro__programruns.programrun_readable_id
```
rachellougee, line 175: Can we add this to line 181 https://github.com/mitodl/ol-data-platform/pull/1357/files#diff-3c6e61ec0ee80260ef47e6eb1f62d6cf0eff2cd30a78161734ed38a752c16122R181 instead of a new field? `product_readable_id` is a combined field. It looks like Bryan mentioned it in the google doc, that way we don't need to make any additional changes in the dashboard

> ----Add the Program Run Description to the ‘product_readable_id’ field.
>                        Instead of: “program-v1:xPRO+SysEngx” it should read:            
>                                             “program-v1:xPRO+SysEngx+R22”

rachellougee (approved):
Code runs fine. 

I would also update the [product_readable_id](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/marts/combined/_marts__combined__models.yml#L215) description to reflect the changes you are making

I noticed that not every program purchase is populated with run tag. I guess we can deploy this and see what Bryan says. So approving it.

```

 select * from "ol_data_lake_production"."ol_warehouse_production_kdelaney_mart"."marts__combined__orders"
 where product_readable_id ='program-v1:xPRO+SysEngx'
```

Issue #1353, opened 2024-10-28
Create dim_platform dimensional table


rachellougee:
### Description/Context

As part of https://github.com/mitodl/ol-data-platform/issues/1325, we need a dimensional table for information about the technical system in which a given event is executed within the context.

### Plan/Design

Create `dim_platform` dimensional table to contain the following fields:
-  platform_id 
-  platform_name (e.g. MITx online, xPro, edX.org)
-  platform_description
-  platform_domain (e.g. mitxonline.mit.edu) 

The actual content could be seeded by a CSV file since these fields are mostly static

PR #1364, opened 2024-11-04
Create dim_platform model

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/1353

Creating `dim_platform` seeded by a csv file

### How can this be tested?

dbt build --select +dim_platform

### DIFF
--- a/src/ol_dbt/seeds/platforms.csv
+++ b/src/ol_dbt/seeds/platforms.csv
@@ -1,0 +1,6 @@
+id,platform_name,platform_description,platform_domain
+1,MITx Online,Online courses from MIT faculty and instructors,mitxonline.mit.edu
+2,edX.org,MIT Online Courses on edX.org,www.edx.org
+3,MIT xPRO,Professional development courses and programs that build skills and augment careers,xpro.mit.edu
+4,Residential MITx,Residential MITx,lms.mitx.mit.edu
+5,MIT Bootcamps,MIT Bootcamps, bootcamp.odl.mit.edu
--- a/src/ol_dbt/seeds/platforms.csv
+++ b/src/ol_dbt/seeds/platforms.csv
@@ -1,7 +1,7 @@
 id,platform_name,platform_description,platform_domain
-1,MITx Online,Online courses from MIT faculty and instructors,mitxonline.mit.edu
-2,edX.org,MIT Online courses on edX.org,www.edx.org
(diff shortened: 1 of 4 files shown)


src/ol_dbt/seeds/platforms.csv
```diff
@@ -0,0 +1,6 @@
+id,platform_name,platform_description,platform_domain
+1,MITx Online,Online courses from MIT faculty and instructors,mitxonline.mit.edu
+2,edX.org,MIT Online courses on edX.org,www.edx.org
+3,MIT xPRO,MIT Professional development courses and programs that build skills and augment careers,xpro.mit.edu
+4,MITx Residential,MITx Residential,lms.mitx.mit.edu
+5,MIT Bootcamps,MIT Bootcamps, bootcamp.odl.mit.edu
```
rachellougee, line 6: @pdpinch Do you have any suggestions on the `platform_description`? 

I grabbed some text from https://openlearning.mit.edu/, but it needs some editing
image

rachellougee:
@KatelynGit Thanks for the review.

> Should we include OCW?
Sure. I added OCW although we don't include OCW into the combined mart. It's good to have that in place.

> One thing is should the platform id be a numeric data type? Usually numeric types are better for id's for size and speed reasons. 
I believe we have discussed this in RFC https://github.com/mitodl/hq/blob/main/rfcs/0004-data-warehouse-dimensional-marts.md#schema-considerations.  The hash key makes sense to me as we need a non-numeric primary key for tables like combined users. I thought we were on the same page on this, but let me know if we need to re-visit this discussion.

> And does the platform info in the csv match the info coming form the applications?
Can you say more about this? Not sure if I understand what you asking.

rachellougee:
> I just meant for things like the platform names this new table lists the platform as "MIT xPRO" but in the marts it is displayed differently like "xPro" in marts__combined__orders etc. Do you think we should change it to match what we have in the marts or applications currently?

Thank you for clarifying that.  For clarity, I intended to spell out `MIT xPRO` instead of `xPRO` in the new model to match with https://openlearning.mit.edu/.  Currently, `xPRO` in the mart is pulled from a variable defined in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/dbt_project.yml#L36. 

Unless we hardcoded the platform somewhere on the superset dataset, this new value will flow to the mart without any side effects.  I can match to what we currently have if needed

PR #1375, opened 2024-11-12
Stage edxorg course video and add to intermediate combined video

rachellougee:

https://github.com/mitodl/hq/issues/6048

creating `stg__edxorg__s3__course_video`
adding to `int__combined__course_videos`

### How can this be tested?

Run 

dbt build --select stg__edxorg__s3__course_video
dbt build --select int__combined__course_videos

Then check the following query to see if the video_duration is populated now

select * from "ol_data_lake_production"."ol_warehouse_production_XXXXX_intermediate".int__combined__course_videos
where platform  ='edX.org' and video_duration is not null

VS 

On production, video_duration is not populated
 
select * from "ol_data_lake_production"."ol_warehouse_production_intermediate".int__combined__course_videos
where platform  ='edX.org' and video_duration is not null

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/int__combined__course_videos.sql
+++ b/src/ol_dbt/models/intermediate/combined/int__combined__course_videos.sql
@@ -7,7 +7,6 @@ with video_structure as (
     where coursestructure_block_category = 'video'
 )
 
---- no edxorg course videos until we parse the data from course xml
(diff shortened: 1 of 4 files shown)


quazi-h (approved):
New changes are good, I got records with video_duration populated after running the dbt build commands. My query returned 157,657 records.

Issue #1374, opened 2024-11-12
Create dim_course_content model 


rachellougee:
### Description/Context

As part of https://github.com/mitodl/ol-data-platform/issues/1325, we need a dimensional table for course content including courses, course blocks, chapters,  videos, etc.

### Plan/Design

Create a `dim_course_content` model including the following fields:

content_id - hashed primary key
content_block_id
content_index
content_category
parent_content_block_id
content_hash
more?

The data sources will include MITx Online, xPro, edX.org, and MITx Residential.

PR #1383, opened 2024-11-18
Create dim_course_content slowly changing dimension model

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/1374

Creating `dim_course_content` slowly changing dimension model with the following fields to keep the historical changes
    id
    courserun_readable_id
    block_index
    block_id
    parent_block_id
    block_category
    block_title
    block_metadata
    retrieved_at
    is_latest

### How can this be tested?

Run `dbt build --select dim_course_content`. Tests should pass

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -24,3 +24,54 @@ models:
     description: string, domain name of the platform, e.g. mitxonline.mit.edu
     tests:
     - not_null
+
+- name: dim_course_content
+  description: Slowly Changing Dimension model for course content, including courses,
+    chapters, subsections, problems, videos, and other contents across different platforms.
+  columns:
+  - name: content_id
+    description: string, primary key for this table
+    tests:
+    - not_null
+    - unique
+  - name: courserun_readable_id
+    description: string, the open edX course ID formatted as course-v1:{org}+{course
+      code}+{run_tag} for courses on MITxOnline, xPro and Residential platform, {org}/{course}/{run_tag}
+      for edX.org courses
+    tests:
+    - not_null
+  - name: content_block_index
+    description: int, sequence number giving order in which this block content appears
+      within the course
+    tests:
+    - not_null
+  - name: content_block_id
+    description: string, a unique identifier for each content block within the course,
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/dimensional/_dim__models.yml
```diff
@@ -24,3 +24,54 @@ models:
     description: string, domain name of the platform, e.g. mitxonline.mit.edu
     tests:
     - not_null
+
+- name: dim_course_content
+  description: Slowly Changing Dimension model for course content, including courses,
+    chapters, subsections, problems, videos, and other contents across different platforms.
+  columns:
+  - name: id
```
rachellougee, line 32: I removed the prefix `content_` in this table since Tobias mentioned it in other PR. We originally prefixed it with the table name largely due to the ambiguous columns when joining two tables, but it also introduced other issues like extremely long names in the mart.  Let me know if we agree on this.

rachellougee: I changed it back to `content_id`

rachellougee:
@blarghmatey Do you have any feedback on this PR?

blarghmatey:
I think this looks reasonable. The one thing that stood out is that we likely want to specify that this is course content in the scope of Open edX platforms. This doesn't look like it would fit in the case of trying to accommodate e.g. OCW courses.

rachellougee:
> I think this looks reasonable. The one thing that stood out is that we likely want to specify that this is course content in the scope of Open edX platforms. This doesn't look like it would fit in the case of trying to accommodate e.g. OCW courses.

I will specify that this is for Open edX platforms. The OCW courses don't fit here because they have a different course structure and metadata.

rachellougee:
> but I think maybe possibly this one table should instead be broken out into separate tables(it contains video, problem, discussion and other info for multiple concepts all together). And it might be good to have the JSON block column broken out into seperate columns instead of one column with JSON. Just a potential idea

The current table is designed to provide a high-level overview of the hierarchical structure of course content. As we discussed yesterday, some content, like videos and problems, can be broken out into dimensional tables with more specific metadata flattened from JSON. I can add that later if needed. However, these separate tables would not replace this one because we still need this table to answer questions like how many chapters are in a course and what videos or problems are under each chapter.  Let me know if you have any thoughts on this @KatelynGit

PR #1402, opened 2024-12-05
Create program summary report

rachellougee:

https://github.com/mitodl/hq/issues/6124

- Adding`program_name` (normalized name) to marts__combined_coursesinprogram so that it's easier to summarize the data for multiple track program
- Adding `user_has_completed_program` to marts__combined_program_enrollment_detail as we can't rely on program_certificate_awarded_on from edx to calculate the program certificate
- Creating a new reporting model `program_summary_report`, built using the mart models.

### How can this be tested?

dbt build --select program_summary_report

### DIFF
--- a/src/ol_dbt/models/reporting/_reporting__models.yml
+++ b/src/ol_dbt/models/reporting/_reporting__models.yml
@@ -2,7 +2,7 @@
 version: 2
 models:
 - name: program_summary_report
-  description: MITx program aggregate statistics
+  description: MITx program summary report
   columns:
   - name: program_name
     description: str, name of the program. If program is split over multiple tracks,
--- a/src/ol_dbt/models/reporting/_reporting__models.yml
+++ b/src/ol_dbt/models/reporting/_reporting__models.yml
@@ -1,0 +1,49 @@
+---
+version: 2
+models:
+- name: program_summary_report
+  description: MITx program aggregate statistics
+  columns:
+  - name: program_name
+    description: str, name of the program. If program is split over multiple tracks,
+      the program name is the same for all tracks. e.g. Data, Economics, and Design
+      of Policy
+    tests:
+    - not_null
+  - name: total_enrollments
+    description: int, count of course enrollments in the program, aggregated by user_email
+      and courserun_readable_id
+    tests:
+    - not_null
+  - name: unique_users
+    description: int, count of unique users who took courses in the program, aggregated
+      by user_email
+    tests:
+    - not_null
+  - name: unique_countries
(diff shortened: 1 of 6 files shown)


rachellougee:
@blarghmatey we discussed a separate reporting group for program summary data. But let me know if there is something else we need to do to make this layer surface in the Superset appropriately. I will ask Kate to review the rest next week. Thanks

src/ol_dbt/models/reporting/_reporting__models.yml
```diff
@@ -0,0 +1,49 @@
...
+    tests:
+    - not_null
+  - name: unique_course_certificate_earners
+    description: int, count of unique users who earned a certificate for one of the
+      program's courses, aggregated by user_email
+    tests:
+    - not_null
+  - name: program_certificates
+    description: int, count of program certificates earned for the the program, aggregated
+      by user_email
```
pdpinch, line 49: What does it mean that the program_certificates are aggregated by user_email?

rachellougee: It means that `program_certificates` is aggregated by counting how many distinct user_email. Maybe there is a better way to describe it?

pdpinch: So how does it get counted if a user gets a SDS program certificate and a SCM program certificate? I think that should be counted as two certificates. 

A more subtle example is if a user gets the SDS general certificate and the SDS social science certificate. I'm not sure but I think that should be counted twice as well.

pdpinch: I see. Yes, that's correct. Can you add that clarification and/or an example to the docs? 

> If a user gets a SDS program certificate and a SCM program certificate, that would be counted 2.

> if a user gets the SDS general certificate and the SDS social science certificate, then that would be counted as 1 as I am grouping by program name which is Statistics and Data Science

src/ol_dbt/models/reporting/_reporting__models.yml
```diff
@@ -0,0 +1,49 @@
+---
+version: 2
+models:
+- name: program_summary_report
+  description: MITx program summary report
```
pdpinch, line 5: Does this include xPRO or is it just MITx? 

```suggestion
  description: Program Summary report
```

```suggestion
  description: Program Summary report. A count of all users, enrollments (audit and verified), certificates and program credentials aggregated for each program. 
```

rachellougee: Updated. This includes both MITx and xPro.

pdpinch (approved):

PR #1403, opened 2024-12-05
fix the Incompatible types errors

rachellougee:

NA
I noticed the error on my open PR since pre-commit failed there.

Fixing two pre-commit errors in the main branch
```
src/ol_orchestrate/resources/openedx.py:40: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
src/ol_orchestrate/resources/openedx.py:41: error: Incompatible types in assignment (expression has type "None", variable has type "datetime")  [assignment]
Found 2 errors in 1 file (checked 70 source files)
```

### DIFF
--- a/src/ol_orchestrate/resources/openedx.py
+++ b/src/ol_orchestrate/resources/openedx.py
@@ -37,8 +37,8 @@ class OpenEdxApiClient(ConfigurableResource):
             "Time (in seconds) to allow for requests to complete before timing out."
         ),
     )
-    _access_token: str = PrivateAttr(default=None)
-    _access_token_expires: datetime = PrivateAttr(default=None)
+    _access_token: Optional[str] = PrivateAttr(default=None)
+    _access_token_expires: Optional[datetime] = PrivateAttr(default=None)
     _http_client: httpx.Client = PrivateAttr(default=None)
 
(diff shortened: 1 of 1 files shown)


src/ol_orchestrate/resources/openedx.py
```diff
@@ -37,8 +37,8 @@ class OpenEdxApiClient(ConfigurableResource):
             "Time (in seconds) to allow for requests to complete before timing out."
         ),
     )
-    _access_token: str = PrivateAttr(default=None)
-    _access_token_expires: datetime = PrivateAttr(default=None)
+    _access_token: Optional[str] = PrivateAttr(default=None)
+    _access_token_expires: Optional[datetime] = PrivateAttr(default=None)
```
rachellougee, line 41: I tried to fix it by adding  `# type: ignore` but the pre-commit  throwed `Unused "type: ignore" comment  [unused-ignore]` due to  `--warn-unused-configs` configured in  https://github.com/mitodl/ol-data-platform/blob/main/.pre-commit-config.yaml#L55. Looking at the code, this should works.

@blarghmatey let me know if you prefer to remove `--warn-unused-configs` and add `# type: ignore` instead

PR #1405, opened 2024-12-13
dagster pipeline for edx.org program API ingestion

rachellougee:

https://github.com/mitodl/hq/issues/6191

This adds edxorg_program_metadata_extract pipeline to ingest the MITx program data from the edx.org API scheduled daily. 

It creates two assets: edxorg/processed_data/program_course_metadata and edxorg/processed_data/program_metadata to be ingested by airbyte.

The API credential is added to vault QA https://vault-qa.odl.mit.edu/ui/vault/secrets/secret-data/show/pipelines/edx/org/edxorg_api

### How can this be tested?

export GITHUB_TOKEN=<>
dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/edxorg_api_data_extract.py

can manually materialize two assets from dagster UI


src/ol_orchestrate/resources/edxorg_api.py
blarghmatey, line 1: What's the motivation for duplicating the edX resource? Does the current definition not function with edx.org? I imagine that it should work as-is. It should only require adding the method for fetching the program URLs rather than replicating the whole file. That way there's less chance of diverging functionality and future confusion on which implementation to use.

src/ol_orchestrate/assets/edxorg_api.py
```diff
@@ -0,0 +1,110 @@
...
+    context: AssetExecutionContext, edxorg_api: EdxorgApiClientFactory
+):
+    program_data_generator = edxorg_api.client.get_edxorg_programs()
+
+    mitx_programs = []
+    mitx_program_courses = []
+    data_retrieval_timestamp = datetime.now(tz=UTC).isoformat()
+    for result_batch in program_data_generator:
+        for program in result_batch:
+            # Filter out Non MITx program data as we only need MITx programs for now
```
blarghmatey, line 46: While we only _need_ MITx programs, is there any harm in collecting all of the data that is available? We can do the filtering in the dbt layer for the downstream tables/reporting that we want to do. By collecting the available data as it is available we don't have to worry about losing access to historical data in the future when we decide that we want/need it.

src/ol_orchestrate/resources/edxorg_api.py
rachellougee: The current edX resource does not work as-is. There are two main methods that do not apply to edX.org :
- [_fetch_access_token](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/resources/openedx.py#L71) that calls `f"{self.lms_url}/oauth2/access_token",` inside the method,  this does not work for edX.org token URL https://api.edx.org/oauth2/v1/access_token
- [_fetch_with_auth](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/resources/openedx.py#L95) requires `{"username": self._username, "page_size": page_size}`, which is not needed for edX.org

Duplicating the edX resource is not ideal, but I wanted to avoid modifying the existing implementation, as I will need to validate the new implementation for open edX, but if you think it’s better to modify the existing resource to support edX.org, I can do that. Let me know?

src/ol_orchestrate/assets/edxorg_api.py
rachellougee: There is no harm in collecting all the data, as stated in the issue, this API returns all the organizations that run on edX.org, including HarvardX, Microsoft, UBCx, etc. 24 out of 1165 are MITx. I don't know the use cases for these programs, but we can keep these.

rachellougee:
@blarghmatey Thanks for the review. Based on your feedback, I've made the changes. Please take a look.

Also, I am still experiencing `hvac.exceptions.Forbidden: permission denied, on get https://vault-qa.odl.mit.edu/v1/secret-data/pipelines/edx/edxorg/edx-oauth-client` when running this locally.  So I had to pass the secret to `OpenEdxApiClient` without Vault manually. That worked for me. 
```
export GITHUB_TOKEN=<>
dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/edxorg_api_data_extract.py
```

src/ol_orchestrate/resources/openedx.py
```diff
@@ -92,8 +94,12 @@ def _fetch_with_auth(
         page_size: int = 100,
         extra_params: dict[str, Any] | None = None,
     ) -> dict[Any, Any]:
-        request_params = {"username": self._username, "page_size": page_size}
-        request_params.update(**(extra_params or {}))
+        if self.token_url == f"{self.lms_url}/oauth2/access_token":
+            request_params = {"username": self._username, "page_size": page_size}
+            request_params.update(**(extra_params or {}))
```
blarghmatey, line 99: I think you might want to move this to outside of the `if` block. Passing the request_url with request params included will cause the params to be ignored, so you'll still want to do the parsing and passing of those params via `extra_params` to handle the pagination values.

rachellougee: Updated in https://github.com/mitodl/ol-data-platform/pull/1405/commits/faf044a3738e85f77e60174d9500c6554b1d9125. Thanks for pointing out.

src/ol_orchestrate/resources/openedx.py
```diff
@@ -181,6 +188,25 @@ def get_course_outline(self, course_id: str):
...
+        from the API
+
+        """
+        request_url = "https://discovery.edx.org/api/v1/programs/"
+        response_data = self._fetch_with_auth(request_url)
+        results = response_data["results"]
+        next_page = response_data["next"]
+        yield results
+        while next_page:
+            response_data = self._fetch_with_auth(next_page)
```
blarghmatey, line 206: You'll either want to parse out the `extra_params` from the `next_page` similar to `get_edx_course_ids`. Alternatively, you can update the `fetch_with_auth` method to handle that logic based on the full URL that is passed to the method.

rachellougee: The edx.org API doesn't need the extra param when going to the next page. It's already in `next_page = response_data["next"]`. 

next": "https://discovery.edx.org/api/v1/catalogs/10/courses/?limit=100&offset=400"

I got 1165 programs. Is it not working for you?

PR #1413, opened 2024-12-16
add new total course engagement mart

KatelynGit:
Will be opening up seperate tickets for additional platforms

add new engagement mart (mart_total_course_engagements) which gives counts of video, problem, and discussion engagement as well as percentages 

### How can this be tested?
dbt build --select mart_total_course_engagements

### DIFF
--- a/src/ol_dbt/models/reporting/report_mitxonline_course_engagements.sql
+++ b/src/ol_dbt/models/reporting/report_mitxonline_course_engagements.sql
@@ -103,24 +103,24 @@ select
     , problems_in_courserun.count_problems_in_courserun
     , discussions_paticipated_in_count_tbl.discussions_paticipated_in
     , discussions_in_courserun.discussions_total
-    , case 
+    , case
         when videos_in_courserun.count_videos_in_courserun <> 0
-            then 
-                videos_played_count_tbl.count_videos_played 
+            then
+                videos_played_count_tbl.count_videos_played
                 / videos_in_courserun.count_videos_in_courserun
     end
     as pct_videos_played
-    , case 
+    , case
         when problems_in_courserun.count_problems_in_courserun <> 0
-            then 
-                problems_submitted_count_tbl.count_problems_submitted 
+            then
+                problems_submitted_count_tbl.count_problems_submitted
                 / problems_in_courserun.count_problems_in_courserun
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/reporting/report_mitxonline_course_engagements.sql
```diff
@@ -0,0 +1,145 @@
...
+    group by
+        courserun_readable_id
+        , user_username
+)
+
+, videos_in_courserun as (
+    select
+        courserun_readable_id
+        , count(distinct video_id) as count_videos_in_courserun
+    from stg_edxval_coursevideo
```
rachellougee, line 41: The underlying table for `videos_in_courserun` should be from the course structure `int__combined__course_structure`.  There is a bug in our data ingestion pipeline for this data.

The `stg__mitxonline__openedx__mysql__edxval_coursevideo` is an internal table to associate video_id to a course. My understanding is not every video in this table is used in the course.

src/ol_dbt/models/reporting/report_mitxonline_course_engagements.sql
```diff
@@ -0,0 +1,145 @@
...
+
+, problems_in_courserun as (
+    select
+        courserun_readable_id
+        , count(distinct coursestructure_block_id) as count_problems_in_courserun
+    from stg_edxval_problems
+    where coursestructure_block_category = 'problem'
+    group by
+        courserun_readable_id
+)
```
rachellougee, line 65: The underlying table for`problems_in_courserun` should also from course structure int__combined__course_structure. 

We can't rely on `stg__mitxonline__openedx__mysql__courseware_studentmodule`  to get how many problems in a course. Here is the edx doc for [this table](https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/sql_schema.html#courseware-studentmodule)

rachellougee (commented):
As discussed in our meeting today, this report is specific to MITx Online. We should consider aggregating similar data for xPro and Residential as well, to avoid having to create separate reports for each platform.

Ideally, these reports should be generated/aggregated from our mart tables. If the base tables are not available in the mart layer, we should create them. I can help with that if its too much work. Just let me know.

PR #1416, opened 2024-12-20
Fix program learner report error

rachellougee:

NA
https://pipelines.odl.mit.edu/runs/4f26f858-f9f1-4405-9deb-9caa359533a3

- fixing the following database error on "date program certificate awarded" in stg__edxorg__s3__program_learner_report
```
  Database Error in model stg__edxorg__s3__program_learner_report (models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql)
  TrinoUserError(type=USER_ERROR, name=INVALID_FUNCTION_ARGUMENT, message="Invalid format: "2024-11-07 00:00:00 Z" is malformed at " 00:00:00 Z"", query_id=20241220_064556_42933_7jh9z)
```
- Updating `program_certificate_awarded_on` to be the first date if there are multiple "date program certificate awarded" for the learner on the same program

### How can this be tested?

dbt build --select stg__edxorg__s3__program_learner_report

This will take almost two hours. https://github.com/mitodl/hq/issues/6342 will address the performance issue

### DIFF
--- a/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql
+++ b/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql
@@ -22,11 +22,11 @@ with source as (
     from source
 )
 
-, add_program_cert_latest as (
+, aggregated_program_certificate as (
     select
         cast("user id" as integer) as user_id
         , "program uuid" as program_uuid
-        , max(program_certificate_awarded_dt) as latest_program_cert_award_on
+        , min(program_certificate_awarded_dt) as earliest_program_cert_award_on
         , max("completed program") as ever_completed_program
     from source_sorted
     group by 1, 2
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__program_learner_report.sql
```diff
@@ -7,20 +7,26 @@ with source as (
...
     from source
 )
 
-, add_program_cert_latest as (
+, aggregated_program_certificate as (
     select
         cast("user id" as integer) as user_id
         , "program uuid" as program_uuid
-        , max(program_certificate_awarded_dt) as latest_program_cert_award_on
+        , min(program_certificate_awarded_at) as earliest_program_cert_award_on
```
rachellougee, line 29: @pdpinch @KatelynGit Per Slack conv, I've updated program_certificate_awarded_on to be the first date if the learner has completed the program multiple times. 

 One potential issue is if there is a data error in 2U data, and they fix it in the latest file, this date will be stuck to the first date. Not sure how likely that will happen.

PR #1434, opened 2025-01-09
Update openedx api resource for Retry-After header handling

rachellougee:

https://github.com/mitodl/hq/issues/6191#issuecomment-2580577438

Updating _fetch_with_auth in OpenEdxApiClient resource to handle 409 Too Many Requests errors by parsing the `Retry-After` header. This change will respect the recommended retry delay from the API if the API request is throttled and the Retry-After time exceeds 60 seconds.

### How can this be tested?

This can be tested on QA, as the local test requires additional steps due to an issue with Vault and the GitHub token implementation

Modify this code to test locally
```
if DAGSTER_ENV == "dev":
    vault = Vault(vault_addr=VAULT_ADDRESS, vault_auth_type="github", token=os.environ.get('VAULT_TOKEN'))
```
export VAULT_TOKEN= - This can be generated when making a POST request to https://vault-qa.odl.mit.edu/v1/auth/github/login  with your GitHub token

Run dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/edxorg_api_data_extract.py

### DIFF
--- a/src/ol_orchestrate/resources/openedx.py
+++ b/src/ol_orchestrate/resources/openedx.py
@@ -110,7 +110,9 @@ class OpenEdxApiClient(ConfigurableResource):
             response.raise_for_status()
         except httpx.HTTPStatusError as error_response:
             if error_response.response.status_code == TOO_MANY_REQUESTS:
-                time.sleep(60)
+                retry_after = error_response.response.headers.get("Retry-After", 60)
+                delay = int(retry_after) if retry_after.isdigit() else 60
+                time.sleep(delay)
                 return self._fetch_with_auth(
                     request_url, page_size=page_size, extra_params=extra_params
                 )


quazi-h (approved):
I was able to get dagster running locally with these changes. The logic for the retry-after from the openedx api client response makes sense me. When I attempted to materialize the two assets (program_course_metadata and program_metadata) with my local instance, a new edxorg_program_metadata run is created but I don't think it is actually connecting to the openedx client because I do not see the 429 error or the execution of the retry logic in the logs for the run. I can approve this now so we can test in QA but I wasn't sure if there was something I am missing in trying to test this on a local dagster instance.

rachellougee:
>  a new edxorg_program_metadata run is created but I don't think it is actually connecting to the openedx client because I do not see the 429 error or the execution of the retry logic in the logs for the run.

@quazi-h Thanks for the review. I am not sure what you meant by "I don't think it is actually connecting to the openedx client " but if a new edxorg_program_metadata run was created successfully then you had to connect to open edx client with credentials from [vault QA](https://vault-qa.odl.mit.edu/ui/vault/secrets/secret-data/show/pipelines/edx/org/edxorg_api)

You would see 429 errors only if the API requests are throttled, which occurred last week. The change was made to ensure that we respect the API's retry policy, as specified in the API header, to avoid retrying too frequently and getting blocked. Peter raised this issue with our partner last week, and it seems the issue may have been resolved. I will run [program_course_metadata](https://pipelines-qa.odl.mit.edu/assets/edxorg/processed_data/program_course_metadata?view=events) on QA, which I couldn't run last week due to this error.

PR #1438, opened 2025-01-13
Increasing timeout to call check_course_export_status

quazi-h:
https://github.com/mitodl/hq/issues/6413

After looking at error logs on the openedx Dagster pipelines, we found that these two:
mitxonline_edx_course_pipeline and xpro_edx_course_pipeline were hitting 500 errors.
This was due to a race condition when the pipeline tries to check on the course export job status too quickly and the resulting behavior is observed.

- Increased the timeout between calls of the check_course_export_status method from 5 to 60 seconds.
- Updated the check_course_export_status code to use the _fetch_with_auth method.

### DIFF
--- a/src/ol_orchestrate/assets/openedx.py
+++ b/src/ol_orchestrate/assets/openedx.py
@@ -150,7 +150,7 @@ def course_xml(context: AssetExecutionContext, courseware):  # noqa: ARG001
     failed_exports: set[str] = set()
     tasks = exported_courses["upload_task_ids"]
     while len(successful_exports.union(failed_exports)) < len(tasks):
-        time.sleep(timedelta(seconds=5).seconds)
+        time.sleep(timedelta(seconds=60).seconds)
         for course_id, task_id in tasks.items():
             task_status = context.resources.openedx.client.check_course_export_status(
                 course_id,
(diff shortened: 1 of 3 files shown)


src/ol_orchestrate/assets/openedx.py
```diff
@@ -150,7 +150,7 @@ def course_xml(context: AssetExecutionContext, courseware):  # noqa: ARG001
     failed_exports: set[str] = set()
     tasks = exported_courses["upload_task_ids"]
     while len(successful_exports.union(failed_exports)) < len(tasks):
-        time.sleep(timedelta(seconds=5).seconds)
+        time.sleep(timedelta(seconds=60).seconds)
```
blarghmatey, line 153: The asset version of the task is functioning well, but it doesn't hurt to increase the delay a bit (maybe ~15-20 seconds). The longer timeout needs to go here: https://github.com/mitodl/ol-data-platform/blob/main/src/ol_orchestrate/ops/open_edx.py#L635

That's the op definition that is used by the legacy edX data export pipeline that IRx relies on.

blarghmatey (approved):

src/ol_orchestrate/assets/openedx.py
quazi-h: Updated the op definition to 60 and decreased the asset one to 20 seconds.

PR #1439, opened 2025-01-15
add logging and pass extra_params to edxorg program API

rachellougee:

https://github.com/mitodl/hq/issues/6191

Adding some logging to help debug `edxorg_api_data_extract` on QA
Updating the `get_edxorg_programs` to pass extra_params explicitly, instead of relying on full URL

### How can this be tested?

 dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/edx/edxorg_api_data_extract.py

go to http://127.0.0.1:3000/assets/edxorg/processed_data to manually trigger the materialization.

you should be able to see the logging

```
Total programs to extract: 1171 programs

Extracted a batch of 20 programs. Total so far: 20 programs.
-----
Extracted a batch of 11 programs. Total so far: 1171 programs.
Total extracted 1171 programs....
Total extracted 3067 program courses....
```

image

### DIFF
--- a/src/ol_orchestrate/assets/edxorg_api.py
+++ b/src/ol_orchestrate/assets/edxorg_api.py
@@ -43,10 +43,10 @@ def edxorg_program_metadata(
     total_extracted_count = 0
     for i, data in enumerate(program_data_generator):
         if i == 0:
-            count, result_batch = data # First iteration gives total count and results
+            count, result_batch = data  # First iteration gives total count and results
             context.log.info("Total programs to extract: %d programs", count)
         else:
-            result_batch = data # Subsequent iterations give only results
+            result_batch = data  # Subsequent iterations give only results
 
         batch_count = len(result_batch)
         total_extracted_count += batch_count
--- a/src/ol_orchestrate/assets/edxorg_api.py
+++ b/src/ol_orchestrate/assets/edxorg_api.py
@@ -40,14 +40,20 @@ def edxorg_program_metadata(
     edxorg_programs = []
     edxorg_program_courses = []
     data_retrieval_timestamp = datetime.now(tz=UTC).isoformat()
-    total_count = 0
-    for result_batch in program_data_generator:
+    total_extracted_count = 0
+    for i, data in enumerate(program_data_generator):
+        if i == 0:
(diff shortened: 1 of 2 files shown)


src/ol_orchestrate/assets/edxorg_api.py
```diff
@@ -40,7 +40,15 @@ def edxorg_program_metadata(
     edxorg_programs = []
     edxorg_program_courses = []
     data_retrieval_timestamp = datetime.now(tz=UTC).isoformat()
+    total_count = 0
```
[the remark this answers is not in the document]
rachellougee: yes, the initial API includes the total number of programs. I will add that in.

src/ol_orchestrate/assets/edxorg_api.py
```diff
@@ -40,7 +40,21 @@ def edxorg_program_metadata(
     edxorg_programs = []
     edxorg_program_courses = []
     data_retrieval_timestamp = datetime.now(tz=UTC).isoformat()
-    for result_batch in program_data_generator:
+    total_extracted_count = 0
+    for i, data in enumerate(program_data_generator):
+        if i == 0:
+            count, result_batch = data  # First iteration gives total count and results
```
blarghmatey, line 46: This is unnecessary. I think the total count tracking is only worthwhile if it's an attribute of the initial API response.

rachellougee: The count is in the initial API response. But I only returned the `results` from the initial call, so this change is to parse both `count` and `results` from the initial call. 
```
    "count": 1167,
    "next": "https://discovery.edx.org/api/v1/programs/?limit=20&offset=20",
    "previous": null,
    "results": []

````
Otherwise, we would not know how many we are expecting. Do you think I don't need to log that?

quazi-h (approved):
I was able to test this locally after explicitly passing my vault token to the edxorg_api_data_extract as well as the vault hvac client. When attempting to materialize the two assets, the pipeline was able to extract 1172 courses before erroring out on a local limitation (not being able to write to /opt/dagster). Approving this so it can be tested in QA.

PR #1450, opened 2025-01-27
add user metrics to total engagement mart

KatelynGit:

adds user_username, videos_user_watched, problems_user_submitted, and user_discussion_count  fields to the marts__combined_total_course_engagements table

### How can this be tested?
dbt build --select marts__combined_total_course_engagements


rachellougee (approved):
LGTM. 

Just a side note: I noticed that the large `Union All` queries for video, problem , and discussion are duplicated across two marts. It might be worth considering a refactor later.

Issue #916, opened 2023-11-06
All the MicroMasters program data should flow into int__micromasters__* tables


rachellougee:
# Description/Context

Currently, DEDP program data are in `int__micromasters__*` tables. But any non DEDP program will not be in these MicroMasters intermediate tables, so it would require code changes if any new MicroMasters program added to MITx Online database in the near future.

Since we've added `program_type` in `raw__mitxonline__app__postgres__courses_program`, which identify any MicroMasters program. We should be able to modify these [subqueries](https://github.com/mitodl/ol-data-platform/tree/main/src/ol_dbt/models/intermediate/micromasters/subqueries) from MITx Online to add any MicroMasters program data there, not just DEDP.

# Plan/Design

All MicroMasters program data should be in these tables
- int__micromasters__course_certificates
- int__micromasters__course_grades
- int__micromasters__program_certificates

PR #1452, opened 2025-01-28
add program_readable_id to micromasters program certificate intermediate and mart

rachellougee:

Close out https://github.com/mitodl/hq/issues/3852

- adding `program_readable_id` to int__micromasters__program_certificates and marts__micromasters_program_certificates 
    - For the micromasters programs that run on edx, program_readable_id is created internally. I've created a macro for this since it will be used in other models later.
    - For the micromasters programs that run on MITx online, program_readable_id is pulled from the application database

### How can this be tested?

Run
`dbt build --select int__micromasters__program_certificates marts__micromasters_program_certificates`, tests should all pass

### DIFF
--- a/src/ol_dbt/macros/generate_program_readable_id.sql
+++ b/src/ol_dbt/macros/generate_program_readable_id.sql
@@ -1,0 +1,19 @@
+{% macro generate_micromasters_program_readable_id(micromasters_program_id, program_title) %}
+ -- Generate readable IDs for MicroMasters programs that runs on the edx platform
+    --- This is needed for programs that have not migrated to MITx Online yet
+    case
+        when {{ micromasters_program_id }} = 4
+             --- Statistics and Data Science has multiple tracks, we need to extract the track name and add it to
+             --- the readable id
+             then 'program-v1:MITx+SDS+' || replace(regexp_extract({{ program_title }}, '\((.*?)\)', 1),' ', '-')
+        when {{ micromasters_program_id }} = 5
+             --- Finance (active) and MIT Finance (retired) has the same readable ID
+             then 'program-v1:MITx+FIN'
+        else
+              'program-v1:MITx+' || upper(array_join(
+                  transform(split({{ program_title }}, ' '), -- noqa: PRS
+                    x -> substring(x, 1, 1)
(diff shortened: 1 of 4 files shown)


pdpinch:
Based on https://github.com/mitodl/hq/issues/6601#issuecomment-2619278638I think we should include some logic here to handle the multiple SCM program as well. 

Like FIN, I expect SCM will want to report on their legacy (2014-2016) and current (2016+) credential holders in aggregate.

rachellougee:
> Like FIN, I expect SCM will want to report on their legacy (2014-2016) and current (2016+) credential holders in aggregate.

@pdpinch Currently, there is one SCM in `int__micromasters__program_certificates` and `marts__micromasters_program_certificates`, the legacy version titled "Supply-Chain Management",  is not included because it falls under the 'xSeries' program. I noticed this issue when you asked about it. I think we will need to add 'Supply-Chain Management' to the micromasters intermediate and mart first.  Do you want me to handle this in this PR, or should I create a separate issue for it?"

src/ol_dbt/macros/generate_program_readable_id.sql
```diff
@@ -0,0 +1,19 @@
+{% macro generate_micromasters_program_readable_id(micromasters_program_id, program_title) %}
+ -- Generate readable IDs for MicroMasters programs that runs on the edx platform
+    --- This is needed for programs that have not migrated to MITx Online yet
+    case
+        when {{ micromasters_program_id }} = 4
```
pdpinch, line 5: I'm curious where these integers come from? Is that how they're defined in the MicroMasters database? We will probably never add a new program there -- new ones will be in MITx Online or, more likely, Learn. Is that OK?

rachellougee: Yes, these integers come from MicroMasters https://micromasters.mit.edu/admin/courses/program/. The generated readable ID applies only to the existing MicroMasters on edX. For new programs, we will be using the program readable ID on MITx Online, but this will require some code changes. Here is the issue I created a while ago https://github.com/mitodl/ol-data-platform/issues/916

PR #1454, opened 2025-01-30
Populate micromasters program_readable_id in combined program enrollment mart

rachellougee:

follow up https://github.com/mitodl/ol-data-platform/pull/1452

Populating the program_readable_id in marts__combined_program_enrollment_detail for the "MicroMasters" program. This is similar to how it was done in the previous linked PR. All other programs are not affected by this PR.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/macros/generate_program_readable_id.sql
+++ b/src/ol_dbt/macros/generate_program_readable_id.sql
@@ -9,7 +9,7 @@
         when {{ micromasters_program_id }} = 5
              --- Finance (active) and MIT Finance (retired) has the same readable ID
              then 'program-v1:MITx+FIN'
-        else
+         when {{ micromasters_program_id }} is not null then
               'program-v1:MITx+' || upper(array_join(
                   transform(split({{ program_title }}, ' '), -- noqa: PRS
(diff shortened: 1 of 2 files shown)


src/ol_dbt/macros/generate_program_readable_id.sql
```diff
@@ -9,7 +9,7 @@
         when {{ micromasters_program_id }} = 5
              --- Finance (active) and MIT Finance (retired) has the same readable ID
              then 'program-v1:MITx+FIN'
-        else
+         when {{ micromasters_program_id }} is not null then
```
rachellougee, line 12: I had to change this because there are other non-micromasters programs in marts__combined_program_enrollment_detail from edx, and I don't think we want to generate the readable_id for these non-micromasters ones

rachellougee:
https://github.com/mitodl/hq/issues/6601 was created to populate these xSeries programs, including Supply-Chain Management

This PR is just to populate these MicroMasters programs in the combined program enrollment so that it's consistent with my previous PR https://github.com/mitodl/ol-data-platform/pull/1452

PR #1464, opened 2025-02-06
create studentmodulehistoryextended under external/irx models

rachellougee:

https://github.com/mitodl/hq/issues/6049
https://github.com/mitodl/hq/issues/6661

Creating `irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended` under external/irx models. These models are specifically designed for data delivery to IRx

### How can this be tested?

dbt build --select irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended

### DIFF
--- a/src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended.sql
+++ b/src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended.sql
@@ -1,0 +1,19 @@
+with studentmodulehistoryextended as (
+    select * from {{ ref('stg__mitxresidential__openedx__courseware_studentmodulehistoryextended') }}
+)
+
+, studentmodule as (
+    select * from {{ ref('stg__mitxresidential__openedx__courseware_studentmodule') }}
+)
+
+select
+    studentmodule.courserun_readable_id as course_id
+    , studentmodule.user_id
+    , studentmodule.coursestructure_block_id as module_id
+    , studentmodule.coursestructure_block_category as module_type
+    , studentmodulehistoryextended.studentmodule_state_data as state_data
+    , studentmodulehistoryextended.studentmodule_problem_grade as grade
+    , studentmodulehistoryextended.studentmodule_problem_max_grade as max_grade
+    , studentmodulehistoryextended.studentmodule_created_on as created
+from studentmodulehistoryextended
+inner join studentmodule on studentmodulehistoryextended.studentmodule_id = studentmodule.studentmodule_id
--- a/src/ol_dbt/models/external/irx/mitx/_irx_mitx__models.yml
+++ b/src/ol_dbt/models/external/irx/mitx/_irx_mitx__models.yml
@@ -209,3 +209,32 @@ models:
     description: int, display_tag_filter_strategy
   - name: consecutive_days_visit_count
     description: int, consecutive_days_visit_count
+
+- name: irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended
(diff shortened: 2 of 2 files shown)


src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended.sql
```diff
@@ -0,0 +1,19 @@
+with studentmodulehistoryextended as (
+    select * from {{ ref('stg__mitxresidential__openedx__courseware_studentmodulehistoryextended') }}
+)
+
+, studentmodule as (
+    select * from {{ ref('stg__mitxresidential__openedx__courseware_studentmodule') }}
+)
+
+select
+    studentmodule.courserun_readable_id as course_id
```
rachellougee, line 10: matching these columns with the raw table, as we use the names from the raw tables in the other IRX tables

quazi-h: Should they be matched with the column names from the raw table or the staging table? I think the user_id and studentmodule_state_data might need to be renamed if that is the case (see code suggestions below).
[stg__mitxresidential__openedx__courseware_studentmodule.sql](https://github.com/mitodl/ol-data-platform/blob/a4045a31ef830cb1e31c91ec5bda2dab61b8fa85/src/ol_dbt/models/staging/mitxresidential/stg__mitxresidential__openedx__courseware_studentmodule.sql#L19)
[stg__mitxresidential__openedx__courseware_studentmodulehistoryextended.sql](https://github.com/mitodl/ol-data-platform/blob/a4045a31ef830cb1e31c91ec5bda2dab61b8fa85/src/ol_dbt/models/staging/mitxresidential/stg__mitxresidential__openedx__courseware_studentmodulehistoryextended.sql#L29)

src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended.sql
```diff
@@ -0,0 +1,19 @@
...
+    , studentmodule.user_id
+    , studentmodule.coursestructure_block_id as module_id
+    , studentmodule.coursestructure_block_category as module_type
+    , studentmodulehistoryextended.studentmodule_state_data as state_data
```
quazi-h, line 14:
```suggestion
    , studentmodulehistoryextended.studentmodule_state_data as state
```

quazi-h (approved):
Changes look good, I just left a comment about the column names since you mentioned we should match the raw table. DBT build ran without errors and the data in the table looks valid. I got a total count of 61,995,550 records in the table I created.

src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__courseware_studentmodulehistoryextended.sql
rachellougee: The `user_id` doesn't need to be renamed. I mentioned it in the context of the existing IRx models. In this case, user_id is more appropriate. See https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/external/irx/mitx/irx__mitx__openedx__mysql__auth_userprofile.sql#L11

rachellougee:
Thanks for the review. THB, these columns' names don't matter that much, I mentioned it most for consistently with the existing models. For example `user_id` in this table is consistent with other foreign key reference `user_id` in irx__mitx__openedx__mysql__auth_userprofile

PR #1465, opened 2025-02-07
alter courserun mart total engagement counts to pick up engagement not captured by course structure

KatelynGit:

Some videos and other engagement opportunities are not captured by the course structure table. Examples include videos on youtube we'd like to capture these better so I've modified the way we calculate engagement metrics on the total combined engagement mart table(marts__combined_total_course_engagements). 

### How can this be tested?
dbt build --select marts__combined_total_course_engagements

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined_total_course_engagements.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_total_course_engagements.sql
@@ -216,34 +216,34 @@ with combined_engagements as (
 )
 
 , user_video_correction as (
-    select 
+    select
         platform
         , courserun_readable_id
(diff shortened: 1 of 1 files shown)


quazi-h (approved):
I was able to build the marts__combined_total_course_engagements model without any errors and the data looks good when querying the "total_courserun_" columns in Starburst.

Is the reason why we are comparing the counts from the coursestructure tables to the  user_correction tables (populated from courseactivity) and using the greater value because there is overlap between those two? I know you mentioned that some of the engagement opportunities are not captured in the coursestructure table but I was wondering why we wouldn't combine those counts instead of using one or the other?

PR #1487, opened 2025-02-24
Ingest sloan courses via api

rachellougee:

https://github.com/mitodl/hq/issues/6687

Creating new dagster code location, definition, and assets for learning resources such as Sloan courses.
Also, creating a new OAuthApiClient resource that is not tied to the edx instance, and modifying OpenEdxApiClient to inherit from it

### How can this be tested?

export GITHUB_TOKEN=
docker compose up
Go to http://127.0.0.1:3000/locations/ol_orchestrate.definitions.learning_resource.extract_api_data/assets, ensure 2 assets listed.  
Click on Materialize for the two assets
```
Total extracted 103 Sloan courses....
Total extracted 226 Sloan course offerings....
```
The existing assets that use OpenEdxApiClient should still work. This could be tested by materializing any of http://127.0.0.1:3000/locations/ol_orchestrate.definitions.edx.retrieve_edxorg_raw_data/assets

### DIFF
--- a/src/ol_orchestrate/resources/oauth.py
+++ b/src/ol_orchestrate/resources/oauth.py
@@ -78,7 +78,7 @@ class OAuthApiClient(ConfigurableResource):
         response.raise_for_status()
         return response.json()["username"]
 
-    def _fetch_with_auth(
+    def fetch_with_auth(
         self,
         request_url: str,
         page_size: int = 100,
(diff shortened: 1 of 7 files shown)


src/ol_orchestrate/resources/oauth.py
```diff
@@ -0,0 +1,158 @@
...
+        return self._fetch_with_auth(course_url)
+
+    def get_sloan_course_offerings(self):
+        """
+        Retrieve the course offerings data from their API as JSON
+
+        returns: JSON document representing an array of course offering objects
+        """
+        course_offering_url = "https://mit-unified-portal-prod-78eeds.43d8q2.usa-e2.cloudhub.io/api/course-offerings"
+        return self._fetch_with_auth(course_offering_url)
```
blarghmatey, line 127: These method seem like they're unnecessary in this context. If we wanted to make a Sloan API specific resource we could put them there, but I think it's easy enough to just use the `fetch_with_auth` and pass the URL in question. Another reason to not have these methods here is that they will then be inherited by the edX resource (which won't break anything, but is a leaky abstraction)

rachellougee: Yes, I realized that they don't belong here. I moved them to the asset file so we don't need to create a new resource file for these two simple methods.

blarghmatey (approved):

PR #1493, opened 2025-02-28
add new all coupons table to marts

KatelynGit:
https://github.com/mitodl/hq/issues/6798

Creates a new mart table with all coupons including unredeemed coupons. Also adds some additional fields to the all orders mart table in order to have the data that was previously only here https://bi.odl.mit.edu/queries/983/ available in the marts.

### How can this be tested?
dbt build --select marts__combined__orders marts__mitxpro_all_coupons


quazi-h (approved):
dbt build ran successfully and I confirmed the new marts__mitxpro_all_coupons table as well as the two new columns in marts__combined__orders.

PR #1499, opened 2025-03-05
Update salesforce models

rachellougee:

https://github.com/mitodl/hq/issues/6787

The new Airbyte connection "OL Salesforce → S3 Data Lake" is created with a new prefix table name. This PR is to update the raw table names and fix any dbt build error for the deprecated airbyte meta fields in Salesforce staging models and re-enable the disabled staging/intermediate models

### How can this be tested?

dbt build --select +int__salesforce__opportunity +int__salesforce__opportunitylineitem int__mitxpro__b2becommerce_b2border

You can either ignore the warning dbt_expectations_expect_table_row_count_to_equal_other_table_int__mitxpro__b2becommerce_b2border since this command doesn't refresh stg__mitxpro__app__postgres__b2becommerce_b2border, or refresh +int__mitxpro__b2becommerce_b2border so the count matches

### DIFF
--- a/src/ol_dbt/.dbtignore
+++ b/src/ol_dbt/.dbtignore
@@ -1,10 +1,1 @@
 # .dbtignore
-
-# ignore individual files
-int__salesforce__opportunitylineitem.sql
-int__salesforce__opportunity.sql
-stg__salesforce__opportunitylineitem.sql
-stg__salesforce__opportunity.sql
-_salesforce__sources.yml
-_stg_salesforce__models.yml
-_int_salesforce__models.yml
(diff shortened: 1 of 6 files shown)


quazi-h (approved):
I was able to run the updated models with no issues. I verified the new V1 raw tables exist but I noticed the original tables (raw__thirdparty__salesforce__Opportunity and raw__thirdparty__salesforce__OpportunityLineItem) still exist - should they?

rachellougee:
> I noticed the original tables (raw__thirdparty__salesforce__Opportunity and raw__thirdparty__salesforce__OpportunityLineItem) still exist - should they?

Yes, the old tables are still there for historical records, but they are no longer updated. I mentioned it in the linked issue.

PR #1500, opened 2025-03-07
Update edx tracking log staging models for the new airbyte connector

rachellougee:

https://github.com/mitodl/hq/issues/6787

The remaining dbt changes after switching to the new S3 Data Lake destination connector

### How can this be tested?

Since the initial runs for 4 new connections are still running (hope to be done over the weekend), there is no data in 4 of these tables. But the changes are the same. I tested the edx.org tracking log table, which has data.

dbt build --select stg__edxorg__s3__tracking_logs__user_activity


rachellougee:
You can ignore the gemini-code-assist feedback. It has no context on the macro `deduplicate_raw_table` that has most_recent_source as CTE.

quazi-h (approved):
Changes look good, I had no issues building these dbt models.

PR #1508, opened 2025-03-13
Course Export Error Handling for Closed Partitions

quazi-h:
https://github.com/mitodl/hq/issues/6729

- Refactored the course_version_sensor to return a list of RunRequests instead of AssetMaterializations. This should trigger pipeline runs for the updated course runs instead of recording asset events.
- Adding logic to handle closed partitions when a course no longer exists in the edxorg api and returns a 404 when the export_courses function is called.

### How can this be tested?
Get the Dagster pipelines running locally:
```
export GITHUB_TOKEN=<your_github_token>
docker compose up
```
Try to materialize one of the problematic courses:
mitxonline (course-v1:MITxT+14.100x+1T2025)
mitx (course-v1:MITx+7.01ASEr_14_2+2024_Summer)

### Checklist:
- [ ] Confirm course version sensor changes fixes noisy course structure/course xml failures
- [ ] Confirm course_export calls for courses that no longer exist are handled appropriately and do not cause a pipeline failure.


src/ol_orchestrate/assets/openedx.py
```diff
@@ -141,10 +142,31 @@ def course_structure(context: AssetExecutionContext, courseware):  # noqa: ARG00
...
+    NOT_FOUND = 404
     course_key = context.partition_key
-    exported_courses = context.resources.openedx.client.export_courses(
-        course_ids=[course_key],
-    )
+    try:
+        exported_courses = context.resources.openedx.client.export_courses(
+            course_ids=[course_key],
+        )
+    except httpx.HTTPStatusError as e:
```
blarghmatey, line 151: This won't work for the course_xml use case unfortunately. This request tells edX to register an asynchronous Celery task to do the export. edX will give a success response to the HTTP request and then generate an error when it actually tries to run the course export. To determine if the course is active in the platform we'll need to use a different endpoint to check for a 404. Instead you can send a request to `/api/courses/v1/courses/{course_key}/` and as long as it gets a 200 response then trigger the XML export.

src/ol_orchestrate/sensors/openedx.py
```diff
@@ -127,9 +119,9 @@ def course_version_sensor(
             )
             run_requests.append(
                 RunRequest(
-                    asset_selection=AssetKey(
-                        (openedx.deployment, "openedx", "courseware")
-                    ),
+                    asset_selection=[
+                        AssetKey((openedx.deployment, "openedx", "courseware"))
+                    ],
```
blarghmatey, line 124: I think we might need to remove the asset_selection from the RunRequest since it is being set at the sensor level.

blarghmatey (approved):

PR #1510, opened 2025-03-17
enhance all coupons mart with additional fields to replace redash reports

KatelynGit:
https://github.com/mitodl/hq/issues/6875

enhance all coupons mart with additional fields to replace redash reports including:
couponpaymentversion_num_coupon_codes, couponpaymentversion_max_redemptions, couponpayment_name, couponpaymentversion_id, couponpaymentversion_created_on, couponpaymentversion_discount_amount_text, company_name, b2border_contract_number

### How can this be tested?
dbt build --select marts__mitxpro_all_coupons

### DIFF
--- a/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_all_coupons.sql
+++ b/src/ol_dbt/models/marts/mitxpro/marts__mitxpro_all_coupons.sql
@@ -31,14 +31,14 @@ with allcoupons as (
 )
 
 , redeemed_b2b_coupons as (
-    select 
+    select
         b2bcoupon_id
         , b2border_contract_number
     from allorders
     where
         redeemed = true
         and coupon_id is null
-    group by 
+    group by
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
The dbt run is good. But I randomly checked a couple of duplicated coupon_id records, which don't seem to match with production.
```
select coupon_code, activated_on, expires_on, couponpaymentversion_id 
from  "ol_data_lake_production"."ol_warehouse_production_kdelaney_mart".marts__mitxpro_all_coupons
where coupon_id=9179
```

These two coupon payment versions should have two different `expires_on`, but they are the same in your mart table. I would expect they match with production here
https://xpro.mit.edu/admin/ecommerce/couponpaymentversion/705/change/
https://xpro.mit.edu/admin/ecommerce/couponpaymentversion/1197/change/


Approving it but we might want to take a look at it

PR #1514, opened 2025-03-21
Dagster refactor course export partitions bugfix 2

quazi-h:
https://github.com/mitodl/hq/issues/6729

- Refactored the course_version_sensor to return a list of RunRequests instead of AssetMaterializations. This should trigger pipeline runs for the updated course runs instead of recording asset events.
- Adding logic to handle closed partitions when a course no longer exists in the edxorg api and returns a 404 when the export_courses function is called.

### How can this be tested?
Get the Dagster pipelines running locally:
```
export GITHUB_TOKEN=<your_github_token>
docker compose up
```
Try to materialize one of the problematic courses:
mitxonline (course-v1:MITxT+14.100x+1T2025)
mitx (course-v1:MITx+7.01ASEr_14_2+2024_Summer)
Create a dummy course, materialize the assets, delete the course and try materializing the assets again.

### Checklist:
- [ ] Confirm course version sensor changes fixes noisy course structure/course xml failures
- [ ] Confirm course_export calls for courses that no longer exist are handled appropriately and do not cause a pipeline failure.


src/ol_orchestrate/resources/openedx.py
```diff
@@ -51,18 +58,46 @@ def check_course_status(self, course_id: str) -> int:
...
-                raise ValueError(err_msg) from e
-            else:
-                return e.response.status_code
+            return e.response.status_code
         return 200
 
+    def backfill_asset(
+        self, context: AssetExecutionContext, course_key: str
+    ) -> Optional[Output]:
+        """Query the edX platform to determine if the course is active.
```
blarghmatey, line 71: This doc-string is misleading as it doesn't do any querying of the edX API.

src/ol_orchestrate/assets/openedx.py
```diff
@@ -152,23 +188,27 @@ def course_xml(context: AssetExecutionContext, courseware):  # noqa: ARG001
...
+    # if the course status query results in some other error, raise an exception
+    else:
+        err_msg = f"Unexpected course status: {course_status} for course: {course_key}"
+        context.log.exception(err_msg)
+        raise ValueError(err_msg)
+    if last_materialization:
+        course_file = Path(f"{course_key}.xml.tar.gz")
+        context.log.info("Using last good materialization for %s", course_key)
+        return Output(
+            (course_file, last_materialization.metadata["object_key"]),
```
blarghmatey, line 203: This will cause the actual course file to be overwritten with an empty file.

src/ol_orchestrate/assets/openedx.py
```diff
@@ -80,9 +83,44 @@ def openedx_live_courseware(context: AssetExecutionContext):
...
+        yield Output(
+            (structures_file, structure_object_key),
+            data_version=last_materialization.data_version,
+            metadata={"course_id": course_id, "object_key": structure_object_key},
+        )
+        yield Output(
+            (blocks_file, blocks_object_key),
+            data_version=last_materialization.data_version,
+            metadata={"course_id": course_id, "object_key": blocks_object_key},
+        )
```
blarghmatey, line 118: These will overwrite the actual materialized assets with empty files.

src/ol_orchestrate/assets/openedx.py
```diff
@@ -80,9 +84,37 @@ def openedx_live_courseware(context: AssetExecutionContext):
...
+        yield Output(
+            None,
+            data_version=last_materialization.data_version,
+            metadata={"course_id": course_id, "object_key": structure_object_key},
+        )
+        yield Output(
+            None,
+            data_version=last_materialization.data_version,
+            metadata={"course_id": course_id, "object_key": blocks_object_key},
+        )
```
blarghmatey, line 112: This has the same problem as the course_xml block, where yielding the outputs will overwrite the existing assets with an empty value.

src/ol_orchestrate/lib/assets_helper.py
```diff
@@ -27,3 +35,25 @@ def add_prefix_to_asset_keys(
     return asset_def.with_attributes(
         asset_key_replacements=key_map,
     )
+
+
+def get_last_materialized_partition(
+    context: AssetExecutionContext, course_key: str
```
blarghmatey, line 41: The course key has no benefit in the context of this function and limits its reusability for other asset types.

blarghmatey (changes requested):

blarghmatey (approved):

src/ol_orchestrate/resources/openedx.py
```diff
@@ -51,16 +51,11 @@ def check_course_status(self, course_id: str) -> int:
...
         try:
-            self.fetch_with_auth(request_url, include_status=True)
+            self.fetch_with_auth(request_url)
         except HTTPStatusError as e:
-            if e.response.status_code != NOT_FOUND:
-                err_msg = f"Unexpected response from the edX API for course {course_id}"
-                raise ValueError(err_msg) from e
-            else:
-                return e.response.status_code
+            return e.response["status_code"]
```
blarghmatey, line 58: This should be `e.response.status_code`

src/ol_orchestrate/io_managers/filepath.py
```diff
@@ -119,3 +119,16 @@ def load_input(self, context: InputContext) -> DagsterPath:  # noqa: ARG002
...
+
+
+class LocalFileObjectIOManager(FileObjectIOManager):
+    bucket: Optional[str] = None
+
+    def handle_output(self, context: OutputContext, obj: tuple[Path, str]) -> Nothing:
+        if self.bucket:
+            dest_path = f"{self.bucket}/{self.path_prefix or ''}/{obj[1]}".replace(
+                "//", "/"
+            )
```
blarghmatey, line 131:
```suggestion
            dest_path = Path(f"{self.bucket}/{self.path_prefix or ''}/{obj[1]}".replace(
                "//", "/"
            ))
            dest_path.parent.mkdir(parents=True, exist_ok=True)
```

src/ol_orchestrate/resources/oauth.py
```diff
@@ -108,7 +109,10 @@ def fetch_with_auth(
                     request_url, page_size=page_size, extra_params=extra_params
                 )
             raise
-        return response.json()
+        if include_status:
+            return response.json(), response.status_code
```
quazi-h, line 113: I had added this logic back in because I was seeing "None" being returned in the asset code after deleting my test course. When testing after this change, I was still getting None for the "course_status".

For some reason this code is not returning the actual status_code which should be a 404 (confirmed by querying my deleted course on the api).

src/ol_orchestrate/assets/openedx.py
```diff
@@ -68,65 +71,79 @@ def openedx_live_courseware(context: AssetExecutionContext):
...
+            (blocks_file, blocks_object_key),
+            output_name="course_blocks",
+            data_version=DataVersion(data_version),
+            metadata={
+                "course_id": course_id,
+                "object_key": blocks_object_key,
+            },
+        )
+    # if the course is not found, refer to the last successful materialization
+    elif course_status in {None, NOT_FOUND}:
```
quazi-h, line 140: changed this to catch "None" course_status because that was the response I was seeing after deleting courses.
This is not consistent with what the actual api-docs as I was getting "404" responses for a deleted course there.
With this change, however, I was able to get the pipeline to complete successfully when a deleted course partition was encountered.

quazi-h:
@blarghmatey I think this is in a working state now. I left some comments about some weirdness I observed with the https responses I was seeing from the edx api for a deleted course.

When testing on api-docs, I was getting a 404 but with testing on dagster, I kept getting "None" back from the check_course_status method. I ended up adding logic to look for None or 404 responses and then tried retesting and the pipeline completed successfully.

I am worried if that might lead to unwanted behavior but I guess the whole intention of these changes was to ensure that the whole pipeline doesn't fail if a few courses were removed/missing from the api.

Issue #1325, opened 2024-10-10
Create initial warehouse star schema


blarghmatey:
### User Story

- As an analytics engineer I would like to have a well-factored star schema to improve the flexibility and re-use of data assets in the warehouse

### Description/Context

For detailed context refer to https://github.com/mitodl/hq/pull/5749

This epic covers the creation of the initial set of fact and dimension tables needed to support building the current "mart" tables from those underlying assets.

### Acceptance Criteria

- [ ] All of the current "mart" tables can be created as views or materialized tables that are constructed from the fact and dimension tables
- [ ] All fact and dimension tables are fully documented such that any data consumer will understand the meaning of each field and record in each table

### Plan/Design

Create the following set of dimensional tables (subject to adjustment as we progress through implementation):
#### Dimensions

- `dim_content`: Contains information about content objects including courses,
  course blocks, videos, site pages (e.g. in OCW), etc.
- `dim_platform`: Information about the technical system that a given event is
  executed within the context of.
- `dim_user`: Contains information about system users including
  demographic/profile attributes. This will need to be a
  slowly-changing-dimension model.
- `dim_purchaseable`: Includes information about elements of content and
  services that can be sold for >=$0 (e.g. programs, course runs, bootcamps,
  subscriptions, etc.)

#### Facts

- `tfact_purchase`: Monetary transactions (including $0 transactions, coupon
  redemption, financial aid, etc.)
- `afact_grade`: Information about a grade computed for a user in a course-run
- `tfact_account`: Contains events related to account actions (e.g. created,
  logged in, etc.)
- `tfact_submit`: Events involving a learner submitting information to a system,
  such as answering a question
- `afact_credential`: Information about the granting of a credential
  (e.g. course certificates, program certificates, etc.)
- `tfact_coupon`: Events pertaining to ecommerce coupons including creation,
  consumption, deletion/deactivating
- `tfact_submit`: Contains event details about information submitted by a user
  into a platform, such as answering a question, executing a search, posting a
  forum question, etc.
- `tfact_video`: Contains event information about interactions with a video such
  as play, pause, seek, etc.

Draft ERD (https://erdlab.page.link/7agMHFfKTiL2yxD57)

![Image](https://github.com/user-attachments/assets/30543929-8896-4b32-a161-9d349d14afb9)

PR #1524, opened 2025-04-04
add new user dimensional table

KatelynGit:
https://github.com/mitodl/ol-data-platform/issues/1325

creates a user dimensional table (dim_user)

### How can this be tested?
dbt build --select dim_user

### DIFF
--- a/src/ol_dbt/models/dimensional/dim_user.sql
+++ b/src/ol_dbt/models/dimensional/dim_user.sql
@@ -1,5 +1,5 @@
 with mitx_users as (
-    select 
+    select
         *
         , "Micromasters" as platform
     from {{ ref('int__mitx__users') }}
@@ -126,17 +126,17 @@ select
     , user_company as company
     , user_job_title as job_title
     , user_industry as industry
-    , case 
-        when 
-            user_is_active_on_mitxonline = true 
-            or user_is_active_on_edxorg = true then true 
+    , case
+        when
+            user_is_active_on_mitxonline = true
+            or user_is_active_on_edxorg = true then true
     end as user_is_active
-    , case 
-        when user_joined_on_mitxonline >= user_joined_on_edxorg 
+    , case
+        when user_joined_on_mitxonline >= user_joined_on_edxorg
             then user_joined_on_edxorg
-        when user_joined_on_mitxonline is null 
+        when user_joined_on_mitxonline is null
             then user_joined_on_edxorg
-        else user_joined_on_mitxonline 
+        else user_joined_on_mitxonline
     end as user_joined_on
 from mitx_users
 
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -0,0 +1,212 @@
+with mitx_users_pull as (
+    select *
+    from {{ ref('int__mitx__users') }}
+)
+
+, mitx_users as (
+    select
+        mitx_users_pull.*
+        , 'Micromasters' as platform
```
rachellougee, line 9: This is a bit misleading. 'Micromasters' is not a platform. `mitx_users` is a combination of MITx Online and edX.org users, not all of them has any association with MicroMasters.

src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -0,0 +1,212 @@
...
+        , coalesce(mitxresidential_users.user_full_name, mitxresidential_profiles.user_full_name)
+        as user_full_name
+    from mitxresidential_users
+    left join mitxresidential_profiles on mitxresidential_users.user_id = mitxresidential_profiles.user_id
+)
+
+select
+    coalesce(
+        user_hashed_id
+        , {{ generate_hash_id('cast(user_micromasters_id as varchar) || platform') }} )
```
rachellougee, line 116: Unfortunitely, this won't work as expected. As mentioned above, not evey MITx users has micromaster account. We are hashing all MITx users with their micromasters user_id?

rachellougee (commented):
Sorry, I didn't noticed that `coalesce`, so we end up adding the legacy MicroMasters users to this table. The dbt runs fine, the only thing is we're relying on `int__mitx__users` for this dimentional user. I thought our intention was to replace intermediate layer.  May be its fine.

rachellougee (approved):

src/ol_dbt/models/dimensional/dim_user.sql
rachellougee, line 116: Can you please confirm this line? We've already hashed the legacy Micromasters user here https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitx/int__mitx__users.sql#L201.

src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -0,0 +1,212 @@
...
+    , user_job_title as job_title
+    , user_industry as industry
+    , user_is_active
+    , user_joined_on
+from bootcamps_user_view
+
+union all
+
+select
+    {{ generate_hash_id('cast(user_id as varchar) || platform') }} as user_id
```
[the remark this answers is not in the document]
blarghmatey: Also, generating a user_id multiple times seems like it will cause a lot of non-determinism in this table. There also doesn't appear to be any business rules about which platform source wins out for setting the user information (e.g. full name, email), which could lead to flakiness in the value.

blarghmatey (changes requested):

pdpinch:
we're handling this elsewhere

PR #1526, opened 2025-04-07
add new user dim - mitxonline specific

KatelynGit:
https://github.com/mitodl/ol-data-platform/issues/1325

creates a user dimensional table (dim_user)

### How can this be tested?
dbt build --select dim_user


src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -0,0 +1,52 @@
...
+        , mitxonline_users.user_id
+        , mitxonline_users.user_joined_on
+    from mitxonline_users
+    left join mitxonline_legaladdress on mitxonline_users.user_id = mitxonline_legaladdress.user_id
+    left join mitxonline_profile on mitxonline_users.user_id = mitxonline_profile.user_id
+)
+
+select
+{{ dbt_utils.generate_surrogate_key(['user_email']) }} as user_pk
+    , user_id as user_mitxonline_id
```
blarghmatey, line 39: This will also require bringing in the `auth_users` info from the MITx Online edX application and including it as the `mitxonline_openedx_user_id`. Along with that, this field should probably be named `mitxonline_application_user_id` to provide disambiguation between the various user IDs.

Issue #1515, opened 2025-03-21
Design and implement dbt models for learner course engagement


rachellougee:
### Description/Context

In order to create a [individual learner dashboard  ](https://docs.openedx.org/projects/openedx-aspects/en/latest/reference/individual_learner_dashboard.html) within our data platform, we first need to design the course engagement dbt models.

The dbt models should cover the following key components:

- Page view / Navigation
   - It should capture the learner's page views at the section and subsection levels.
- Problem Submission
  -  It should capture the learner's problem attempts, including the number of attempts, the success or failure of each attempt, and the grade for the problem
- Video engagement 
  - It should track video engagement, including whether the learner watched the entire video or only a portion of it.
  
We could take inspiration from the existing [aspect-dbt ](https://github.com/openedx/aspects-dbt/tree/main/models), which can help guide the design of our new models.

### Plan/Design

- [x] Design the dbt models to cover page views, problem submissions, and video engagements listed above
    - ERD https://app.erdlab.io/designer/schema/1742844268-engagement_models
- [x] Implement the fact event and dim models 
   - tfact_course_navigation_events
   - tfact_video_events
   - tfact_problem_events
   - dim_course_content
   - dim_problem
   - dim_video
   - dim_user
- [x] Implement the aggregated fact event models: afact_course_page_engagement, afact_problem_engagement, afact_video_engagement

pdpinch:
We should be prepared to add columns to the dim_content model to accomodate problem-specific and video-specific metadata. dimensional models can be sparse and have nulls for fields that aren't applicable.

pdpinch:
We need a `dim_user` table. It can be sparse. It should coalesce user records into one (based on email address, and where possible, socialauth). If an application requires additional ids, we can add another column, e.g. "mitxonline-username" "mitxonline-userid" and "mitxonlined-edx-userid"

rachellougee:
These models are on production. I updated the ERD https://app.erdlab.io/designer/schema/1742844268-engagement_models to reflect current design. I am sure we will iteract on these. The next step is to create dashboard, but that's a seprate issue. This issue is all set for the first pass.

PR #1528, opened 2025-04-07
Creating dim_video and dim_problem

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/1515

Updating `dim_course_content` 
Creating `dim_video` and `dim_problem` as foreign key reference to tfact models

### How can this be tested?

dbt build --select dim_video dim_problem

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -6,7 +6,7 @@ models:
   description: Platform information about the technical system that a given event
     is executed within the context of.
   columns:
-  - name: platform_id
+  - name: platform_pk
     description: string, primary key for this table
     tests:
     - not_null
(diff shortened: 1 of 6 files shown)


src/ol_dbt/models/dimensional/_dim__models.yml
```diff
@@ -25,12 +25,76 @@ models:
     tests:
     - not_null
 
+- name: dim_video
+  description: video metadata across the Open edX platforms
+  columns:
+  - name: video_block_pk
+    description: string, primary key for this table, formatted as block-v1:{org}+{course}+{run}type@video+block@{hash
```
blarghmatey, line 32: Is that an ID that's subject to change with modifications to the content? Not sure if we need to use a different PK field to allow for a Type 2 SCD evolution of these dimensions.

src/ol_dbt/models/dimensional/_dim__models.yml
```diff
@@ -25,12 +25,76 @@ models:
...
+  - name: courserun_readable_id
+    description: string, the edX course ID formatted as course-v1:{org}+{course code}+{run_tag}
+    tests:
+    - not_null
+  - name: video_name
+    description: string, name of the video
+  - name: start_date
+    description: timestamp, start date
+  - name: end_date
+    description: timestamp, end date
```
blarghmatey, line 50: Are these for use as SCD Type 2 fields? If so, can you add that information here?

src/ol_dbt/models/dimensional/_dim__models.yml
```diff
@@ -25,12 +25,76 @@ models:
...
+  - name: html5_sources
+    description: array, html5 video sources
+  - name: transcripts
+    description: json, video transcripts
+
+- name: dim_problem
+  description: problem metadata across the Open edX platforms
+  columns:
+  - name: problem_block_pk
+    description: string, primary key for this table, formatted as block-v1:{org}+{course}+{run}type@problem+block@{hash
```
blarghmatey, line 62: Same considerations as above about if this is appropriate as a PK. We should definitely include this as a column, just not sure if it's stable enough as the PK field.

src/ol_dbt/models/dimensional/_dim__models.yml
```diff
@@ -25,12 +25,76 @@ models:
...
+  - name: problem_name
+    description: string, name of the problem
+  - name: markdown
+    description: string, markdown content of the problem
+  - name: max_attempts
+    description: int, maximum number of attempts allowed for this problem
+  - name: start_date
+    description: timestamp, start date of the problem
+  - name: end_date
+    description: timestamp, end date of the problem
```
blarghmatey, line 84: Is this for SCD Type 2 start/end validity, or is this the start/end date of the availability of the problem in the platform?

src/ol_dbt/models/dimensional/_dim__models.yml
rachellougee: This ID doesn't seem to change based on the course content. It's the same ID as in the tracking log. The dim_course_content table is an SCD Type 2 dimension table, and the `content_block_fk` in this table references that table

If we use a different PK to allow for SCD type 2, it would be difficult to associate it with tracking log events because the tracking log uses, which doesn’t specify which version of the course content is being referred to.

src/ol_dbt/models/dimensional/_dim__models.yml
rachellougee: These are the dates associated with the specific video and problem. I didn't find any video or problem where the start_date or end_date differ from their metadata.

src/ol_dbt/models/dimensional/_dim__models.yml
rachellougee: As mentioned, this is start/end date of the availability of the problem. It's not for SCD Type 2 start/end validity.

blarghmatey (approved):

PR #1530, opened 2025-04-10
create video engagement fact table

KatelynGit:
https://github.com/mitodl/ol-data-platform/issues/1515

Added model definitions and dbt build code for:
- afact_video_engagement

### How can this be tested?
```
dbt build --select afact_video_engagement
```

### DIFF
--- a/src/ol_dbt/models/dimensional/afact_video_engagement.sql
+++ b/src/ol_dbt/models/dimensional/afact_video_engagement.sql
@@ -7,7 +7,7 @@ with course_content as (
 )
 
 , mitxonline_video_events as (
-    select 
+    select
         a.user_username
         , a.openedx_user_id
         , a.courserun_readable_id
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/dimensional/afact_video_engagement.sql
```diff
@@ -0,0 +1,102 @@
...
+        , d.block_title as section_title
+        , d.content_block_pk as section_content_fk
+        , json_query(a.useractivity_event_object, 'lax $.id' omit quotes) as video_id
+        , json_query(a.useractivity_event_object, 'lax $.currentTime' omit quotes) as currenttime
+        , lag(json_query(a.useractivity_event_object, 'lax $.currentTime' omit quotes), 1) 
+            over (
+                partition by a.user_username, a.courserun_readable_id, json_query(a.useractivity_event_object, 'lax $.id' omit quotes)  
+                order by a.useractivity_timestamp desc
+        ) 
+            as nextcurrenttime
```
rachellougee, line 29: Just a quck question -  is there a reasn we don't use `tfact_video_events`, but instead 
 use `stg__mitxonline__openedx__tracking_logs__user_activity `, even though these columns are already handled there?

rachellougee: yeah, so neither `tfact_video_events` or `stg__mitxonline__openedx__tracking_logs__user_activity` captures the time played, which is why we created this table to calculate it. I think all the fields we need should already in `tfact_video_events`. but I don't want to pitpick on this so I think its ok for now

rachellougee:
dbt runs fine. As discussed, we just need to adjust the logic for `estimated_time_played` to ensure it's calculated close enough

rachellougee (approved):

PR #1542, opened 2025-04-30
Create discussion tfact

rachellougee:

https://github.com/mitodl/hq/issues/7199

Creating tfact_discussion_events

### How can this be tested?

dbt build --select tfact_discussion_events


rachellougee:
> Build ran clean, sql looks good, one thing I'm not sure about is including json in a raw format in a fact table it's not ideal but probably doesn't hurt anything approving the changes

@KatelynGit Thanks for the review. I agree that including the raw json isn't ideal, but I included it to ensure we have complete data in this table, as there are some less common fields like `undo_vote`, `mark_type`. This way, we won't need to backfill them later if there is a use case for these fields.

Also, I added a post_body field and tested. No need to review - just FYI.

PR #1549, opened 2025-05-08
add product readable id to all coupons mart

KatelynGit:
https://github.com/mitodl/hq/issues/6997

adds the product readable id field to all coupons mart table

### How can this be tested?
dbt build --select marts__mitxpro_all_coupons

### DIFF
--- a/src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
+++ b/src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
@@ -127,10 +127,12 @@ models:
     description: int, count of coupons used by coupon payment name
   - name: product_readable_id
     description: string, the product identifier either referring to courserun_readable_id,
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/mitxpro/_marts__mitxpro__models.yml
```diff
@@ -125,9 +125,14 @@ models:
     description: string, contract number used to identify the order
   - name: coupons_used_count
     description: int, count of coupons used by coupon payment name
+  - name: product_readable_id
+    description: string, the product identifier either referring to courserun_readable_id,
+      programrun_readable_id, or program_readable_id. May be null for bootcamps and
+      edX.org.
```
rachellougee, line 131: "May be null for bootcamps and edX.org"  - This doesn't seem to apply here. We probably should add short description for null value.

rachellougee (approved):

Issue #1533, opened 2025-04-11
Improve Dimensional Models - Enforce PK/FK Constraints and Resolve Duplicate Users


rachellougee:
### Description/Context

As part of our recent work in https://github.com/mitodl/ol-data-platform/issues/1515, we created the following tables:

tfact_course_navigation_events
tfact_video_events
tfact_problem_events
afact_course_page_engagement 
afact_problem_engagement
afact_video_engagement
dim_course_content
dim_problem
dim_video
dim_user

The current design is https://app.erdlab.io/designer/schema/1742844268-engagement_models

Problems:
- Missing PK/FK Constraints: Some fact tables are missing foreign key `fk` columns, which can lead to confusion in joins
- Duplicate Users in dim_user: A few users exist in dim_user with the same mitxonline_openedx_user_id, leading to ambiguity when joining in the tfact tables. Note https://github.com/mitodl/ol-data-platform/pull/1529#issuecomment-2792519317. Some of these may be retired users, but  need to confirm.

### Plan/Design

-  Investigate and resolve duplicate users in dim_user with the same `mitxonline_openedx_user_id`
-  Add the appropriate foreign key (_fk) columns in fact tables. Do not remove any existing columns without confirming with Kate, as they might already in use in the dashboard.

quazi-h:
I have added user_fk and platform_fk to the following tables:
```
tfact_course_navigation_events
tfact_video_events
afact_course_page_engagement
afact_problem_engagement
afact_video_engagement
```

I believe this should take care of all the missing fks unless there were any new tables created since this ticket where we didn't include them.

PR #1551, opened 2025-05-12
Add xPro, Residential, and edXorg users to dim_user

quazi-h:

Adding logic to populate MITxPro, Residental MITx, and edX.org users in the dim_user model.

### How can this be tested?
```
dbt build --select dim_user
```

### DIFF
--- a/src/ol_dbt/models/dimensional/dim_user.sql
+++ b/src/ol_dbt/models/dimensional/dim_user.sql
@@ -1,12 +1,13 @@
 -- MITx users from MITx Online and edX with dedup
 -- For users exist on both MITx Online and edX.org
-with mitxonline_user_view as (
+with mitx_users as (
     select
         user_mitxonline_id as mitxonline_application_user_id
         , user_mitxonline_username
         , user_edxorg_id as edxorg_openedx_user_id
         , user_edxorg_username
-        , user_mitxonline_email as email
+        , user_mitxonline_email
+        , user_edxorg_email
         , user_full_name as full_name
         , user_address_country as address_country
         , user_highest_education as highest_education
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -38,25 +58,317 @@ with mitxonline_users as (
...
+        , coalesce(user_info.user_highest_education, person_course.user_highest_education) as user_highest_education
+        , coalesce(user_info.user_gender, person_course.user_gender) as user_gender
+        , coalesce(user_info.user_birth_year, person_course.user_birth_year) as user_birth_year
+        , coalesce(
+            coalesce(user_info.courserunenrollment_is_active = 1, false)
+            , person_course.courserunenrollment_is_active
+            , s3_user.user_is_active
+        ) as user_is_active
+        , coalesce(user_info.courserun_platform, person_course.courserun_platform) as platform
+        , coalesce(user_info.user_id, person_course.user_id, s3_user.user_id) as user_id
```
quazi-h, line 215: Would this be the correct way to get the edxorg openedx user id?
stg__edxorg__bigquery__mitx_user_info_combo.user_id
stg__edxorg__bigquery__mitx_person_course.user_id
stg__edxorg__s3__user.user_id

src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -38,25 +58,317 @@ with mitxonline_users as (
...
+        mitxresidential_openedx_users.user_username
+        , mitxresidential_openedx_users.user_email
+        , mitxresidential_openedx_users.user_full_name
+        , mitxresidential_profile.user_address_country
+        , mitxresidential_profile.user_highest_education
+        , mitxresidential_profile.user_gender
+        , mitxresidential_profile.user_birth_year
+        , mitxresidential_openedx_users.user_is_active
+        , 'residential' as platform
+        , mitxresidential_openedx_users.user_id
```
quazi-h, line 154: Is stg__mitxresidential__openedx__auth_user.user_id the correct source for the residential openedx user_id?

quazi-h:
Counts from testing against dev_production build:
Total Count 6,897,376
mitxonline_openedx_user_id IS NOT NULL; --200043
mitxpro_openedx_user_id IS NOT NULL; --149583
residential_openedx_user_id IS NOT NULL; --40199
edxorg_openedx_user_id IS NOT NULL; --6719598

src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -38,25 +58,317 @@ with mitxonline_users as (
...
+        , cast(null as varchar) as job_title
+        , cast(null as varchar) as industry
+        , edxorg_user_view.user_is_active
+        , edxorg_user_view.user_joined_on
+    from edxorg_user_view
+)
+
+, ranked_users as (
+    select
+        user_pk
```
quazi-h, line 342: Need feedback on this attempt at getting users from all the platforms. I need a few examples of users who are on multiple platforms to test that they are resolving to the same user in this table.

From my counts, there are 47 users who have an openedx_user_id for all 4 platforms, 3121 users that have both an mitxonline and xpro id. How would I go about getting those actual counts from the source tables?

src/ol_dbt/models/dimensional/dim_user.sql
rachellougee: yes, stg__mitxresidential__openedx__auth_user is the only source for residential user.

src/ol_dbt/models/dimensional/dim_user.sql
rachellougee: This query explode due to many-to-many joins. Please take a look at how [int__edxorg__mitx_users](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_users.sql) is done. It should be similar.  `stg__edxorg__s3__user` is our version of users table, similar to the one in BigQuery, but I don't think its ready to be used yet.

src/ol_dbt/models/dimensional/dim_user.sql
rachellougee: Not sure if I understand what you are asking, but to find out users across multiple platform, it should be something like after its combined (before this query):

select count(*),user_email
from combined_users
group by user_email
having count(*) >1

src/ol_dbt/models/dimensional/dim_user.sql
```diff
@@ -38,25 +58,317 @@ with mitxonline_users as (
...
+        , cast(null as varchar) as company
+        , cast(null as varchar) as job_title
+        , cast(null as varchar) as industry
+        , edxorg_user_view.user_is_active
+        , edxorg_user_view.user_joined_on
+    from edxorg_user_view
+)
+
+, ranked_users as (
+    select
```
rachellougee, line 341: This query is a bit odd because of the use of max() in combination of row_number().  Since we are using row_number() to filter for the most recent row, then the max is not needed and it might return the incorrect values.

Also, I am not entirely sure which data source we should use in case of a conflict. But I think the most `user_joined_on` makes sense, though I am not certain.

rachellougee (commented):
I left comments. Also, I am unable to run dim_user due to the query resulting in excessive records from many-to-many joins, which I have commented below.

src/ol_dbt/models/dimensional/dim_user.sql
quazi-h: Okay, I made changes to filter down to the most recent user using those two tables. Should I keep the coalesce logic using both the info_combo and person_course tables or just rely on the data in stg__edxorg__bigquery__mitx_user_info_combo?

src/ol_dbt/models/dimensional/dim_user.sql
quazi-h: I've removed the max() logic.

quazi-h:
> I've made some changes [8051ca7](https://github.com/mitodl/ol-data-platform/commit/8051ca78b6037741684f14bdc27cadf2c9106a34) to optimize the query. The slowness is due to multiple factors, but the most obvious one is the OR condition in the JOIN clause. I've ran the new build, and it only took 53 sec. I've also spot-checked some user records to ensure data are populated as expected
> 
> @quazi-h can you run it to confirm?

I was able to run it in 35.27 seconds, thanks for the optimizations!

PR #1537, opened 2025-04-17
Adding user_fk and platform_fk columns to tfact and afact tables

quazi-h:
### What are the relevant tickets?
https://github.com/mitodl/ol-data-platform/issues/1533

### Description (What does it do?)
A number of the tfact and fact tables are missing the user_fk and platform_fk foreign keys to join back to dim_user and dim_platform.
- Missing foreign keys to ensure user_fk and platform_fk have been included in the correct tfact and afact models.
- The additional openedx_user_id, application_user_id, and username columns for the additional platforms (xpro, residential, and edxorg) have been stubbed out in dim_user. Changes to the dim_user code to populate those fields will come in a separate PR.
- Attempted to optimize some of the longer running builds by limiting the subset of columns being selected by the intermediate tables.

**Affected tables:**
dim_user
tfact_course_navigation_events
tfact_problem_events
tfact_video_events
afact_course_page_engagement
afact_problem_engagement
afact_video_engagement

### How can this be tested?
```
dbt build --select dim_user
dbt build --select tfact_course_navigation_events
dbt build --select tfact_problem_events
dbt build --select tfact_video_events
dbt build --select afact_course_page_engagement
dbt build --select afact_problem_engagement
dbt build --select afact_video_engagement
```

PR #1555, opened 2025-05-19
Add _fk to tfact_problem_events and tfact_discussion_events

rachellougee:

adding user_fk and platform_fk to tfact_problem_events and tfact_discussion_events

### How can this be tested?

Run `dbt build --select dim_platform tfact_problem_events tfact_discussion_events` with no errors

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -52,6 +52,12 @@ models:
     tests:
     - not_null
     - unique
+  - name: platform_readable_id
+    description: string, readable id of the platform. e.g. mitxonline, mitxpro, edxorg,
+      residential.
+    tests:
+    - not_null
+    - unique
   - name: platform_name
     description: string, name of the platform
     tests:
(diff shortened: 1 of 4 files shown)


rachellougee:
@quazi-h Can you take a look at this? 

The out-of-memory issue you ran into in your PR https://github.com/mitodl/ol-data-platform/pull/1537 is due to the INNER JOIN dim_user, which consumes a lot of memory because both sides of the join need to be fully loaded and processed. Using a LEFT JOIN can reduce memory usage, as Trino doesn't need to filter out unmatched rows like it would in an inner join. 

For these null user_fk records, we can just leave them as they are.

quazi-h (approved):
Changes look good. I was able to build the models with no issues and confirmed the user_pk columns in tfact_discussion_events and tfact_problem_events. I was thinking that maybe we should rename the "platform" column in the tfact tables to be "platform_readable_id" to be consistent with dim_platform but I understand that might break any existing dashboards at the moment.

rachellougee:
> I was thinking that maybe we should rename the "platform" column in the tfact tables to be "platform_readable_id" to be consistent with dim_platform but I understand that might break any existing dashboards at the moment.

Yes, if we want to rename platform to platform_readable_id, it should apply to all the tfact and afact models, not just these two. It’s better to wait until we've migrated all the dashboard queries into dbt, so we have full visibility into what's being used in the dashboards. cc @KatelynGit

PR #1557, opened 2025-05-21
add program completion days field to program enrollment mart

KatelynGit:
https://github.com/mitodl/hq/issues/7321

adds the program_completion_days, unique_courses_taken_in_program,  capstone_indicator  fields to the marts__combined_program_enrollment_detail table

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
@@ -297,12 +297,12 @@ with mitxpro__programenrollments as (
     left join courses_in_programs
         on combined_programs.program_readable_id = courses_in_programs.program_readable_id
     left join combined_enrollments
-        on 
+        on
             cast(combined_programs.user_id as varchar) = combined_enrollments.user_id
             and courses_in_programs.course_readable_id = combined_enrollments.course_readable_id
     left join combined_courseruns
         on combined_enrollments.courserun_readable_id = combined_courseruns.courserun_readable_id
-    group by 
+    group by
         combined_programs.platform_name
         , combined_programs.program_id
         , combined_programs.program_title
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/marts/combined/marts__combined_program_enrollment_detail.sql
```diff
@@ -193,27 +217,144 @@ with mitxpro__programenrollments as (
...
+        as capstone_sum
+        , min(cast(substring(combined_courseruns.courserun_start_on, 1, 10) as date))
+        as first_courserun_start_on_date
+        , count(distinct combined_enrollments.course_readable_id) as unique_courses_taken_in_program
+    from combined_programs
+    left join courses_in_programs
+        on combined_programs.program_readable_id = courses_in_programs.program_readable_id
+    left join combined_enrollments
+        on
+            cast(combined_programs.user_id as varchar) = combined_enrollments.user_id
```
quazi-h, line 301:
```suggestion
            combined_programs.user_id = combined_enrollments.user_id
```
This line was throwing an error when I tried to build the model.
`Database Error in model marts__combined_program_enrollment_detail (models/marts/combined/marts__combined_program_enrollment_detail.sql)
  TrinoUserError(type=USER_ERROR, name=TYPE_MISMATCH, message="line 334:56: Cannot apply operator: varchar = bigint", query_id=20250522_214547_26950_suwmf)
  compiled Code at target/run/open_learning/models/marts/combined/marts__combined_program_enrollment_detail.sql`
  I checked on Starburst and I believe the user_id fields from both tables (int__mitxpro__programenrollments and int__combined__courserun_enrollments) are bigint, so you shouldn't need to cast it here. Unless my dev schema where I am testing is off from what's in production.

quazi-h (approved):
Code logic looks good, just had one issue during my build, left a comment with a suggestion. Did you happen to ever see that same error when you were testing?

PR #1558, opened 2025-05-22
Automate publishing dbt models to Superset via Dagster sensor and job

rachellougee:

https://github.com/mitodl/hq/issues/7268

Adds a Dagster sensor and job to detect Superset datasets in the `mart` and `reporting` layers and create them in [Superset](https://bi.ol.mit.edu/tablemodelview/list/?pageIndex=0&sortColumn=changed_on_delta_humanized&sortOrder=desc) automatically. Can be extended to other models if needed.

### How can this be tested?

Since we don't have QA instance of Superset, I tested against our production to ensure it works. This is how I tested it:

- export GITHUB_TOKEN=
- docker compose up
- Go to http://0.0.0.0:3000/locations/ol_orchestrate.definitions.lakehouse.elt/sensors/sync_superset_datasets_from_dbt
- Click on "Preview tick result" and then continue
- These models are automatically created to  https://bi.ol.mit.edu/tablemodelview/list/?pageIndex=0&sortColumn=changed_on_delta_humanized&sortOrder=desc
```
ops:
  create_superset_datasets:
    config:
      models:
      - schema_suffix: mart
        table: marts__micromasters_dedp_exam_grades
      - schema_suffix: mart
        table: marts__micromasters_summary
      - schema_suffix: mart
        table: marts__micromasters_summary_timeseries
      - schema_suffix: mart
        table: marts__mitxonline_user_profiles
      - schema_suffix: mart
        table: marts__mitxonline_video_engagements
      - schema_suffix: mart
        table: marts__ocw_courses
```

You would need to delete the datasets I created for testing purposes.

### DIFF
--- a/src/ol_orchestrate/jobs/superset_api.py
+++ b/src/ol_orchestrate/jobs/superset_api.py
@@ -1,0 +1,48 @@
+import httpx
+from dagster import Array, OpExecutionContext, Shape, job, op
+
+
+@op(
+    name="create_superset_datasets",
+    config_schema={
+        "models": Array(
+            Shape(
+                {
+                    "schema_suffix": str,
+                    "table": str,
+                }
+            )
+        )
+    },
+    required_resource_keys={"vault", "superset_api"},
+)
+def create_superset_datasets(context: OpExecutionContext):
+    context.log.info("Publishing Superset create_superset_datasets")
+    models = context.op_config["models"]
+    for model in models:
+        schema_suffix = model["schema_suffix"]
+        table = model["table"]
+        try:
+            context.log.info("Creating Superset dataset: %s.%s", schema_suffix, table)
+            context.resources.superset_api.client.create_dataset(
+                schema_suffix=schema_suffix,
+                table_name=table,
+            )
+            context.log.info(
+                "Successfully created dataset: %s.%s", schema_suffix, table
+            )
+
+        except httpx.HTTPStatusError:
+            context.log.exception(
+                "HTTPStatusError while creating dataset for %s.%s",
+                schema_suffix,
+                table,
+            )
+
+
+@job(
+    name="superset_sync_job",
+    description="Syncs dbt models to Superset as datasets",
+)
+def superset_sync_job():
+    create_superset_datasets()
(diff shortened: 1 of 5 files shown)


src/ol_orchestrate/jobs/superset_api.py
```diff
@@ -0,0 +1,47 @@
...
+            Shape(
+                {
+                    "schema_suffix": str,
+                    "table": str,
+                }
+            )
+        )
+    },
+    required_resource_keys={"vault", "superset_api"},
+)
```
blarghmatey, line 18: Can we turn this into an asset? That way we can wire in the upstream dbt assets as well as showing the Superset datasets in the asset graph lineage view.

blarghmatey (commented):
This is a great step forward. If we can turn the `op` into an `asset` that should allow us to remove the sensor logic in place of making the datasets part of the asset lineage graph.

src/ol_orchestrate/jobs/superset_api.py
rachellougee: Yes, I can turn this into an asset

src/ol_orchestrate/assets/superset.py
```diff
@@ -0,0 +1,52 @@
...
+
+from ol_orchestrate.assets.lakehouse.dbt import full_dbt_project
+from ol_orchestrate.lib.automation_policies import upstream_or_code_changes
+from ol_orchestrate.resources.superset_api import SupersetApiClientFactory
+
+
+def create_superset_asset(dbt_asset_group_name: str, dbt_model_name: str):
+    @asset(
+        key=AssetKey(("superset", "dataset", dbt_model_name)),
+        deps=[get_asset_key_for_model([full_dbt_project], dbt_model_name)],
```
blarghmatey, line 17: It seems like if we set an `io_manager` for the `full_dbt_project` asset then we can actually use the dbt model asset output as an `in`, but it seems like that's not required for this use-case.

rachellougee: For our current use case, we don't need to use the dbt model asset output at this point, but we might need it later if we want to clean up Superset datasets when a dbt model is renamed — since Superset won’t automatically relink to the renamed table, and the old dataset might become invalid.

PR #1559, opened 2025-05-23
add fields to program summary report to allow for additional filtering of the report

KatelynGit:
https://github.mit.edu/MIT-OL-Data-Platform/Data-Platform-Issues/issues/4

adds platform, program_type, program_track, and program_readable_id fields to the program_summary_report

### How can this be tested?
dbt build --select program_summary_report

### DIFF
--- a/src/ol_dbt/models/reporting/_reporting__models.yml
+++ b/src/ol_dbt/models/reporting/_reporting__models.yml
@@ -6,12 +6,24 @@ models:
     (audit and verified), certificates and program credentials aggregated for each
     program.
   columns:
+  - name: platform
+    description: str, name of the platform
   - name: program_name
     description: str, name of the program. If program is split over multiple tracks,
       the program name is the same for all tracks. e.g. Data, Economics, and Design
       of Policy
     tests:
     - not_null
+  - name: program_type
+    description: str,  the type of the program. Possible values are 'MicroMasters',
+      'XSeries' and 'Professional Certificate'
+  - name: program_track
+    description: str, name of the specific track for MicroMasters programs. For example,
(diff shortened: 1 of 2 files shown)


rachellougee (commented):
Sorry, I didn't get to it last week. The dbt run successfully. However, when I check the data, the `program_type`, `program_track` and `program_readable_id` are all NULL for edX.org, which doesn't seem correct.  

The initial design for program_summary_report was to group all the variants into one (see the [original issue](https://github.com/mitodl/hq/issues/6124) - `To keep the summary simple, we will summarize all these variants as one program`). This is why the program_track was not included in the report. @pdpinch is this requirement changed?

However, the current logic appears to be grouping by both `program_readable_id`, `program_type` and `program_track`, which should create separate rows for each variant. That is fine as long as the grouping is intentional, but for the program `Statistics and Data Science`, the summary should not be a single row now because it has 4 variants (4 program tracks). Here is the dashboard for "`Statistics and Data Science`" in https://bi.ol.mit.edu/superset/dashboard/p/ed9qrbjqxan/  for comparison.

@KatelynGit Can you please take a look?

rachellougee:
>  Just to give background- the reason for this request was really driven by Jeremy Rossen and others looking as SDS program data and were trying to use the filters on the left hand side of the dashboard here- https://bi.ol.mit.edu/superset/dashboard/p/vmQq1YwqAdJ/ they noticed the filters weren't working for the summary tab and needed them to answer some of their questions.

Good to know. https://github.com/mitodl/hq/issues/6124#issuecomment-2546322904 - this is where we left when Peter noticed the same thing. So I can understand the confusion.

PR #1563, opened 2025-05-28
Fix intermediate ovs video model error

rachellougee:

https://pipelines.odl.mit.edu/runs/de98f752-abd6-4813-8ffc-224d4da120f4

fixing the dbt error due to the deprecation of `dj_elastictranscoder_encodejob` and new table `ui_encodejob` in OVS app. 

Note that there is a PR https://github.com/mitodl/odl-video-service/pull/1212 to add `duration` field to `ui_video`, which will simplify the intermediate. I think it is still useful to keep stg__ovs__studio__postgres__ui_encodejob for other use cases.

### How can this be tested?

Running `dbt build --select  +int__ovs__videos` successfully, and check `duration` field populated correctly


blarghmatey:
Some questions related to the source and historical data:
- Is the old table completely deleted in the application?
- Is the data from historical videos/transcodes present in the new table?
- If the historical data is in the previous table, how can we manage the transition so that we keep the data in a single dimension table in the intermediate layer?

PR #1564, opened 2025-05-29
add learner engagement reporting table

KatelynGit:

adds the learner_engagement_report table. It will improve the performance of this chart https://bi.ol.mit.edu/superset/explore/p/gJk7gr2GaBD/  after creation and after it's added as a source for that chart.

### How can this be tested?


rachellougee (approved):
dbt runs ok, but the queries seem too complex. We might want to simplify them or moving some of them to the afact model later. For now, it's probably fine.

PR #1569, opened 2025-06-02
add join format fixes for learner engagement dashboard

KatelynGit:
https://github.com/mitodl/hq/issues/7388

course run readable id's in different formats complicate joins this should fix the formatting when joining for the learner engagement report/dashboard

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/reporting/learner_engagement_report.sql
+++ b/src/ol_dbt/models/reporting/learner_engagement_report.sql
@@ -140,13 +140,13 @@ with video_pre_query as (
             c.content_block_fk = v.content_block_pk
             and replace(replace(a.courserun_readable_id, 'course-v1:', ''), '+', '/') = v.courserun_readable_id
     inner join ol_warehouse_production_dimensional.dim_course_content as cc_subsection
-        on 
+        on
             v.parent_block_id = cc_subsection.block_id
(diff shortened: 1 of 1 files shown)


rachellougee (changes requested):
It looks like this change fixed the edxorg data but broke the MITx Online and xPro data  - now there is only _one_ course from MITx Online and xPro

```
select distinct courserun_readable_id 
from ol_warehouse_production_x_reporting.learner_engagement_report
where platform in ('mitxonline', 'mitxpro')
```

As mentioned, the issue we're addressing here applies only to the edxorg tracking log data, since it remains the original course format. You can refer to how we handle it in [int__edxorg__mitx_user_activity](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_user_activity.sql#L14).


 I think the easiest way to fix this is to apply the same logic in tfact_*, so the data flows through consistently. Alternatively, we could convert the course ID in edxorg tracking log staging model, but that might complicate things since it's an incremental model.

rachellougee (approved):
Looks good now. As mentioned, I will make a PR to format edxorg course ID in tfact_ so that the data is consistent with other tables.

PR #1570, opened 2025-06-05
feat: Canvas course export and S3 upload

rachellougee:

https://github.com/mitodl/hq/issues/7459

Implementing a Dagster pipeline to export canvas course content and upload to our designed bucket in S3
- s3://ol-devops-sandbox/ for local testing
- s3://ol-data-lake-landing-zone-production/ for production

For now, we only need to export the course content for https://canvas.mit.edu/courses/7023. We will add more later once Peter give us a list of courses.

### How can this be tested?

```
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
export GITHUB_TOKEN=
dagster dev -d src/ol_orchestrate/ -f src/ol_orchestrate/definitions/canvas_course_export.py
```
Navigate to http://127.0.0.1:3000/assets/canvas?view=folder and click on "Materialize"
Course ID 7023 should be uploaded to `s3://ol-devops-sandbox/pipeline-storage/canvas/course_export/`

### DIFF
--- a/src/ol_orchestrate/assets/canvas.py
+++ b/src/ol_orchestrate/assets/canvas.py
@@ -1,0 +1,91 @@
+import hashlib
+import time
+from pathlib import Path
+
+import requests
+from dagster import (
+    AssetExecutionContext,
+    AssetKey,
+    AssetOut,
+    DataVersion,
+    Output,
+    multi_asset,
+)
+
+# Define a list of course IDs to export.
+course_ids = [7023]
+
+
+@multi_asset(
+    outs={
+        f"course_{course_id}": AssetOut(
+            key=AssetKey(["canvas", f"course_{course_id}"]),
+            io_manager_key="s3file_io_manager",
+        )
+        for course_id in course_ids
+    },
+    group_name="canvas",
+    required_resource_keys={"canvas_api"},
+)
+def export_courses(context: AssetExecutionContext):
+    for course_id in course_ids:
+        course = context.resources.canvas_api.client.get_course(course_id)
+
+        export_course_response = (
+            context.resources.canvas_api.client.export_course_content(course_id)
+        )
+        context.log.info(
+            "Initiated export of course ID %s: %s", course_id, export_course_response
+        )
+        export_id = export_course_response["id"]
+
+        # Poll until the export is ready
+        while True:
+            export_status = (
+                context.resources.canvas_api.client.check_course_export_status(
+                    course_id, export_id
+                )
+            )
+            if export_status["workflow_state"] == "exported":
+                attachment = export_status.get("attachment", {})
+                download_url = attachment.get("url")
+                context.log.info(
+                    "Export started for course ID %s at %s", course_id, download_url
+                )
+                break
+            elif export_status["workflow_state"] in ["failed", "error"]:
+                message = f"Export failed for course {course_id}"
+                context.log.error(message)
+                raise Exception(message)  # noqa: TRY002
+            else:
+                context.log.info(
+                    "Waiting for course export (state: %s)",
+                    export_status["workflow_state"],
+                )
+                time.sleep(30)
+
+        course_export_path = Path(f"{course_id}_course_export.imscc")
+        response = requests.get(download_url, timeout=20)
+        response.raise_for_status()
+
+        course_export_path.write_bytes(response.content)
+
+        with course_export_path.open("rb") as f:
+            data_version = hashlib.file_digest(f, "sha256").hexdigest()
+
+        target_path = f"canvas/course_export/{course_id}/{data_version}.imscc"
+
+        context.log.info("Downloading %s file to %s", download_url, target_path)
+
+        yield Output(
+            value=(course_export_path, target_path),
+            output_name=f"course_{course_id}",
+            data_version=DataVersion(data_version),
+            metadata={
+                "course_id": course_id,
+                "course_name": course["name"],
+                "course_code": course["course_code"],
+                "course_readable_id": course["sis_course_id"],
+                "object_key": target_path,
+            },
+        )
--- a/src/ol_orchestrate/assets/canvas.py
+++ b/src/ol_orchestrate/assets/canvas.py
@@ -1,98 +1,94 @@
 import hashlib
 import time
+from datetime import UTC, datetime
 from pathlib import Path
 
-import requests
 from dagster import (
     AssetExecutionContext,
     AssetKey,
-    AssetOut,
     DataVersion,
     Output,
-    multi_asset,
+    StaticPartitionsDefinition,
+    asset,
 )
 
-# Define a list of course IDs to export.
-course_ids = [7023]
+# predefined course IDs to export
+course_partitions = StaticPartitionsDefinition(["7023"])
 
 
-@multi_asset(
-    outs={
-        f"course_{course_id}": AssetOut(
-            key=AssetKey(["canvas", f"course_{course_id}"]),
-            io_manager_key="s3file_io_manager",
-        )
-        for course_id in course_ids
-    },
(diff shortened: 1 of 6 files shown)


src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,98 @@
...
+course_ids = [7023]
+
+
+@multi_asset(
+    outs={
+        f"course_{course_id}": AssetOut(
+            key=AssetKey(["canvas", f"course_{course_id}"]),
+            io_manager_key="s3file_io_manager",
+        )
+        for course_id in course_ids
```
blarghmatey, line 25: The asset itself is a singleton object. The course IDs are keys in a partition object along with the course asset is segmented. The mental model for an asset is similar to that of a class, where it is the abstract representation of an object. The partitions are instances of the object.

src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,98 @@
...
+    group_name="canvas",
+    required_resource_keys={"canvas_api"},
+)
+def export_courses(context: AssetExecutionContext):
+    # only export common cartridge for now
+    export_type = "common_cartridge"
+
+    extension = "imscc" if export_type == "common_cartridge" else export_type
+
+    for course_id in course_ids:
```
blarghmatey, line 36: Each execution of this function should only operate against a single course, so this should be something like `course_id = context.partition_key`

src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,98 @@
...
+                raise Exception(message)  # noqa: TRY002
+            else:
+                context.log.info(
+                    "Waiting for course export (state: %s)",
+                    export_status["workflow_state"],
+                )
+                time.sleep(30)
+
+        course_export_path = Path(f"{course_id}_course_export.{extension}")
+        response = requests.get(download_url, timeout=20)
```
blarghmatey, line 75: We've largely been using httpx in this repo, so we should stick with that for consistency.

src/ol_orchestrate/definitions/canvas_course_export.py
```diff
@@ -0,0 +1,37 @@
...
+    name="canvas_course_export_schedule",
+    target=AssetSelection.assets(export_courses),
+    cron_schedule="@daily",
+    execution_timezone="Etc/UTC",
+)
+
+canvas_course_export = Definitions(
+    resources={
+        "io_manager": default_io_manager(DAGSTER_ENV),
+        "s3file_io_manager": S3FileObjectIOManager(
```
blarghmatey, line 27: This should use the helper method for the IO Manager to make it easier to use across local and deployed environments https://github.com/mitodl/ol-data-platform/blob/integrate-dbt-with-superset/src/ol_orchestrate/lib/dagster_helpers.py#L33

quazi-h (commented):
I was able to get the pipeline deployed and was able to start an asset materialization job for a canvas course export but the job eventually failed with the following error:
```botocore.exceptions.ReadTimeoutError: Read timeout on endpoint URL: "https://ol-devops-sandbox.s3.amazonaws.com/pipeline-storage/canvas/course_export/7023/ad9cf912abadef30deb0165201951c80e9015f3fbeb6b5fc5eeff12cb032b1bc.imscc?partNumber=1&uploadId=B4O8y.S3xBIu6G3YbKBsmuAkSSTd7tJVi2BYDtTQFnuClIs3gu8OQb9ufOAf6L9GfdAzGeYGUiYll6shSJPsvS3WeAkFasWzZ8BFYbuZDkGW.OdrRrl_EikjTRiHaR06"```
I did try creating a new set of IAM credentials to test with but got the same result, I am not sure why it would time out because Rachel confirmed the jobs ran and completed successfully for her this morning. I wonder if I am setting up the access key incorrectly, I selected the "Local Code" option ("Enable application code in a local development environment to access your AWS account") but maybe it requires the "Third-party service" or "Application running outside AWS" option?

File: src/ol_orchestrate/assets/canvas.py [hunk 5]
```diff
@@ -0,0 +1,94 @@
...
+    AssetExecutionContext,
+    AssetKey,
+    DataVersion,
+    Output,
+    StaticPartitionsDefinition,
+    asset,
+)
+
+# predefined course IDs to export
+course_partitions = StaticPartitionsDefinition(["7023"])
```
blarghmatey, line 16:
```suggestion
canvas_course_ids = StaticPartitionsDefinition(["7023"])
```

src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@asset(
+    key=AssetKey(["canvas", "course_export"]),
+    group_name="canvas",
+    required_resource_keys={"canvas_api"},
+    io_manager_key="s3file_io_manager",
+    partitions_def=course_partitions,
```
blarghmatey, line 24:
```suggestion
    partitions_def=canvas_course_ids,
```

src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,94 @@
...
+)
+def export_courses(context: AssetExecutionContext):
+    course_id = int(context.partition_key)
+    export_date = datetime.now(tz=UTC).strftime("%Y%m%d")
+    # only export common cartridge for now
+    export_type = "common_cartridge"
```
blarghmatey, line 30: [nit] this might be good to create as a constant

src/ol_orchestrate/resources/canvas_api.py
```diff
@@ -0,0 +1,135 @@
...
+            f"{self.base_url}/api/v1/courses/{course_id}/content_exports/{export_id}"
+        )
+        response = self.http_client.get(
+            request_url,
+            headers={"Authorization": f"{self.token_type} {self.access_token}"},
+        )
+        response.raise_for_status()
+        return response.json()
+
+    def download_course_export(self, url: str, output_path: Path) -> None:
```
blarghmatey, line 93: [nit] To make it a bit more of a ["fluent API"](https://en.wikipedia.org/wiki/Fluent_interface) it might be useful to return the Path object of the written file.

src/ol_orchestrate/assets/canvas.py
```diff
@@ -0,0 +1,94 @@
...
+        context.log.error(message)
+        raise Exception(message)  # noqa: TRY002
+
+    course_export_path = Path(f"{course_id}_course_export.{extension}")
+    context.resources.canvas_api.client.download_course_export(
+        download_url, course_export_path
+    )
+
+    with course_export_path.open("rb") as f:
+        data_version = hashlib.file_digest(f, "sha256").hexdigest()
```
blarghmatey, line 78: To get a proper data version, we would want to do a recursive hash of the contents, excluding the metadata.xml file to avoid erroneous hash changes due to the export timestamp. Something like:

```python
import hashlib
import zipfile
import os
hasher = hashlib.new("sha256")

with zipfile.ZipFile(course_export_path, 'r') as zip_file:
    for item in zip_file.infolist():
        hasher.update(item.filename.encode('utf-8'))
        if not item.is_dir() and not item.filename == "metadata.xml":
            with zip_file.open(item.filename) as file_content:
                while True:
                    chunk = file_content.read(4096)
                    if not chunk:
                        break
                    hasher.update(chunk)

data_version = hasher.hexdigest()
```

It's probably worth building this into a lib module since we'll likely do this for other types of data assets.

rachellougee:
@blarghmatey Thanks. I think I’ve addressed all your feedback and ran some tests — the data version seems to be working. Can you please take another look?

PR #1573, opened 2025-06-09
Fix dbt test errors related to course_number for UAI courses

rachellougee:

N/A

There are a few dbt test errors like https://pipelines.odl.mit.edu/runs/124511d4-c301-4e18-81c7-48be2187392d due to course_number is not unique for UAI courses. e.g. course-v1:UAI_SOURCE+UAI.2 and course-v1:UAI_HHC+UAI.2

This PR is to fix the join on `int__combined__course_runs` to use course_readable_id instead of course_numbe, also remove the unique check on course_number since its not unique anymore.

### How can this be tested?

No error running

dbt build --select +int__combined__course_runs

### DIFF
--- a/src/ol_dbt/models/intermediate/combined/int__combined__course_runs.sql
+++ b/src/ol_dbt/models/intermediate/combined/int__combined__course_runs.sql
@@ -70,7 +70,7 @@ with mitx_courses as (
             else false
         end as courserun_is_current
     from mitxonline_runs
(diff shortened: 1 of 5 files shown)


quazi-h (approved):
Changes look good, I was able to build int__combined__course_runs with no test errors and marts__combined_course_enrollment_detail threw a few errors unrelated to the changes in this PR. Approving this, will confirm to see if a full refresh addresses the errors I saw.

PR #1578, opened 2025-06-20
fix learner engagement reporting to adjust for different course run id formats in the fact tables

KatelynGit:
related to adding data for holistic assessment project work https://github.com/mitodl/hq/issues/7388

We [modified](https://github.com/mitodl/ol-data-platform/pull/1571) the format of the course run id's in the tfacts and that flows through into the afacts as well. As a result joins for this reporting table(which is sourced from the fact tables) needed to be adjusted as well.  The course run id's are used in the "on" clauses for some of the joins for this reporting table. This PR adjusts the joins so that the missing course runs show up. 

This PR modifies the learner_engagement_report in order to pull missing course runs including course-v1:UAI_HHC+UAI.1+2T2025_HHC.

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/reporting/learner_engagement_report.sql
+++ b/src/ol_dbt/models/reporting/learner_engagement_report.sql
@@ -29,7 +29,7 @@ with video_pre_query as (
         , a.courserun_readable_id
         , d.block_title as section_title
         , c.block_title as subsection_title
-        , coalesce(h.course_title, h2.course_title) as course_title
+        , h2.course_title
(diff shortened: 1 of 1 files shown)


pdpinch:
Can you make the PR title more descriptive of what's being fixed? The commit message is a bit better, but even that is vague. I'm curious to know why the prior approach had the affect of omitting specific course ids. 

Remember, this is part of our documentation for our future selves. Why did we make this change? What does it do? If it helps, feel free to ask Copilot for help.

PR #1596, opened 2025-07-09
Added platform_fk and user_fk columns where missing

quazi-h:

A number of the tfact and fact tables are missing the user_fk and platform_fk foreign keys to join back to dim_user and dim_platform.
- user_fk and platform_fk have been included in the correct tfact and afact models (course, video, and problem fact tables).
- Column definitions were added for the respective tables in _dim__models.yml
- Used the "arbitrary" function instead of adding the new columns to the group by

**Affected tables:**
tfact_course_navigation_events
tfact_video_events
afact_course_page_engagement
afact_problem_engagement
afact_video_engagement

### How can this be tested?
```
dbt build --select tfact_course_navigation_events
dbt build --select tfact_video_events
dbt build --select afact_course_page_engagement
dbt build --select afact_problem_engagement
dbt build --select afact_video_engagement
```

### DIFF
--- a/src/ol_dbt/models/dimensional/afact_course_page_engagement.sql
+++ b/src/ol_dbt/models/dimensional/afact_course_page_engagement.sql
@@ -17,6 +17,8 @@ with vertical_structure as (
         , openedx_user_id
         , courserun_readable_id
         , block_fk
+        , arbitrary(platform_fk) as platform_fk
+        , arbitrary(user_fk) as user_fk
         , count(*) as num_of_views
         , max(event_timestamp) as last_view_timestamp
     from {{ ref('tfact_course_navigation_events') }}
@@ -26,7 +28,9 @@ with vertical_structure as (
 
 , combined as (
     select
-        page_navigation_events.platform
+        page_navigation_events.platform_fk
+        , page_navigation_events.user_fk
+        , page_navigation_events.platform
         , page_navigation_events.openedx_user_id
         , page_navigation_events.courserun_readable_id
         , page_navigation_events.block_fk
@@ -40,7 +44,9 @@ with vertical_structure as (
 )
 
 select
-    platform
+    platform_fk
+    , user_fk
+    , platform
     , openedx_user_id
     , courserun_readable_id
     , block_fk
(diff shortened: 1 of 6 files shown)


src/ol_dbt/models/dimensional/afact_course_page_engagement.sql
```diff
@@ -17,6 +17,8 @@ with vertical_structure as (
         , openedx_user_id
         , courserun_readable_id
         , block_fk
+        , arbitrary(platform_fk) as platform_fk
```
quazi-h, line 20: Let me know if I should be adding these columns to the group by clauses instead of using "arbitrary". I figured that adding them to the group by was going to be more expensive to the query compared to using an aggregate function.

quazi-h:
> I think we could capture that value by doing the user joins on usernames only instead. For example- mitxresidential_navigation_events.user_username = users.user_residential_username
> 
> instead of using both the id and username?

I think if it helps us get more coverage on users, it's worth a try. Since we are joining the source specific usernames to data tied to that respective source, we should be safe from incorrect joins between users. I'll make the change and see if we still get records missing user_fk.

@rachellougee Any thoughts on proceeding with joining to the user table on just username?

quazi-h:
> That said, I’m okay with switching to match only username as long as we:
> 
> * Apply it consistently across all fact tables.
> * Add `username` to the fact tables, to avoid confusion when user_fk and openedx_user_id don't align.

I ran counts after trying the change and we only got ~200k more user_fk in two of the tables (tfact_course_navigation_events and tfact_video_events) which is a small amount compared to the total number of records. The problem is that this change results in a few conflicts between the id and associated username which Rachel has highlighted above. After discussing with Rachel, I don't think we want to proceed with removing the openedx_user_id from the joins to the user table since the resulting impersonation cases might be confusing for users of this data.

Regardless of how we join to the user table, we will still have blank users. Instead of putting the burden on the user on how to handle those cases and potentially confusing them, we should just let them join to the user tables as needed if they are trying to drill down into records without a matched user. We may want to open another ticket to include the usernames in the tfact tables to help with that.

PR #1614, opened 2025-07-23
add new discount combined mart table

KatelynGit:
https://github.com/mitodl/hq/issues/7653

adds a new discount combined mart table(marts__combined_discounts)

### How can this be tested?
dbt build --select marts__combined_discounts


src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -901,3 +901,69 @@ models:
     description: boolean, indicating if the course run is currently running. True
       if courserun_start_on is in the past and blank courserun_end_on, or courserun_start_on
       is in the past and courserun_end_on is in the future.
+
+- name: marts__combined_discounts
```
pdpinch, line 905: I know we've been in the habit for some time of calling these "combined" marts and I'm okay with that. But I do want to ask if this will still make sense a year or two from now when everything is part of learn.

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -901,3 +901,69 @@ models:
...
+  description: discounts from xpro and mitxonline
+  columns:
+  - name: platform
+    description: str, name of the platform
+    tests:
+    - not_null
+  - name: discount_code
+    description: string, discount code for the redeemed coupon if applicable
```
pdpinch, line 913: example, please? Would be nice to have an example of a marketing discount and an enrollment code

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -901,3 +901,69 @@ models:
     description: boolean, indicating if the course run is currently running. True
       if courserun_start_on is in the past and blank courserun_end_on, or courserun_start_on
       is in the past and courserun_end_on is in the future.
+
+- name: marts__combined_discounts
+  description: discounts from xpro and mitxonline
```
pdpinch, line 906: Not all users may be familar with the "discount" nomenclature. Maybe clarify this by saying that it includes coupons and enrollment codes (assuming that it does include both)

src/ol_dbt/models/marts/combined/_marts__combined__models.yml
```diff
@@ -901,3 +901,69 @@ models:
...
+  - name: discount_expires_on
+    description: timestamp, expiration timestamp for the coupon payment version
+  - name: num_discount_codes
+    description: int, number of discount objects associated with the payment version
+  - name: discount_max_redemptions
+    description: int, maximum number of redemptions for the discount
+  - name: couponpayment_name
+    description: string, human readable name for the coupon payment
+  - name: b2border_contract_number
+    description: string, contract number used to identify the order
```
pdpinch, line 939: please double-check the nomenclature for this with Bryan. Will he understand and recognize this?

src/ol_dbt/models/marts/combined/marts__combined_discounts.sql
```diff
@@ -0,0 +1,106 @@
...
+    , null as b2border_contract_number
+    , null as b2breceipt_reference_number
+    , d.coupons_used_count as discounts_used_count
+    , c.product_readable_id
+    , coalesce(case when d.discount_code is not null then true end, false) as redeemed
+    , case
+        when a.discount_type = 'percent-off' then cast(a.discount_amount as varchar) || '%'
+        when a.discount_type = 'dollars-off' then '$' || cast(a.discount_amount as varchar) || ' off'
+        when a.discount_type = 'fixed-price' then '$' || cast(a.discount_amount as varchar) || ' fixed price'
+    end
```
rachellougee, line 68: This field is already handled in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/staging/mitxonline/stg__mitxonline__app__postgres__ecommerce_discount.sql#L17-L24 and propagated in `marts__combined__orders`. The discount_amount_text here is worded differently than in the marts__combined__orders. Is there a specific reason for the discrepancy?

src/ol_dbt/models/marts/combined/marts__combined_discounts.sql
```diff
@@ -0,0 +1,106 @@
...
+    , null as company_name
+    , null as discount_type
+    , null as b2bcoupon_id
+    , a.discount_id
+    , null as couponpaymentversion_id
+from mitxonline_discount as a
+left join mitxonline_product_joins as c
+    on a.discount_id = c.discount_id
+left join mitxonline_order_sum as d
+    on a.discount_code = d.discount_code
```
rachellougee, line 79: Nitpit: The aliases a, c and d seem random and arbitrary. Since the CTE names are already descriptive, it might be clearer to use that instead of a, c, d

rachellougee (commented):
dbt runs successfully, but I spot-checked the `discounts_used_count`, and it doesn't appear to match the actual usage. For example, the following discount code is a one-time redemption in the MITx Online application, but this new mart shows 2 as discounts_used_count.

```
select discounts_used_count from "ol_data_lake_production"."ol_warehouse_production_rlougee_mart".marts__combined_discounts
where platform ='MITx Online' and discount_code ='CS-0ae65ed0-141d-42ef-86e4-d29056b66aa5'
```

~~Can you take a look?~~
Actually this seems to two duplicated orders https://mitxonline.mit.edu/admin/ecommerce/order/15818/change/ https://mitxonline.mit.edu/admin/ecommerce/order/15819/change/, but the user only redeemed once.

rachellougee:
> I think we'd need to solve this elsewhere as there are 2 distinct transactions being recorded. Any suggests on calculating when the underlying data is incorrect though? maybe we can compensate in some way

I will create an issue in MITx Online for this. The following query should catch these. But one thing I was going to suggest is that it might be helpful to add `redemption_type` to the mart since `discounts_used_count` shouldn't be more than one when redemption_type is one-time. I would think the `discounts_used_count` only make sense if its unlimited. Take it or leave it

```
select user_id, courserun_id, count(*) 
from "ol_data_lake_production"."ol_warehouse_production_intermediate".int__mitxonline__ecommerce_order
where order_state='fulfilled'
group by user_id, courserun_id
having count(*) > 1
```

PR #1623, opened 2025-07-30
Stage zendesk tables

rachellougee:

close https://github.com/mitodl/hq/issues/7938

Creating staging tables 
  - stg__zendesk__ticket
  - stg__zendesk__ticket_comment
  - stg__zendesk__user
  - stg__zendesk__brand
  - stg__zendesk__organization

### How can this be tested?

dbt build --select staging.zendesk

### DIFF
--- a/src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
+++ b/src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
@@ -69,9 +69,6 @@ models:
     description: array, the IDs of users currently following this ticket
   - name: ticket_collaborator_user_ids
     description: array, The IDs of users currently CC'ed on the ticket
-  - name: forum_topic_id
-    description: int, the topic in the Zendesk Web portal this ticket originated from,
-      if any. The Web portal is deprecated
   - name: ticket_form_id
     description: int, the ID of the ticket form to render for the ticket
   - name: organization_id
@@ -102,7 +99,8 @@ models:
   - name: ticket_is_from_messaging_channel
     description: bool, true if the ticket's via type is a messaging channel.
   - name: ticket_custom_fields
-    description: array, custom fields for the ticket
+    description: array, custom fields for the ticket. e.g. {"id":360028099972,"value":"course_information"}
+      indicates this ticket is related to course information.
   - name: ticket_source_channel
     description: str, the channel through which the ticket was created. e.g. "web",
       "email"
--- a/src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
+++ b/src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
@@ -2,8 +2,8 @@
 version: 2
 
 models:
-- name: stg__zendesk__tickets
-  description: Staging model for Zendesk tickets
+- name: stg__zendesk__ticket
+  description: Zendesk tickets
   columns:
   - name: ticket_id
     description: int, primary key for the ticket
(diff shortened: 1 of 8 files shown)


src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
```diff
@@ -0,0 +1,353 @@
...
+    description: int, foreign key to zendesk user assigned to the ticket
+  - name: ticket_email_cc_user_ids
+    description: array, array of IDs agent or end users email CC to the ticket.
+  - name: ticket_follower_user_ids
+    description: array, the IDs of users currently following this ticket
+  - name: ticket_collaborator_user_ids
+    description: array, The IDs of users currently CC'ed on the ticket
+  - name: forum_topic_id
+    description: int, the topic in the Zendesk Web portal this ticket originated from,
+      if any. The Web portal is deprecated
```
pdpinch, line 74: > The Web portal is deprecated

What does this mean? Can it be removed?

src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
```diff
@@ -0,0 +1,353 @@
...
+    description: str, the satisfaction rating score of the ticket. e.g. "good", "bad",
+      "offered".
+  - name: ticket_satisfaction_rating_comment
+    description: str, the satisfaction rating comment of the ticket
+  - name: ticket_satisfaction_rating_reason
+    description: str, the satisfaction rating reason
+  - name: ticket_is_from_messaging_channel
+    description: bool, true if the ticket's via type is a messaging channel.
+  - name: ticket_custom_fields
+    description: array, custom fields for the ticket
```
pdpinch, line 105: Do we have any examples of custom fields? I wonder if there's some useful metadata here that the CS team adds to tickets.

src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
```diff
@@ -0,0 +1,353 @@
...
+    description: int, primary key for the brand
+    tests:
+    - not_null
+    - unique
+  - name: brand_api_url
+    description: str, the API url of the brand
+  - name: brand_url
+    description: str, the url of the brand
+  - name: brand_name
+    description: str, the name of the brand
```
pdpinch, line 200: Can we get a list of possible values, or are there too many?

src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml
rachellougee: This was taken from their documentation https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#json-format. I've removed it as I realized that this field is all null and not useful to include

quazi-h (approved):
Looks good to me. I was able to build the dbt models with no issues or errors. I verified the tables exist with data on Starburst. I ran the changes from the most recent commit as well.

A few of the fields have a lot of null values but since these are just pulling from raw tables, I'm guessing that's just what we have for that data?
For example, the foreign key organization_id is null for most records in stg__zendesk__ticket and stg__zendesk__user.

rachellougee:
> A few of the fields have a lot of null values but since these are just pulling from raw tables, I'm guessing that's just what we have for that data?
For example, the foreign key organization_id is null for most records in stg__zendesk__ticket and stg__zendesk__user.

Yes, most of the tickets don't have the requester's organization_id associated with them. I guess not all requesters tell us which company they work for. You can also verify it by running
```
select count(distinct id) from "ol_data_lake_production"."ol_warehouse_production_raw".raw__thirdparty__zendesk_support__tickets
where organization_id is not null
```

PR #1627, opened 2025-08-04
version: Updating dbt package dependencies

blarghmatey:
N/A

Update dbt package deps


rachellougee (approved):
Looks good to me. I was able to run  `dbt run-operation generate_source` and it looks like the new version of codegen adds `data_type` which was not available in previous version.

PR #1647, opened 2025-08-18
Course Engagement Model Improvements

quazi-h:
https://github.com/mitodl/hq/issues/8184

Added the "course_number" field to int__combined__course_runs so that it doesn't need to be calculated when viewing the dashboards.
Renamed video engagement reports for consistency.

### How can this be tested?
```
dbt build --select +int__combined__course_runs
dbt build --select +marts__combined_course_engagements
dbt build --select +marts__combined_video_engagements
dbt build --select +video_engagement_report
```


src/ol_dbt/models/intermediate/combined/int__combined__course_runs.sql
```diff
@@ -191,6 +200,7 @@ with mitx_courses as (
         , global_alumni_runs.courserun_external_readable_id as course_readable_id
         , global_alumni_runs.courserun_title
         , global_alumni_runs.courserun_external_readable_id
+        , global_alumni_runs.course_number
```
rachellougee, line 203: It looks like this was added to the wrong order

src/ol_dbt/models/intermediate/combined/int__combined__course_runs.sql
```diff
@@ -220,6 +230,10 @@ with mitx_courses as (
         , bootcamps_courses.course_readable_id
         , bootcamps_runs.courserun_title
         , bootcamps_runs.courserun_readable_id
+        , case
+            when cardinality(split(bootcamps_runs.courserun_readable_id, '+')) >= 2
+                then split(bootcamps_runs.courserun_readable_id, '+')[2]
+        end as course_number
```
rachellougee, line 236: It looks like this was added to the wrong order - can you check?

src/ol_dbt/models/intermediate/combined/int__combined__course_runs.sql
```diff
@@ -247,6 +261,10 @@ with mitx_courses as (
         , course_readable_id
         , courserun_title
         , courserun_readable_id
+        , case
+            when cardinality(split(courserun_readable_id, '+')) >= 2
+                then split(courserun_readable_id, '+')[2]
+        end as course_number
```
rachellougee, line 267: this was added to the wrong order - can you check?

rachellougee (changes requested):
The dbt test failed due to the incorrect order. Can you take a look and fix it?

PR #1660, opened 2025-08-27
fix: add course export get status timeout to avoid infinite blocking

Anas12091101:
https://github.com/mitodl/hq/issues/8223

This PR introduces a timeout in the polling loop that calls the course export plugin’s GET API to fetch task status. The timeout ensures the loop does not block indefinitely in cases where a task takes too long to complete.

I wasn’t able to do a proper smoke test for these changes. Could the reviewer please handle the smoke testing as part of the review?

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### DIFF
--- a/src/ol_orchestrate/assets/openedx.py
+++ b/src/ol_orchestrate/assets/openedx.py
@@ -179,8 +179,8 @@ def course_xml(context: AssetExecutionContext, courseware):  # noqa: ARG001
                 datetime.now(tz=UTC) - start_time
                 > COURSE_EXPORT_GET_TASKS_STATUS_TIMEOUT
             ):
-                context.log.warning("Timeout reached before all tasks completed.")
-                raise TimeoutError(f"Course export timed out for {course_key}")
+                err_msg = f"Course export timed out for {course_key}"
+                raise TimeoutError(err_msg)
             time.sleep(timedelta(seconds=20).seconds)
             for course_id, task_id in tasks.items():
                 task_status = (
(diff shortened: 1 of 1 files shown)


src/ol_orchestrate/assets/openedx.py
```diff
@@ -33,6 +33,7 @@
 
 HTTP_SUCCESS = 200
 HTTP_NOT_FOUND = 404
+COURSE_EXPORT_GET_TASKS_STATUS_TIMEOUT = timedelta(minutes=60)
```
Anas12091101, line 36: I’m not fully confident about the timeout value. From reviewing a few events on Sentry, the longest duration I observed was around 18 minutes.

src/ol_orchestrate/assets/openedx.py
```diff
@@ -181,10 +189,17 @@ def course_xml(context: AssetExecutionContext, courseware):  # noqa: ARG001
...
+                if state == "Succeeded":
                     successful_exports.add(course_id)
-                if task_status["state"] in {"Failed", "Canceled", "Retrying"}:
+                elif state in {"Failed", "Canceled", "Retrying"}:
                     failed_exports.add(course_id)
+                elif details:
+                    context.log.info(
+                        "Details of export task for course %s (task %s): %s",
+                        course_id, task_id, details,
+                    )
```
Anas12091101, line 202: This is the expected response if the task has not started yet:
https://github.com/mitodl/open-edx-plugins/pull/592/files#diff-f01ae1a34afe26ef96ce446b7ca6657ed5f2c79bcb6f5c48636937718d22ab99R162

rachellougee (approved):
Looks good to me. I left a small comment but its not a blocker.

src/ol_orchestrate/assets/openedx.py
Anas12091101: Yes, the "details" key is only returned when no task is found. You can find more details in this [comment](https://github.com/mitodl/open-edx-plugins/pull/592#discussion_r2303485348)

PR #1706, opened 2025-09-30
add a new enrollment migration table to help with migration of enrollments from 2U into mitxonline

KatelynGit:
https://github.com/mitodl/hq/issues/8679

Creating edxorg_to_mitxonline_enrollments to be used in mitxonline application for the migration

### How can this be tested?
dbt build --select edxorg_to_mitxonline_enrollments


src/ol_dbt/models/migration/edxorg_to_mitxonline_enrollments.sql
```diff
@@ -0,0 +1,156 @@
...
+    from combined_enrollments
+    where combined_enrollments.platform = '{{ var("mitxonline") }}'
+    group by
+        combined_enrollments.courserun_readable_id
+        , combined_enrollments.user_email
+        , combined_enrollments.course_readable_id
+)
+
+, edxorg_enrollment as (
+    select
```
rachellougee, line 54: This subquery can be simplified a lot by using int__edxorg__mitx_courserun_enrollments and removing unused columns, as there is no need to join any micromasters table especially micromasters_completed_orders. 

Also, the use of `coalesce(mitxonline_certificates.courseruncertificate_created_on` is a bit confusing. Since we are migrating data to mitxonline, the certificate data in this table should not exist on mitxonline

rachellougee (approved):
dbt runs fine. I left a few comments, but its fine to leave it as is for now since I'll need to make some changes to add the certificate page revision later.

PR #1738, opened 2025-10-20
Chatbot tfact event model

rachellougee:

https://github.com/mitodl/hq/issues/7941

Creating tfact_chatbot_events for openedx and canvas chatbot interactions sourced from tracking logs.
 - For the canvas chatbots, older messages are missing the canvas_course_id, but it will be populated going forward

### How can this be tested?

dbt build +tfact_chatbot_events

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -841,6 +841,10 @@ models:
     description: string, name of the content block where the chatbot event occurred.
   - name: chatbot_type
     description: string, type of the chatbot. e.g. Syllabus, Tutor, etc.
+  - name: event_value
+    description: string, the value of the event, e.g. the message submitted by the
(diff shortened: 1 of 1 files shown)


KatelynGit (approved):
The build ran clean, definitions and sql look good, however the data doesn't look right. As an example(the username and course readable id values are not correct): select * from 
ol_warehouse_production_kdelaney_dimensional.tfact_chatbot_events
where openedx_user_id= 489247

I think this may be because of issues unrelated to this PR specifically but they are definitely affecting the new table and we'll need to fix them as part of this or separately

rachellougee:
As we discussed, this specific example `openedx_user_id= 489247` is related to the existing canvas chatbots where we didn't capture the course ID in the tracking log. I noted in the PR description 

`For the canvas chatbots, older messages are missing the canvas_course_id, but it will be populated going forward`

Also, I noted that this user doesn't exist in dim_user, which is a separate issue. I will work on a fix.

PR #1775, opened 2025-11-14
add additional org key values

KatelynGit:
https://github.com/mitodl/hq/issues/8886

Add missing org key values to the organizational field in the organization_administration_report table and remove the course title field for now since it's causing issues and might not be useful.

### How can this be tested?
dbt build --select organization_administration_report


rachellougee (approved):
As mentioned in the slack, we may want to add `organization_name` https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/intermediate/mitxonline/_int_mitxonline__models.yml#L2158 to this table since some of the b2b organization_key values aren't very readable. It should be an easy and useful change, but I’ll leave it up to you.

PR #1796, opened 2025-12-04
Add learn ai thread_id and checkpoint_pk to tfact_chatbot_events

rachellougee:

related to https://github.com/mitodl/hq/issues/9015#issuecomment-3581509568

Adds `learn_ai_thread_id` and `learn_ai_checkpoint_pk ` to tfact_chatbot_events, which would help us associate and align data between Learn AI and the open edX tracking logs. These two fields will be populated going forward.

### How can this be tested?

dbt build --select tfact_chatbot_events

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -859,3 +859,7 @@ models:
       the event occurred
     tests:
     - not_null
+  - name: learn_ai_thread_id
+    description: string, unique identifier for the chat thread in the Learn AI database.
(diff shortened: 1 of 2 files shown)


rachellougee:
> I think my only thought on this is should we backfill the data since there are a lot of null values is that doable?

@KatelynGit Thank you for the review. To answer your question about the backfill: without thread_id or checkpoint_pk in the tracking log, there’s no unique identifier to link the two data sources. That’s why we added these two identifiers in the tracking log— to allow us to associate the two sources. I believe Matt added two identifiers in the chat message in Sept, so it’s expected that older messages don’t have these fields populated.

PR #1797, opened 2025-12-05
add video metrics to instructor reporting

KatelynGit:
https://github.com/mitodl/hq/issues/9307

adds video metrics to instructor level reporting table

### How can this be tested?
 dbt build --select instuctor_module_report


rachellougee (approved):
dbt runs fine. I left a small comment below. Also, some queries appear similar to the ones in https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/dimensional/afact_video_engagement.sql, we can probably consolidate or reuse them to avoid repeating the same logic in multiple places. Just a thought - no need to address it here.

PR #1882, opened 2026-02-04
add order type to combined orders mart

KatelynGit:
https://github.mit.edu/xpro/xpro-issues/issues/656

add order type to combined orders mart

### How can this be tested?


quazi-h (approved):
Changes and logic looks good to me. Nulls are being handled properly. I was able to build the model successfully with no issues and I observed the new column on my dev build. I also verified that there was data for it on dev (B2B, B2C, Boeing, Staff, NULL).

PR #1906, opened 2026-02-13
Add mitxonline_course_engagements_daily_report

quazi-h:
https://github.com/mitodl/hq/issues/7644

This PR converts the Superset virtual dataset `marts__mitxonline_course_engagements_daily` into a physical dbt reporting model to improve query performance in Superset.

The new `mitxonline_course_engagements_daily_report` model extends the existing `marts__mitxonline_course_engagements_daily` mart by adding three columns that were previously calculated in the virtual dataset:
- `platform` - Set to 'MITx Online' for consistent cross-platform filtering
- `courserun_is_current` - Boolean indicating if the course run is currently active (checks if current date falls between courserun start and end dates)
- `course_readable_id` - Course identifier joined from `int__mitx__courses` for additional course metadata

### How can this be tested?
**Build and test the model:**
```bash
dbt build --select mitxonline_course_engagements_daily_report --vars 'schema_suffix: <your_username>' --target dev_production
```
Once this PR is merged and deployed, the corresponding Superset virtual dataset can be replaced with this physical table.


rachellougee (approved):
Looks good. Just a small comment but it's not a blocker

PR #1962, opened 2026-03-06
fix org chart date formats

KatelynGit:
https://github.com/mitodl/hq/issues/10436

All the charts on the Org dashboard https://bi.ol.mit.edu/superset/dashboard/p/M4e54gnqRbp/ need to be changed to "2026-1", "2026-2"  instead of months like "January".

### How can this be tested?
Review the code and check that the changes look good in QA


rachellougee (approved):
It looks good. but I would review sentry bot's feedback. It appears that sqlExpression still uses SQLdate_format(activity_date, '%c-%Y'), which does not match the intended format. It may be fine.

PR #2191, opened 2026-04-29
enhance user full name on learn user dashboard

KatelynGit:
https://github.com/mitodl/hq/issues/9331

enhance user full name on learn user dashboard to populate it more frequently

**there might be warnings but they are unrelated to this change**

### How can this be tested?
dbt build --select dim_user

### DIFF
--- a/src/ol_dbt/models/dimensional/dim_user.sql
+++ b/src/ol_dbt/models/dimensional/dim_user.sql
@@ -250,7 +250,10 @@ with mitx_users as (
         learn_user.user_global_id
         , learn_user.user_id as mitlearn_user_id
         , learn_user.user_email as email
-        , concat(learn_user.user_first_name, ' ', learn_user.user_last_name) as full_name
+        , case when learn_user.user_first_name is not null or learn_user.user_last_name is not null
+            then concat(learn_user.user_first_name, ' ', learn_user.user_last_name)
+            else learn_profile.user_name 
+            end as full_name
         , learn_profile.user_current_education as highest_education
         , learn_user.user_is_active as user_is_active_on_mitlearn
         , learn_user.user_joined_on as user_joined_on_mitlearn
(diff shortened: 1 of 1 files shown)


rachellougee (changes requested):
dbt runs successfully, but after reviewing the data, it appears there are actually more blank values than before the change. I’m not fully convinced this addresses the issue yet. Most of the Learn users in this table still don’t have a full name populated.

before the change 
```
--479378
select * from "ol_data_lake_production"."ol_warehouse_production_rlougee_dimensional".dim_user
where mitlearn_user_id is not null and (full_name is null or trim(full_name) = '')
```

after the change 

```
--496756
select * from "ol_data_lake_production"."ol_warehouse_production_rlougee_dimensional".dim_user
where mitlearn_user_id is not null and (full_name is null or trim(full_name) = '')
```

KatelynGit:
> dbt runs successfully, but after reviewing the data, it appears there are actually more blank values than before the change. I’m not fully convinced this addresses the issue yet. Most of the Learn users in this table still don’t have a full name populated.
> 
> before the change
> 
> ```
> --479378
> select * from "ol_data_lake_production"."ol_warehouse_production_rlougee_dimensional".dim_user
> where mitlearn_user_id is not null and (full_name is null or trim(full_name) = '')
> ```
> 
> after the change
> 
> ```
> --496756
> select * from "ol_data_lake_production"."ol_warehouse_production_rlougee_dimensional".dim_user
> where mitlearn_user_id is not null and (full_name is null or trim(full_name) = '')
> ```

I think it could definitely be improved. I tested it using this case statement instead and it increased the number of records with a populated full name so I'm going change it to that before merging:

case when (learn_profile.user_name is not null and trim(learn_profile.user_name) <> '')
        then learn_profile.user_name
        else concat(learn_user.user_first_name, ' ', learn_user.user_last_name)
            end as full_name

rachellougee (approved):
👍  There are still a large number of Learn users (321k) who don’t have full names, but it’s an improvement.

PR #2211, opened 2026-05-08
fix: Prevent infinite loop in legacy edX course export op

quazi-h:
Fixes https://github.com/mitodl/hq/issues/11104

The `export_edx_courses` op in the legacy IRx pipeline (which populates `s3://mitx-etl-mitxonline-production/<date>/courses/`) had two bugs that caused the `courses/` folder to silently go missing since ~2026-01-21.

**Bug 1 — Infinite loop in the status-polling while loop (primary cause)**

When `check_course_export_status()` returned HTTP 404 (e.g. the `django-user-tasks` record was cleaned up after a retention window, or a course was deleted from Studio), the bare `continue` inside the `except httpx.HTTPStatusError` block swallowed the error without ever adding the course to `failed_exports`. The while-loop condition `len(successful ∪ failed) < len(tasks)` was therefore never satisfied, so the job hung indefinitely. The Dagster run would eventually be killed and the `courses/` folder was never written to S3.

Fix: catch `HTTPStatusError`, log the HTTP status code and course ID for observability, and add the course to `failed_exports` so the loop can make progress. A 60-minute hard timeout (matching the pattern in the newer asset-based `openedx` code location) is also added as a safety net, and already-resolved courses are skipped to avoid redundant API calls.

**Bug 2 — Premature abort on partial-failure from the export trigger API**

The `ol_openedx_course_export` plugin returns **HTTP 400** (not 200) when *any* course fails to queue during the initial trigger phase — even when other courses were queued successfully. The unconditional `raise_for_status()` in `OpenEdxApiClient.export_courses()` caused the method to raise immediately, aborting the entire op before any successfully-queued tarballs could be processed.

Fix: allow HTTP 400 through without raising; log the initially-failed courses as a warning via `failed_uploads` in the response body. All other non-2xx codes still raise as before.

### How can this be tested?
1. Run the `mitxonline_edx_course_pipeline` Dagster job in staging/QA.
2. Confirm the `courses/` sub-folder appears in the dated S3 prefix (e.g. `s3://mitx-etl-mitxonline-qa/<date>/courses/`) with the expected course tarballs.
3. To reproduce Bug 1 locally: mock `check_course_export_status` to always raise `httpx.HTTPStatusError` with a 404 response and verify the op now terminates (previously it would loop forever).
4. To reproduce Bug 2 locally: mock `export_courses` to return a 400 with a mix of `upload_task_ids` and `failed_uploads`, and verify the op continues to process the successful tasks instead of aborting.

### Additional Context
- The same timeout pattern (`COURSE_EXPORT_GET_TASKS_STATUS_TIMEOUT = timedelta(minutes=60)`) already exists in the newer asset-based `openedx` code location (`dg_projects/openedx/openedx/assets/openedx.py`); this PR aligns the legacy op with that approach.
- Once this is merged and deployed, historical data for dates since 2026-01-21 will need to be backfilled by manually triggering the pipeline with a `date_override` for each missing date.

### DIFF
--- a/dg_projects/legacy_openedx/legacy_openedx/ops/open_edx.py
+++ b/dg_projects/legacy_openedx/legacy_openedx/ops/open_edx.py
@@ -31,6 +31,8 @@ from pydantic import Field
 from pypika import MySQLQuery as Query
 from pypika import Table, Tables
 
+COURSE_EXPORT_TIMEOUT = timedelta(minutes=60)
+
 
 class ListCoursesConfig(Config):
     edx_course_api_page_size: int = Field(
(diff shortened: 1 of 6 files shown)


quazi-h:
Closing in favour of a clean branch rebased on main — see #2212

PR #2238, opened 2026-05-21
feat(lakehouse): automate Iceberg table maintenance across the lakehouse

blarghmatey:

N/A — operational hygiene; motivated by a Glue catalog scan finding 1,331 raw-layer Iceberg tables that have never had maintenance run. The worst offender (`raw__edxorg__program_learner_report`) has ~70K snapshots producing a ~110 MB `metadata.json` that is re-downloaded on every Airbyte sync, every dbt source read, and every local DuckDB query.

Adds two nightly Dagster maintenance assets to the `lakehouse` code location plus the shared library and Trino resource they depend on.

**Four maintenance operations, two execution engines:**

| Operation | Engine | Why that engine |
|---|---|---|
| OPTIMIZE (compact small files) | Trino / Starburst Galaxy | Distributes file rewriting across Galaxy workers; incremental dbt models produce one small Parquet file per run |
| ANALYZE (collect query stats) | Trino / Starburst Galaxy | Not available in pyiceberg; required for Trino query optimizer |
| EXPIRE SNAPSHOTS | pyiceberg | Metadata-only op (Glue API + targeted S3 deletes); no distributed engine needed |
| REMOVE ORPHAN FILES | pyiceberg | Same; stub for now — pyiceberg 0.11.x does not yet expose a built-in API |

**Two assets, staggered schedules (both ship as `STOPPED`):**

- `iceberg_dbt_layer_maintenance` — 02:00 UTC. Reads `manifest.json`, uses the Dagster materialization event log to count runs since last maintenance, gates OPTIMIZE/ANALYZE on configurable thresholds (`optimize_after_every_n_runs`, `analyze_after_every_n_runs`). Also covers non-dbt singleton tables (`student_risk_probability`).
- `iceberg_raw_layer_maintenance` — 03:00 UTC. Live Glue catalog scan of `ol_warehouse_production_raw`; EXPIRE + ORPHAN only; `ThreadPoolExecutor(max_workers=8)` to handle 1,300+ tables without overwhelming Glue API rate limits.

**Phase-0 remediation in `scripts/expire_iceberg_snapshots.py`:**
Adds the 5 worst raw-layer offenders to the existing script so they can be cleaned up immediately without waiting for the full automation system to be enabled. Also fixes the pyiceberg API call: `table.expire_snapshots()` does not exist on `Table` in pyiceberg 0.11.x; the correct path is `table.maintenance.expire_snapshots().older_than(dt).commit()`.

**New files:**
- `packages/ol-orchestrate-lib/src/ol_orchestrate/lib/iceberg_maintenance.py` — shared library (catalog factory, maintenance ops, manifest parser, Glue scanner, config dataclasses and constants)
- `packages/ol-orchestrate-lib/src/ol_orchestrate/resources/trino_maintenance.py` — `TrinoMaintenanceResource(ConfigurableResource)` with `optimize()`, `analyze()`, `execute()`; Vault credentials with env-var fallback for local dev
- `dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py` — the two Dagster assets

**Modified files:**
- `scripts/expire_iceberg_snapshots.py` — phase-0 remediation + API fix
- `packages/ol-orchestrate-lib/pyproject.toml` — add `trino>=0.329`
- `src/ol_dbt/dbt_project.yml` — add `+meta.iceberg_maintenance` at project root (7-day defaults) and per-folder overrides in `dimensional` (14-day, time-travel window for delete+insert models) and `marts` (14-day snapshots)
- `dg_projects/lakehouse/lakehouse/definitions.py` — register `TrinoMaintenanceResource`, both assets, `trino_host_map`/`trino_catalog_map`, and two `DefaultScheduleStatus.STOPPED` nightly schedules

### How can this be tested?

**Phase-0 script (immediate):**
```bash
# Dry-run to verify snapshot counts before touching anything
uv run python scripts/expire_iceberg_snapshots.py --dry-run

# Run for real (requires production AWS credentials)
uv run python scripts/expire_iceberg_snapshots.py --older-than-days 7
```

**Library unit smoke test:**
```bash
# Verify manifest parsing returns configs for all dbt models
uv run python -c "
from ol_orchestrate.lib.iceberg_maintenance import load_maintenance_configs_from_manifest
configs = load_maintenance_configs_from_manifest('src/ol_dbt/target/manifest.json')
print(f'{len(configs)} model configs loaded')
print(next(c for c in configs if c.schema_name.endswith('dimensional')))
"
```

**Dagster asset (staging/production):**
1. Ensure `DAGSTER_TRINO_VAULT_PATH` is set (or `DBT_TRINO_USERNAME`/`DBT_TRINO_PASSWORD` for local dev)
2. In the Dagster UI, navigate to the `lakehouse_maintenance` asset group
3. Manually materialize `iceberg_raw_layer_maintenance` first (EXPIRE only, lower risk)
4. Check the Output metadata for `tables_scanned`, `tables_cleaned`, `failure_count`
5. Once verified, enable the schedules via the Dagster UI

### Additional Context

- Both schedules ship as `DefaultScheduleStatus.STOPPED`. Enable them explicitly in production after verifying a manual run succeeds and confirming the Trino service account has write permissions on the catalog.
- `remove_orphan_files` is a no-op stub — pyiceberg 0.11.x does not expose a built-in orphan removal API. The docstring explains exactly what to replace it with when upstream support lands.
- The raw-layer asset does not use the Dagster event log for timing (unlike the dbt layer) because `OLAirbyteTranslator` slugifies connection names in a way that doesn't map cleanly back to Glue table names. Iceberg snapshot timestamps are used instead.
- IAM pre-check before enabling: confirm the Dagster worker role has `s3:DeleteObject` on `ol-data-lake-*-production/*` before enabling orphan removal once the pyiceberg API ships.

### DIFF
--- a/dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
+++ b/dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
@@ -47,6 +47,7 @@ from dagster import (
     Output,
     asset,
 )
+from ol_orchestrate.lib.constants import DAGSTER_ENV
 from ol_orchestrate.lib.iceberg_maintenance import (
     NON_DBT_SINGLETON_TABLES,
     TableMaintenanceConfig,
@@ -63,8 +64,19 @@ from lakehouse.assets.lakehouse.dbt import dbt_project
 
 log = logging.getLogger(__name__)
 
-RAW_GLUE_DATABASE = "ol_warehouse_production_raw"
-ENV_SCHEMA_PREFIX = "ol_warehouse_production"
+# Each environment runs maintenance only against its own catalog and schema set.
+# dev uses production because DAGSTER_ENV="dev" targets the production dbt profile
+# (dev_production), meaning developers run against real production Iceberg tables.
+_RAW_GLUE_DATABASE_MAP: dict[str, str] = {
+    "production": "ol_warehouse_production_raw",
+    "qa": "ol_warehouse_qa_raw",
+    "dev": "ol_warehouse_production_raw",
+    "ci": "ol_warehouse_qa_raw",
+}
+
+RAW_GLUE_DATABASE = _RAW_GLUE_DATABASE_MAP.get(
+    DAGSTER_ENV, "ol_warehouse_production_raw"
+)
 # Parallelism for raw layer: enough to be fast across 1,300+ tables without
 # hitting Glue API rate limits (default per-account limit is ~200 req/s).
 RAW_LAYER_WORKERS = 8
@@ -230,7 +242,6 @@ def iceberg_dbt_layer_maintenance(
     """Run Iceberg maintenance for all dbt-managed tables and non-dbt singletons."""
     configs = load_maintenance_configs_from_manifest(
         manifest_path=dbt_project.manifest_path,
-        env_schema_prefix=ENV_SCHEMA_PREFIX,
     )
     all_configs = [*configs, *NON_DBT_SINGLETON_TABLES]
     context.log.info(
--- a/dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
+++ b/dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
@@ -1,0 +1,398 @@
+"""Dagster assets for Iceberg table maintenance across the OL data lakehouse.
+
+Two assets run on staggered nightly schedules:
+
+``iceberg_dbt_layer_maintenance`` (02:00 UTC)
+    Processes all dbt-managed tables discovered from ``manifest.json`` plus the
+    non-dbt singleton tables listed in ``NON_DBT_SINGLETON_TABLES``.
+
+    For each table it runs:
+    - OPTIMIZE  (Trino)   — compact small files; triggered when materialization
+                            count since last maintenance >= optimize_after_every_n_runs
+    - ANALYZE   (Trino)   — refresh query statistics; same threshold logic with
+                            analyze_after_every_n_runs
+    - EXPIRE SNAPSHOTS (pyiceberg) — always runs; uses snapshot_retention_days
+    - REMOVE ORPHAN FILES (pyiceberg) — always runs; uses orphan_retention_days
+
+    Config comes from ``dbt_project.yml`` ``+meta.iceberg_maintenance`` blocks,
+    compiled into each node's ``config.meta.iceberg_maintenance`` in the manifest.
+    Per-model overrides in schema YAML files are shallow-merged on top.
+
+``iceberg_raw_layer_maintenance`` (03:00 UTC)
+    Processes all Iceberg tables in ``ol_warehouse_production_raw`` (1,300+
+    Airbyte-ingested tables).  Uses a live Glue catalog scan rather than the
+    Dagster event log because the OLAirbyteTranslator slugifies connection names
+    in a way that doesn't map cleanly back to Glue table names.
+
+    Runs EXPIRE SNAPSHOTS and REMOVE ORPHAN FILES only — OPTIMIZE and ANALYZE
+    are not needed for raw tables since they are not Trino analytics targets and
+    Airbyte writes complete files per sync (no small-file accumulation).
+
+    Tables are processed in a ThreadPoolExecutor(max_workers=8) to handle the
+    volume without overwhelming the Glue API rate limits.
+"""
+
+from __future__ import annotations
+
+import logging
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Any
+
+from dagster import (
+    AssetExecutionContext,
+    AssetKey,
+    DagsterEventType,
+    EventRecordsFilter,
+    MetadataValue,
+    Output,
+    asset,
+)
+from ol_orchestrate.lib.iceberg_maintenance import (
+    NON_DBT_SINGLETON_TABLES,
+    TableMaintenanceConfig,
+    expire_snapshots,
+    get_glue_catalog,
+    load_maintenance_configs_from_manifest,
+    load_raw_layer_maintenance_work,
+    raw_config_for_table,
+    remove_orphan_files,
+)
+from ol_orchestrate.resources.trino_maintenance import TrinoMaintenanceResource
+
+from lakehouse.assets.lakehouse.dbt import dbt_project
+
+log = logging.getLogger(__name__)
+
+RAW_GLUE_DATABASE = "ol_warehouse_production_raw"
+ENV_SCHEMA_PREFIX = "ol_warehouse_production"
+# Parallelism for raw layer: enough to be fast across 1,300+ tables without
+# hitting Glue API rate limits (default per-account limit is ~200 req/s).
+RAW_LAYER_WORKERS = 8
+
+
+# ── Helpers ───────────────────────────────────────────────────────────────────
+
+
+def _get_last_maintenance_cursor(
+    instance, maintenance_asset_key: AssetKey
+) -> int | None:
+    """Return the storage_id of the most recent maintenance materialization.
+
+    This cursor is passed to ``_count_materializations_since`` so we only count
+    dbt model materializations that occurred *after* the previous maintenance run.
+    Returns ``None`` on the first ever run, which the caller treats as "run all
+    operations unconditionally".
+    """
+    records = instance.get_event_records(
+        EventRecordsFilter(
+            event_type=DagsterEventType.ASSET_MATERIALIZATION,
+            asset_key=maintenance_asset_key,
+        ),
+        limit=1,
+        ascending=False,
+    )
+    return records[0].storage_id if records else None
+
+
+def _count_materializations_since(
+    instance,
+    asset_key: AssetKey,
+    after_cursor: int | None,
+) -> int:
+    """Count how many times *asset_key* was materialized since *after_cursor*.
+
+    Up to 500 records are fetched — enough to saturate any reasonable
+    optimize_after_every_n_runs or analyze_after_every_n_runs threshold.
+    """
+    if after_cursor is not None:
+        records = instance.get_event_records(
+            EventRecordsFilter(
+                event_type=DagsterEventType.ASSET_MATERIALIZATION,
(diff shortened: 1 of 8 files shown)


src/ol_dbt/dbt_project.yml
```diff
@@ -122,6 +133,25 @@ models:
       format: "'PARQUET'"
     +meta:
       required_docs: true
+      # Iceberg maintenance config -- inherited by all models, overrideable per
+      # folder (see dimensional/marts above) or per-model in schema YAML files.
+      #
+      # dbt meta merging is a shallow dict merge: a per-model override of
+      # iceberg_maintenance replaces the entire dict, not individual keys.
+      # load_maintenance_configs_from_manifest() re-applies these values as the
+      # base before overlaying node-level meta so partial overrides work.
```
quazi-h: I believe this comment is outdated due to the change in your second commit (removing DEFAULT_MAINTENANCE_CONFIG and base.update(node_iceberg) overlay

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/iceberg_maintenance.py
```diff
@@ -0,0 +1,450 @@
...
+                model_name=model_name,
+                schema_name=schema_name,
+                materialized=materialized,
+                # AssetKey: [schema_suffix, model_name] per DbtAutomationTranslator
+                asset_key=[schema_suffix, model_name],
+                enabled=iceberg_cfg.get("enabled", True),
+                snapshot_retention_days=iceberg_cfg["snapshot_retention_days"],
+                orphan_retention_days=iceberg_cfg["orphan_retention_days"],
+                optimize_after_every_n_runs=iceberg_cfg["optimize_after_every_n_runs"],
+                analyze_after_every_n_runs=iceberg_cfg["analyze_after_every_n_runs"],
```
quazi-h: I hit the following error when running the library unit smoke test:
```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
    configs = load_maintenance_configs_from_manifest('src/ol_dbt/target/manifest.json')
  File "/Users/qhoque/GIT/ol-data-platform/packages/ol-orchestrate-lib/src/ol_orchestrate/lib/iceberg_maintenance.py", line 355, in load_maintenance_configs_from_manifest
    orphan_retention_days=iceberg_cfg["orphan_retention_days"],
                          ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'orphan_retention_days'
```

This is due to partial iceberg_maintenance overrides hitting KeyError within the load_maintenance_configs_from_manifest asset. Since the asset uses direct list indexing on the compiled dbt model and then DBT meta does a shallow merge, any single-key per-model override in a schema YAML will lead to the entire entire dictionary getting replaced with just that one key and the manifest node will be missing the other required keys.

I believe there are 72 marts models that have partial overrides in the compiled manifest (they only have snapshot_retention_days) and are missing the other 3 keys.

I think this could be addressed by completing the YAML overrides and adding the missing keys **Dimensional** - optimize_after_every_n_runs and analyze_after_every_n_runs
**Marts** - orphan_retention_days, optimize_after_every_n_runs, analyze_after_every_n_runs
but copilot sees this as a band-aid fix because: any future partial overrides at the folder or model level in schema.yml will cause the same KeyError and this defeats the purpose of partial overrides.

The preferred method would be to restore the DEFAULT overlay since it is more defensive and allows partial overrides and matches the documented behavior (comment in dbt_project.yml that I flagged in my other comment on this PR).

I did confirm that rolling back the changes to the DEFAULT overlay and partial overrides to the code in your initial commit made the smoke test pass.

dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
```diff
@@ -0,0 +1,409 @@
...
+            "tables_processed": MetadataValue.int(tables_processed),
+            "tables_optimized": MetadataValue.int(tables_optimized),
+            "tables_analyzed": MetadataValue.int(tables_analyzed),
+            "snapshots_expired": MetadataValue.int(snapshots_expired),
+            "orphans_removed": MetadataValue.int(orphans_removed),
+            "failure_count": MetadataValue.int(len(failures)),
+            # Cap at 20 entries so the Dagster UI doesn't time out rendering
+            "failure_details": MetadataValue.json(failures[:20]),
+        },
+    )
```
quazi-h, line 336:
```suggestion
    # Fail the asset if maintenance failures exceed threshold
    failure_threshold = max(1, int(tables_processed * 0.1)) # 10% or at least 1
    if len(failures) >= failure_threshold:
        context.log.error(
            "Maintenance failed for %d/%d tables (threshold: %d). Failing asset."
            len(failures,
            tables_processed,
            failure_threshold,
        )
        raise RuntimeError(
            f"Iceberg maintenance failed for {len(failures)}/{tables_processed} tables. "
            f"First failures: {failures[:5]}"
        )
    return Output(
        value=None,
        metadata={
            "tables_processed": MetadataValue.int(tables_processed),
            "tables_optimized": MetadataValue.int(tables_optimized),
            "tables_analyzed": MetadataValue.int(tables_analyzed),
            "snapshots_expired": MetadataValue.int(snapshots_expired),
            "orphans_removed": MetadataValue.int(orphans_removed),
            "failure_count": MetadataValue.int(len(failures)),
            # Cap at 20 entries so the Dagster UI doesn't time out rendering
            "failure_details": MetadataValue.json(failures[:20]),
        },
    )
```
Since this asset always returns Output(value=None) regardless of failure count, Dagster would mark the runs as SUCCESS even when maintenance fails for multiple tables. This advances the maintenance cursor and prevents retries. If we have a scenario where Trino credentials break or the Glue API is unavailable, failure_count accumulates in metadata but operations silently stop. Low-frequency models won't meet the optimize threshold for weeks. Raising an exception here or returning a failure Output when len(failures) exceeds a threshold would help alert us when this happens.

dg_projects/lakehouse/lakehouse/assets/iceberg_maintenance.py
```diff
@@ -0,0 +1,409 @@
...
+        value=None,
+        metadata={
+            "tables_scanned": MetadataValue.int(len(tables)),
+            "tables_cleaned": MetadataValue.int(tables_cleaned),
+            "snapshots_expired": MetadataValue.int(snapshots_expired),
+            "orphans_removed": MetadataValue.int(orphans_removed),
+            "failure_count": MetadataValue.int(len(failures)),
+            "failure_details": MetadataValue.json(failures[:20]),
+        },
+    )
```
quazi-h, line 451:
```suggestion
    # Fail the asset if maintenance failures exceed threshold
    failure_threshold = max(1, int(tables_processed * 0.05)) # 5% or at least 10
    if len(failures) >= failure_threshold:
        context.log.error(
            "Maintenance failed for %d/%d tables (threshold: %d). Failing asset."
            len(failures,
            tables_processed,
            failure_threshold,
        )
        raise RuntimeError(
            f"Iceberg maintenance failed for {len(failures)}/{tables_processed} tables. "
            f"First failures: {failures[:5]}"
        )
    return Output(
        value=None,
        metadata={
            "tables_scanned": MetadataValue.int(len(tables)),
            "tables_cleaned": MetadataValue.int(tables_cleaned),
            "snapshots_expired": MetadataValue.int(snapshots_expired),
            "orphans_removed": MetadataValue.int(orphans_removed),
            "failure_count": MetadataValue.int(len(failures)),
            "failure_details": MetadataValue.json(failures[:20]),
        },
    )
```
Same reasoning as the asset iceberg maintenance asset above:
_Since this asset always returns Output(value=None) regardless of failure count, Dagster would mark the runs as SUCCESS even when maintenance fails for multiple tables. This advances the maintenance cursor and prevents retries. If we have a scenario where Trino credentials break or the Glue API is unavailable, failure_count accumulates in metadata but operations silently stop. Low-frequency models won't meet the optimize threshold for weeks. Raising an exception here or returning a failure Output when len(failures) exceeds a threshold would help alert us when this happens._

src/ol_dbt/dbt_project.yml
```diff
@@ -94,6 +94,13 @@ models:
       +grants:
         select: ['read_only_production', 'ol_data_analyst', 'business_intelligence',
           'ol_business_analyst', 'reverse_etl']
+      # Incremental dims use delete+insert and accumulate one small file per run.
+      # 14-day snapshot retention provides a time-travel window for point-in-time
+      # queries and safe re-runs across two weekly optimize cycles.
+      +meta:
+        iceberg_maintenance:
+          snapshot_retention_days: 14
+          orphan_retention_days: 14
```
quazi-h, line 103:
```suggestion
          orphan_retention_days: 14
          optimize_after_every_n_runs: 1
          analyze_after_every_n_runs: 7
```
Even though there is a global +meta config defined at line 134 below, dbt's shallow merge means the folder-level dict defined here replaces the parent's keys instead of merging all the inner keys. This is expected behavior, as documented by your comment on lines 142-143, and results in a KeyError when running the smoke test.

src/ol_dbt/dbt_project.yml
```diff
@@ -106,6 +113,10 @@ models:
       +grants:
         select: ['read_only_production', 'ol_data_analyst', 'business_intelligence',
           'ol_business_analyst', 'reverse_etl', 'finance']
+      # 14-day retention so mart snapshots survive a missed nightly run.
+      +meta:
+        iceberg_maintenance:
+          snapshot_retention_days: 14
```
quazi-h, line 121:
```suggestion
          snapshot_retention_days: 14
          orphan_retention_days: 7
          optimize_after_every_n_runs: 1
          analyze_after_every_n_runs: 7
```
Even though there is a global +meta config defined at line 134 below, dbt's shallow merge means the folder-level dict defined here replaces the parent's keys instead of merging all the inner keys. This is expected behavior, as documented by your comment on lines 142-143, and results in a KeyError when running the smoke test.

quazi-h (approved):
Smoke tests are now passing, all the feedback I brought up has been addressed.

Issue #2073, opened 2026-04-06
[Epic] Complete pending dimensional warehouse model improvements (Phases B, C, G)


blarghmatey:
## Context

During the development of the dimensional warehouse layer, a comprehensive set of improvements was identified and documented in `dimensional_warehouse_todo.md`. Phases A, D, E, F have been completed. This epic tracks the remaining pending work in Phases B, C, and G.

## Phase B — Complete Platform Coverage

The dimensional layer currently covers MITx Online, MITx Pro, edX.org, and Residential OpenEdX. Phase B extends it to Bootcamps and OCW, and fixes incremental watermark bugs.

## Phase C — New Models and Refinements

New fact tables and dimension enrichments that add analytical depth to the star schema.

## Phase G — Grain Audit Findings

Correctness improvements identified during a comprehensive grain audit of all dimensional models. These address fan-out risks, missing QUALIFY guards, and documentation gaps.

## Definition of Done

- All Phase B, C, and G tasks have associated completed PRs
- All new/modified dimensional models pass their dbt tests
- `dimensional_warehouse_todo.md` reflects completed status

blarghmatey:
## 2026-07 warehouse audit — additional finding (epic #2375)

The audit identified the **only two pieces of mart-layer content with no dimensional successor** — these are the concrete blockers to retiring the MicroMasters and MITx Online user marts, and are good candidates for Phase B/C here:

1. **Proctored exam grades** — surfaced only by `marts__micromasters_dedp_exam_grades` (grep for `proctored` in `models/dimensional` returns nothing). → add `tfact_exam_grade`, or a `grade_type` discriminator on `tfact_grade`.
2. **Flexible-pricing income** — `int__mitxonline__flexiblepricing_flexiblepriceapplication` surfaced only via `marts__mitxonline_user_profiles.latest_income_usd` (propagated into `marts__combined__users`). → add an income attribute/fact keyed to `dim_user`.

Adding these unblocks deleting/migrating the MicroMasters and MITx Online user marts (see the dead-model deletion work under #2072).

Tracked as witan task `tk-add-dimensional-successors-for-exam-grades-and-f-8cd9e3`.

blarghmatey:
Triage pass 2026-08-20 - re-verified every Phase B/C/G task in `dimensional_warehouse_todo.md` against `src/ol_dbt/models/dimensional/` on main. **10 of 12 are implemented.** Only B4 and C2 remain.

[...]

PR #2259, opened 2026-05-28
C1: Create tfact_grade — course run grade fact table

### LINKED ISSUE #2125: C1: Create `tfact_grade` — course run grade fact table
**Epic:** #2073

Add a new fact table for course run grades to the dimensional layer.

quazi-h:
Closes https://github.com/mitodl/ol-data-platform/issues/2125
Epic: https://github.com/mitodl/ol-data-platform/issues/2073

Adds `tfact_grade` — a new incremental transaction fact table in the dimensional star schema.

**Grain:** One row per (user × course run) grade record.

**Platforms covered:** MITx Online, MITx Pro, edX.org

**Source models:**
- `int__mitxonline__courserun_grades` — filtered to `courserun_platform = 'MITx Online'` to exclude edX.org-hosted MITx Online course grades (which are captured via the edxorg source and would otherwise be double-counted)
- `int__mitxpro__courserun_grades`
- `int__edxorg__mitx_courserun_grades` — uses `courserungrade_user_grade` (column name differs from other platforms); timestamps unavailable so `grade_date_key` and `grade_created_on` are null for edX.org rows; `grade_id` is a surrogate from `(user_id, courserun_readable_id)`

**Key implementation details:**
- Incremental strategy: `delete+insert` on `grade_key`; watermark computed via a pre-computed CTE per platform (avoids Trino correlated-subquery fan-out anti-pattern, consistent with `tfact_certificate` and `tfact_enrollment`)
- edX.org rows are fully re-ingested on each incremental run (null timestamps; `delete+insert` is idempotent)
- Defensive `ROW_NUMBER` dedup guard (Trino does not support `QUALIFY`)
- Foreign keys: `dim_user`, `dim_course_run` (`is_current = true`), `dim_platform`, `dim_date`
- edX.org rows include MicroMasters-linked grades (`micromasters_program_id IS NOT NULL`); a separate MicroMasters platform source can be added in a future phase

### How can this be tested?
1. Compile: `cd src/ol_dbt && ol-dbt run --select tfact_grade` (or `dbt compile --select tfact_grade`)
2. Run against QA: `ol-dbt run --select tfact_grade --target dev_qa`
3. Validate row counts and pass rates:
```sql
SELECT
    platform,
    COUNT(*) AS grade_count,
    ROUND(AVG(CAST(is_passing AS double)), 2) AS pass_rate
FROM ol_warehouse_production_dimensional.tfact_grade
GROUP BY platform;
-- Expected: pass_rate ~0.5–0.7 for mitxonline; edxorg rows have null grade_date_key
```
4. Verify FK integrity (run dbt tests): `dbt test --select tfact_grade`

### Additional Context
The platform-filtering fix for MITx Online (filtering to `courserun_platform = 'MITx Online'`) was identified during a pre-implementation rubber duck review. `int__mitxonline__courserun_grades` contains grades for both MITx Online-hosted and edX.org-hosted course runs; without this filter, edX.org course grades would appear twice in the fact table (once from mitxonline, once from edxorg).

This model is a new leaf node — it consumes existing intermediates and has no impact on any existing downstream models (confirmed via OpenMetadata lineage analysis).


src/ol_dbt/models/dimensional/tfact_grade.sql
```diff
@@ -0,0 +1,215 @@
...
+        on (
+            -- mitxonline/mitxpro: join on integer source_id + platform
+            (combined_grades.platform in ('mitxonline', 'mitxpro')
+                and combined_grades.courserun_id = dim_course_run.source_id
+                and combined_grades.platform = dim_course_run.platform)
+            -- edxorg: join on readable_id + platform to prevent fan-out across platforms
+            or (combined_grades.platform = 'edxorg'
+                and combined_grades.courserun_readable_id = dim_course_run.courserun_readable_id
+                and dim_course_run.platform = 'edxorg')
+        )
```
rachellougee: No need to change it now, but the OR clause in the dim_course_run join here has a performance cost . This was likely copied from `tfact_enrollment` and `tfact_certificate` which have the same pattern. 

All three platforms could join dim_course_run on (courserun_readable_id, platform) if we populate courserun_readable_id in above CTEs, the same natural key used to build `courserun_pk`, eliminating the OR entirely.

rachellougee (approved):
dbt build fine but a bit slow. I left a comment but not a blocker.

src/ol_dbt/models/dimensional/tfact_grade.sql
quazi-h: Good call — I've populated `courserun_readable_id` from the mitxonline and mitxpro intermediates (both already expose it) so all three platforms now join `dim_course_run` on `(courserun_readable_id, platform)` uniformly. The `source_id` column and the OR branch are gone. Also merged latest main in.

rachellougee (approved):
Change looks good. The build ran faster than last time.

PR #2264, opened 2026-05-29
feat: add TikaResource to ol-orchestrate-lib

blarghmatey:

https://github.com/mitodl/hq/issues/11513

Adds `TikaResource` to `ol-orchestrate-lib` — a `ConfigurableResource` wrapping the OL-deployed Apache Tika REST API for extracting plain text and metadata from binary documents (PDFs, HTML, Office files, etc.).

This is a prerequisite for Cohort 4 OCW text extraction and OpenEdX ContentFile extraction assets. Tika is already deployed and running; this resource provides the standard way for Dagster assets to call it.

**API surface:**

```python
tika = TikaResource(
    base_url="https://tika-production.ol.mit.edu",
    access_token=vault_secret["value"],
)

# Extract plain text from a PDF
text = tika.extract_text(pdf_bytes, "application/pdf")

# Extract plain text with OCR strategy
text = tika.extract_text(pdf_bytes, "application/pdf", ocr_strategy="no_ocr")

# Extract metadata
meta = tika.extract_metadata(pdf_bytes, "application/pdf")

# Check if a MIME type is supported before downloading the file
if tika.is_supported(content_type):
    text = tika.extract_text(file_bytes, content_type)
```

**Implementation notes:**
- `httpx.Client` is created lazily on first use and closed in `yield_for_execution`
- Default timeout is 300 s (large PDFs can take 60–120 s)
- Auth via `X-Access-Token` header read from Vault at execution time
- Calls the Tika HTTP API directly with `httpx` rather than using the `tika` Python library, which manages its own JVM — not appropriate when calling a remote service
- `SUPPORTED_CONTENT_TYPES` frozenset guards against sending unsupported files to Tika (which would waste time and return garbage)

**Vault paths for wiring in definitions.py:**
- Production: `secret-operations/production-apps/tika/access-token` → key `value`
- QA: `secret-operations/rc-apps/tika/access-token` → key `value`

Infrastructure source: [`ol-infrastructure/src/ol_infrastructure/applications/tika/`](https://github.com/mitodl/ol-infrastructure/tree/main/src/ol_infrastructure/applications/tika/)

### How can this be tested?

13 unit tests covering:
- `extract_text`: successful extraction, empty response → None, unsupported MIME type → None (no HTTP call), OCR strategy header, error propagation
- `extract_metadata`: JSON parsing, non-JSON fallback to `{}`
- `is_supported`: true/false cases
- `yield_for_execution`: client closed on normal exit and on exception

```bash
cd packages/ol-orchestrate-lib
uv run pytest tests/resources/test_tika.py -v
# → 13 passed
```

For an integration test against the real service (requires VPN + Vault access):
```python
from ol_orchestrate.resources.tika import TikaResource
tika = TikaResource(base_url="https://tika-qa.ol.mit.edu", access_token="<token>")
text = tika.extract_text(open("sample.pdf", "rb").read(), "application/pdf")
assert text and len(text) > 0
```


rachellougee (approved):
Looks good.  I've ran integration test against tika-qa.ol.mit.edu, `extract_text` successfully extracted text from a sample PDF with the QA access token.

PR #2285, opened 2026-06-08
fix(dbt): harden DEDP fan-out guard in int__mitxonline__courserunenrollments

### LINKED ISSUE #2131: G4: Harden `int__mitxonline__courserunenrollments` DEDP fan-out guard with explicit comment
**Epic:** #2073

**Priority: LOW-MEDIUM** — currently protected by `select distinct` but pattern is fragile.

quazi-h:
Closes #2131

Adds deduplication to `stg__mitxonline__app__postgres__courses_courserunenrollment` to fix confirmed production duplicate `(user_id, courserun_id)` pairs.

**Root cause:** Django enforces `unique_together` on `(user, course_run)`, but application-level hard-deletes followed by re-enrollment produce two rows with different IDs for the same `(user_id, run_id)` in the raw Iceberg table (append-only CDC). The staging model was passing both rows through unchanged.

Confirmed from production raw data:

| id | user_id | run_id | active | mode | created |
|---|---|---|---|---|---|
| 470891 | 42170 | 533 | True | verified | 2025-10-03 |
| 477165 | 42170 | 533 | True | audit | 2025-10-23 |
| 966253 | 1294262 | 3487 | True | verified | 2026-05-20 |
| 967247 | 1294262 | 3487 | False | audit/unenrolled | 2026-05-21 |

Both affected runs are UAI courses (program_id=14), not DEDP.

**The fix:** `deduplicate_raw_table(order_by="id", partition_columns="user_id, run_id")`

Ordering logic:
- **`id DESC`** — keep the highest id (most recently created enrollment). Auto-increment PKs are monotonically increasing: the re-enrolled record always receives a higher id than any previously hard-deleted enrollment.
- `active` is **not** used as a sort key. A hard-deleted enrollment still has `active=True` in the append-only Iceberg table, so sorting by `active DESC` would incorrectly prefer the deleted ghost record over the current live one.

Follows the same pattern as `ecommerce_basket`, `wagtailcore_revision`, and other staging models that handle Airbyte incremental-append duplicates.

> ⚠️ A **full refresh** of `stg__mitxonline__app__postgres__courses_courserunenrollment` in production is required after deploy to clear existing duplicate rows from the materialized table.

### How can this be tested?

```bash
source .env

# Build staging model + tests
uv run ol-dbt run --select stg__mitxonline__app__postgres__courses_courserunenrollment --target dev_qa --vars '{"schema_suffix": "<suffix>"}'
# Expected: PASS=15 WARN=0 ERROR=0

# Build intermediate model + tests
uv run ol-dbt run --select int__mitxonline__courserunenrollments --target dev_qa --vars '{"schema_suffix": "<suffix>"}'
# Expected: PASS=23 WARN=0 ERROR=0
```

Validated locally against QA Trino: **PASS=38 WARN=0 ERROR=0**.

After production deploy + full refresh, confirm:
```sql
SELECT user_id, courserun_id, COUNT(*) AS cnt
FROM ol_data_lake_production.ol_warehouse_production_staging.stg__mitxonline__app__postgres__courses_courserunenrollment
GROUP BY user_id, courserun_id HAVING COUNT(*) > 1;
-- Expected: 0 rows
```

### Additional Context

Production validation against both known duplicate pairs (run in dev schema against production raw data):

| user_id | courserun_id | returned id | active | mode |
|---------|-------------|------------|--------|------|
| 42170 | 533 | 477165 | True | audit |
| 1294262 | 3487 | 967247 | False | audit |

Both match the expected current live records.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
     - src/ol_dbt/package-lock.yml
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: 'v0.15.15'
+  rev: 'v0.15.16'
   hooks:
   - id: ruff-format
   - id: ruff-check
@@ -71,7 +71,7 @@ repos:
     - types-pymysql
     - types-requests
 - repo: https://github.com/sqlfluff/sqlfluff
-  rev: 4.2.1
+  rev: 4.2.2
   hooks:
   - id: sqlfluff-fix
       # Arbitrary arguments to show an example
--- a/dg_projects/data_loading/.dlt/config.toml
+++ b/dg_projects/data_loading/.dlt/config.toml
@@ -29,6 +29,22 @@ destination_type = "filesystem"
 bucket_url = "file:///tmp/.dlt/data"
 layout = "{table_name}/{load_id}.{file_id}.{ext}"
 
+# Podcast RSS destinations (used by podcast_rss_ingest)
+[destination.podcast_local]
+destination_type = "filesystem"
+bucket_url = "file:///tmp/.dlt/data/podcast"
+layout = "{table_name}/{load_id}.{file_id}.{ext}"
+
+[destination.podcast_qa]
+destination_type = "filesystem"
+bucket_url = "s3://ol-data-lake-raw-qa/podcast"
+layout = "{table_name}/{load_id}.{file_id}.{ext}"
+
+[destination.podcast_production]
+destination_type = "filesystem"
+bucket_url = "s3://ol-data-lake-raw-production/podcast"
+layout = "{table_name}/{load_id}.{file_id}.{ext}"
+
 # QA Destination - Filesystem with Iceberg + Glue Catalog (named destination)
 # Writes Iceberg tables to S3 and registers in ol_warehouse_qa_raw Glue database
 [destination.qa]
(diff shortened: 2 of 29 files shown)


rachellougee:
I originally wrote this DEDP query because the micromasters orders exist in two systems. I looked into the two duplicated records in `int__mitxonline__courserunenrollments` on production, and they actually don’t seem to be DEDP-related. The duplicates are caused by **deleted enrollments** for UAI courses in raw enrollment table.

```
SELECT user_id, courserun_id, COUNT(*) AS cnt
FROM"ol_data_lake_production"."ol_warehouse_production_intermediate".int__mitxonline__courserunenrollments
GROUP BY user_id, courserun_id
HAVING COUNT(*) > 1;
```

I also checked this branch, and the same duplicates are still present. So, in short, the refactored query doesn’t appear to address the duplicated issue. Can you confirm?

One way to clear the duplicated records would be to do a full refresh of the raw table, since the deduplication logic in the code likely won’t work in this case. Just my thoughts.

quazi-h:
Thanks @rachellougee — you're correct, and this is a really helpful catch.

The DEDP fan-out refactor (first commit) improved code structure but didn't touch the actual root cause. The real issue is exactly what you identified: hard-deletes followed by re-enrollment produce two rows with different `id`s for the same `(user_id, run_id)` in the raw Iceberg table, which the staging model was passing through unchanged.

Second commit adds `deduplicate_raw_table(order_by='id', partition_columns='user_id, run_id')` to the staging model, keeping the most recently created enrollment per user+run — same pattern as `ecommerce_basket`, `wagtailcore_revision`, etc.

Both models pass all tests in QA (PASS=38). Since QA doesn't have the deleted enrollment data, the dedup logic isn't exercised there — but the fix is correct and a **full refresh of the staging model in production** will be needed after deploy to clear the existing duplicate rows from the materialized table.

quazi-h:
> I originally wrote this DEDP query because the micromasters orders exist in two systems. I looked into the two duplicated records in `int__mitxonline__courserunenrollments` on production, and they actually don’t seem to be DEDP-related. The duplicates are caused by **deleted enrollments** for UAI courses in raw enrollment table.
> 
> ```
> SELECT user_id, courserun_id, COUNT(*) AS cnt
> FROM"ol_data_lake_production"."ol_warehouse_production_intermediate".int__mitxonline__courserunenrollments
> GROUP BY user_id, courserun_id
> HAVING COUNT(*) > 1;
> ```
> 
> I also checked this branch, and the same duplicates are still present. So, in short, the refactored query doesn’t appear to address the duplicated issue. Can you confirm?
> 
> One way to clear the duplicated records would be to do a full refresh of the raw table, since the deduplication logic in the code likely won’t work in this case. Just my thoughts.

I confirmed what you said is correct.

The issue: The raw Iceberg table has two enrollment records for the same (user_id, run_id) because Django hard-deleted an enrollment and re-created it — two different ids, same user+run. The staging model passes both through unchanged. Our DEDP fix improved code quality but didn't touch the staging layer where the duplicates originate.

The fix: Add deduplicate_raw_table to the staging enrollment model, partitioning by (user_id, run_id) and keeping the highest id (most recently created enrollment). Does that seem reasonable?

quazi-h:
Following up with more detail after querying production data directly.

**The duplicates are confirmed as UAI courses (program_id=14), not DEDP.** The DEDP CTEs never touch these records, so the intermediate model refactor was unrelated to the actual bug — it's been reverted in the latest commit.

**On the dedup ordering:** The initial fix used `order_by='id'` (keep highest ID), but production data showed this would produce incorrect results for user 1294262:
- id=966253 (lower): active=True, **verified**, created 2026-05-20 → the live enrollment
- id=967247 (higher): active=False, **audit/unenrolled**, created 2026-05-21 → stale ghost record

Keeping the highest id would have discarded the active verified enrollment in favour of the stale unenrolled one. The fix now uses `order_by='active desc, id'` — preferring `active=True` records first, then highest id as tiebreaker.

**What's in the PR now:** only the staging dedup change. One commit, one file changed.

**A full refresh of the staging model in production is needed after deploy to clear the existing duplicate rows from the materialised table.**

rachellougee:
Unfortunately, the duplicated data in the staging table isn't correct.

-- Returns 966253, but the correct ID should be 967247.
-- 966253 is the deleted record.
```
select courserunenrollment_id from "ol_data_lake_production"."ol_warehouse_production_rlougee_staging".stg__mitxonline__app__postgres__courses_courserunenrollment
where user_id = 1294262 and courserun_id = 3487;
```

The other duplicated record is correct:
```
-- Returns 477165
select courserunenrollment_id from "ol_data_lake_production"."ol_warehouse_production_rlougee_staging".stg__mitxonline__app__postgres__courses_courserunenrollment
where user_id=42170 and courserun_id=533
```

Based on these two examples, the deduplication logic 

`{{ deduplicate_raw_table(order_by="active desc, id", partition_columns="user_id, run_id") }}`

doesn't handle all cases correctly. The underlying issue seems to be that we can't reliably determine which record was deleted from the data available in dbt. Since this information isn't controlled at the dbt layer, my recommendation would be to perform a full refresh in Airbyte to rebuild the staging table with the correct records.

What do you think?

quazi-h:
Thanks Rachel — you were right, and I've pushed a fix.

The root cause of the incorrect result: the hard-deleted enrollment (id=966253) still has `active=True` in the append-only Iceberg table, so `ORDER BY active DESC` incorrectly ranked it above the current live record (id=967247, active=False). The `active` field can't be used as a proxy for "current" when records are hard-deleted rather than soft-deleted.

**Fix:** changed to `order_by='id'` only. Auto-increment PKs are monotonically increasing — the re-enrolled record always receives a higher id than any previously hard-deleted enrollment, so highest id = current live record.

**Validated against your exact queries on production data (my dev schema):**

| user_id | courserun_id | returned id | active | mode | correct? |
|---------|-------------|------------|--------|------|----------|
| 42170 | 533 | 477165 | True | audit | ✅ |
| 1294262 | 3487 | 967247 | False | audit | ✅ |

Both now return the records you expected.

quazi-h:
We are going to proceed with keeping the dedupe logic as:
`{{ deduplicate_raw_table(order_by="id", partition_columns="user_id, run_id") }}`
ORDER BY id DESC alone is sufficient for both cases. Since id is auto-incrementing in Django, a higher id always means a more recent write — the re-enrollment after a hard-delete always gets a higher id than the original row. So ROW_NUMBER() OVER (PARTITION BY user_id, courserun_id ORDER BY id DESC) = 1 correctly picks 477165 over 470891 (and analogously 967247 over its lower-id duplicate) without needing the active flag at all.

rachellougee (approved):

PR #2286, opened 2026-06-09
docs(dbt): document enrollment join grain in int__edxorg__mitx_courserun_certificates

quazi-h:
Closes #2132

Documents that the `enrollment` LEFT JOIN in the DEDP certificate path of `int__edxorg__mitx_courserun_certificates` is safe and cannot produce fan-out rows.

**Changes:**
- Adds a header comment to `int__edxorg__mitx_courserun_certificates.sql` stating the model grain (`one row per (user_id, courserun_readable_id)`) and the full grain analysis for the enrollment LEFT JOIN
- Adds an inline comment on the enrollment LEFT JOIN explaining why joining on `(courserun_readable_id, user_username)` is equivalent to the enrollment model's `(user_id, courserun_readable_id)` unique key and therefore cannot fan out
- Adds an inline comment on the `micromasters_users` LEFT JOIN in the non-DEDP path explaining why `user_edxorg_username` is safe despite not having a standalone uniqueness test
- Updates the model description in `_int_edxorg__models.yml` to document the grain and the join grain analysis

**Grain analysis summary:**
- `int__edxorg__mitx_courserun_enrollments` is unique on `(user_id, courserun_readable_id)` (tested)
- The join condition in the DEDP path is `(courserun_readable_id, user_username)`. Since `user_username` is derived from `int__edxorg__mitx_users` which has a unique test on `user_username`, this maps 1:1 to `(courserun_readable_id, user_id)` — i.e. the enrollment model's unique key — so no fan-out is possible.

### How can this be tested?

Sanity-check queries were run against production data to confirm zero fan-out. A reviewer can validate by running the following against the production Trino/Starburst catalog:

**1. Confirm enrollment model is unique on (user_id, courserun_readable_id):**
```sql
select user_id, courserun_readable_id, count(*)
from ol_data_lake_production.marts_intermediate.int__edxorg__mitx_courserun_enrollments
group by 1, 2
having count(*) > 1
limit 10
```
Expected: 0 rows.

**2. Confirm the certificates model has no fan-out (row count should not change after DISTINCT):**
```sql
select count(*) as total_rows,
       count(distinct (user_id, courserun_readable_id)) as distinct_pairs
from ol_data_lake_production.marts_intermediate.int__edxorg__mitx_courserun_certificates
```
Expected: `total_rows = distinct_pairs`.

**Production results at time of implementation:** 356,972 total rows = 356,972 distinct (user_id, courserun_readable_id) pairs; 7,491 DEDP-path rows confirmed with zero fan-out.

### Additional Context
This PR is purely documentation — no logic changes. It addresses the G5 grain verification task for the edxorg certificates intermediate model. The uniqueness tests on `int__edxorg__mitx_users.user_username` and `int__edxorg__mitx_courserun_enrollments.(user_id, courserun_readable_id)` provide the structural guarantee; these comments make that reasoning explicit for future readers.

### DIFF
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -79,8 +79,13 @@ models:
     description: string, short description of the course run
 
 - name: int__edxorg__mitx_courserun_certificates
-  description: Intermediate model for course run certificates for edx.org courses
-    For edx.org DEDP courses, certificates are sourced from MicroMasters
+  description: "Intermediate model for course run certificates for edx.org courses.\n\
+    For edx.org DEDP courses, certificates are sourced from MicroMasters.\nGrain:\
+    \ one row per (user_id, courserun_readable_id).\nThe LEFT JOIN to int__edxorg__mitx_courserun_enrollments\
+    \ in the DEDP CTE joins on\n(courserun_readable_id, user_username). Since user_username\
+    \ is unique per user_id\nin that model (sourced from int__edxorg__mitx_users),\
+    \ this resolves to the enrollment\nmodel’s (user_id, courserun_readable_id) unique\
+    \ key and cannot fan out."
   columns:
   - name: user_id
     description: int, Numerical user ID of a learner on edx.org
--- a/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
+++ b/src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
@@ -79,13 +79,14 @@ models:
     description: string, short description of the course run
 
 - name: int__edxorg__mitx_courserun_certificates
-  description: "Intermediate model for course run certificates for edx.org courses.\n\
-    For edx.org DEDP courses, certificates are sourced from MicroMasters.\nGrain:\
-    \ one row per (user_id, courserun_readable_id).\nThe LEFT JOIN to int__edxorg__mitx_courserun_enrollments\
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
```diff
@@ -79,8 +79,14 @@ models:
     description: string, short description of the course run
 
 - name: int__edxorg__mitx_courserun_certificates
-  description: Intermediate model for course run certificates for edx.org courses
-    For edx.org DEDP courses, certificates are sourced from MicroMasters
+  description: |
```
KatelynGit, line 82: looks like an accidental "|" character here possiby

KatelynGit (approved):
tests passed

src/ol_dbt/models/intermediate/edxorg/_int_edxorg__models.yml
quazi-h: Good catch! That looks weird to me too but apparently it's proper syntax for this according to copilot:

It's actually intentional and correct YAML — the | is a literal block scalar indicator (https://yaml.org/spec/1.2-old/spec.html#id2795688), which allows the multi-line description to be written cleanly across multiple lines. It was applied to replace the double-quoted string with \n\ / \  escapes that you see in the old code's diff on the left, which is hard to read/edit and is the only instance of this pattern in the dbt schema YAML. Using a YAML block scalar keeps the same content while being more maintainable and less error-prone for YAML parsing/escaping.

PR #2292, opened 2026-06-10
update enrollment to dim tables

KatelynGit:
https://github.com/mitodl/hq/issues/11622

updates enrollment reporting to use dimensional tables

### How can this be tested?
dbt build --select enrollment_detail_report

### DIFF
--- a/src/ol_dbt/models/reporting/enrollment_detail_report.sql
+++ b/src/ol_dbt/models/reporting/enrollment_detail_report.sql
@@ -11,7 +11,8 @@ with enrollment as (
 )
 
 , f_certificate as (
-    select * from {{ ref('f_certificate') }}
+    select * from {{ ref('tfact_certificate') }}
+    where certificate_scope = 'course'
 )
 
 , grade as (
--- a/src/ol_dbt/models/reporting/enrollment_detail_report.sql
+++ b/src/ol_dbt/models/reporting/enrollment_detail_report.sql
@@ -170,8 +170,6 @@ select
     , certificate_created_on as courseruncertificate_created_on
     , certificate_issued_on as courseruncertificate_issued_on
     , case when certificate_issued_on is not null then true else false end as courseruncertificate_is_earned
-    , case when certificate_uuid is not null
-        then concat('https://mitxonline.mit.edu/certificate/', certificate_uuid)
     , case enrollment.platform
         when 'mitxonline' then
             case when certificate_uuid is not null
--- a/src/ol_dbt/models/reporting/enrollment_detail_report.sql
+++ b/src/ol_dbt/models/reporting/enrollment_detail_report.sql
@@ -193,7 +193,7 @@ select
     , discount_type_name
     , order_emails.redeemed_email
     , if(
-        enrollment.platform = '{{ var("mitxonline") }}'
+        enrollment.platform = 'mitxonline'
         , concat('https://mitxonline.mit.edu/orders/receipt/', cast(f_order.order_id as varchar))
         , null
     ) as receipt_url
--- a/src/ol_dbt/models/reporting/enrollment_detail_report.sql
+++ b/src/ol_dbt/models/reporting/enrollment_detail_report.sql
@@ -132,6 +132,15 @@ with enrollment as (
         , user_email
 )
 
+, course_passed_counts as (
+    select
+        user_email
+        , count(distinct course_title) as num_of_course_passed
+    from combined_enrollment_detail
+    where courserungrade_is_passing = true
+    group by user_email
+)
+
 select
     case enrollment.platform
         when 'bootcamps' then '{{ var("bootcamps") }}'
@@ -177,6 +186,7 @@ select
         else null
       end as user_username
     , d_user.email as user_email
+    , course_passed_counts.num_of_course_passed
     , discount.discount_code as coupon_code
     , discount_names.discount_name as coupon_name
     , discount_amount as discount
@@ -272,3 +282,5 @@ left join order_emails
     end = order_emails.platform
     and course_run.courserun_readable_id = order_emails.courserun_readable_id
     and d_user.email = order_emails.user_email
+left join course_passed_counts
+    on d_user.email = course_passed_counts.user_email
--- a/src/ol_dbt/models/reporting/enrollment_detail_report.sql
+++ b/src/ol_dbt/models/reporting/enrollment_detail_report.sql
@@ -113,7 +113,16 @@ with enrollment as (
 )
 
 select
-    enrollment.platform
+    case enrollment.platform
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/reporting/enrollment_detail_report.sql
```diff
@@ -1,121 +1,263 @@
...
-            and program_enrollments.program_name = coursesinprogram.program_name
-            and program_enrollments.program_id = coursesinprogram.program_id
-    where coursesinprogram.program_name <> 'Computer Science'
-        and (programenrollment_enrollment_status is null or trim(programenrollment_enrollment_status)= '')
-)
-
-, course_passed_counts as (
+        order_id
+        , platform
+        , redeemed_email
```
quazi-h, line 108:
```suggestion
        , courserun_readable_id
        , user_email
        , redeemed_email
```
The final join below references [order_emails.courserun_readable_id](https://github.com/mitodl/ol-data-platform/pull/2292/changes#diff-bafa4358660041608111ab5487e776cd2a527e5d62705ad7b454be99e8566b37R262) and [user_email](https://github.com/mitodl/ol-data-platform/pull/2292/changes#diff-bafa4358660041608111ab5487e776cd2a527e5d62705ad7b454be99e8566b37R263), so I think those field needs to be included here or it will fail to compile.

src/ol_dbt/models/reporting/enrollment_detail_report.sql
```diff
@@ -1,121 +1,263 @@
...
+    on payment.user_fk = d_user_payer.user_pk
+left join program
+    on enrollment.program_fk = program.program_pk
+left join discount_type
+    on f_order.discount_type_fk = discount_type.discount_type_pk
+left join discount_names
+    on discount.discount_code = discount_names.discount_code
+left join enrollment_upgrades
+    on f_order.order_id = enrollment_upgrades.order_id
+    and enrollment.platform = enrollment_upgrades.platform
```
quazi-h:
```suggestion
     and case enrollment.platform
        when 'bootcamps' then '{{ var("bootcamps") }}'
        when 'edxorg' then '{{ var("edxorg") }}'
        when 'emeritus' then '{{ var("emeritus") }}'
        when 'global_alumni' then '{{ var("global_alumni") }}'
        when 'mitxonline' then '{{ var("mitxonline") }}'
        when 'mitxpro' then '{{ var("mitxpro") }}'
        when 'residential' then '{{ var("residential") }}'
        else enrollment.platform
    end = enrollment_upgrades.platform
```
I think the platform field coming from these two data sources does not match up so we have to reference the right name for this join to work properly:

enrollment (tfact_enrollment.platform = 'mitxonline')
enrollment_upgrades (marts__combined_course_enrollment_detail.platform = 'MITx Online')

quazi-h (changes requested):

src/ol_dbt/models/reporting/enrollment_detail_report.sql
```diff
@@ -1,121 +1,286 @@
...
+        order_id
+        , platform
+        , courserun_readable_id
+        , user_email
+        , redeemed_email
+    from combined__orders
+    group by
+        order_id
+        , platform
+        , redeemed_email
```
quazi-h, line 115:
```suggestion
        , courserun_readable_id
        , user_email
        , redeemed_email
```

Sorry, I should've caught this in my last review, adding them to the select would also require they be added to the group by clause (non-aggregated columns in SELECT must be in GROUP BY)

src/ol_dbt/models/reporting/enrollment_detail_report.sql
```diff
@@ -1,121 +1,288 @@
...
+    , enrollment_mode as courserunenrollment_enrollment_mode
+    , enrollment_status as courserunenrollment_enrollment_status
+    , enrollment_is_active as courserunenrollment_is_active
+    , enrollment_upgrades.courserunenrollment_upgraded_on
+    , certificate_created_on as courseruncertificate_created_on
+    , certificate_issued_on as courseruncertificate_issued_on
+    , case when certificate_issued_on is not null then true else false end as courseruncertificate_is_earned
+    , case when certificate_uuid is not null
+        then concat('https://mitxonline.mit.edu/certificate/', certificate_uuid)
+        else null end as courseruncertificate_url
```
quazi-h:
```suggestion
    , case enrollment.platform
        when 'mitxonline' then
            case when certificate_uuid is not null
            then concat('https://mitxonline.mit.edu/certificate/', certificate_uuid)
            else null end
        when 'mitxpro' then
            case when certificate_uuid is not null
            then concat('https://xpro.mit.edu/certificate/', certificate_uuid)
            else null end
        when 'bootcamps' then
            case when certificate_uuid is not null
            then concat('https://bootcamps.mit.edu/certificate/', certificate_uuid)
            else null end
        else null
    end as courseruncertificate_url
```
The URL is hardcoded to mitxonline.mit.edu/certificate/ for all platforms. mitxpro and bootcamps have different base URLs and valid certificate UUIDs in tfact_certificate, so this will produce wrong URLs for those platforms

Both mitxpro and bootcamps have real certificate UUIDs in tfact_certificate. With the current code, those ~56k certificates would get https://mitxonline.mit.edu/certificate/<uuid> — pointing to the wrong site entirely. This should probably be upgraded from nice-to-have to should fix before merging.

quazi-h (commented):
Changes look good so far, just one more thing, the hardcoded course run certificate url just needs some conditional logic so we get the right url for mitxpro and bootcamps. I included a code suggestion but you can do that however you see fit.

src/ol_dbt/models/reporting/enrollment_detail_report.sql
```diff
@@ -1,121 +1,302 @@
...
+    , enrollment_created_on as courserunenrollment_created_on
+    , enrollment_mode as courserunenrollment_enrollment_mode
+    , enrollment_status as courserunenrollment_enrollment_status
+    , enrollment_is_active as courserunenrollment_is_active
+    , enrollment_upgrades.courserunenrollment_upgraded_on
+    , certificate_created_on as courseruncertificate_created_on
+    , certificate_issued_on as courseruncertificate_issued_on
+    , case when certificate_issued_on is not null then true else false end as courseruncertificate_is_earned
+    , case when certificate_uuid is not null
+        then concat('https://mitxonline.mit.edu/certificate/', certificate_uuid)
```
quazi-h:
```suggestion
```
Whoops, these two lines should have been removed since now we're including that "certificate_uuid is not null" in all the cases below.  That was my fault, I was trying to highlight these two lines as well to be replaced by the new change but I think I did the code suggestion wrong. Might want to just do it on your end instead of using it here

quazi-h (approved):
Looks great!

PR #2315, opened 2026-06-17
fix(edxorg-s3): fix dlt resource naming, incremental loading, and schema metadata

blarghmatey:
N/A

Fixes three related bugs in the EdX.org S3 dlt ingest pipeline.

**Root cause: wrong destination table name**

A `@dlt.resource` wrapper whose function body *returns* another `DltResource` causes dlt to discard the outer resource entirely — its `name`, `write_disposition`, `primary_key`, and `table_format` are all dropped. Every table was therefore being written to a table named `read_csv_duckdb` (the transformer's own name). The schema and job lookups in `DagsterDltResource.extract_resource_metadata` searched for `raw_edxorg_s3_tables_*` and found nothing, producing:

- `jobs: []`
- `dagster/column_schema: { "columns": [] }`

**`loads.py` changes**

- Replace the `@dlt.resource` wrapper + `return pipe` antipattern with `.with_name()` + `.apply_hints()` applied directly on the `filesystem | read_csv_duckdb` pipe. This is the documented way to attach hints to a transformer-based resource.
- Pass `incremental=incremental("modification_date")` natively to `filesystem()` (it has a first-class `incremental` parameter) rather than via `apply_hints()`. Each per-table resource gets its own `modification_date` cursor keyed by resource name, so subsequent runs only process files whose modification date is newer than the last successful run.
- Fix `table_format`: was always `"iceberg"` even for `dev`/`local` environments where there is no `[iceberg_catalog.local]` stanza in `config.toml`. Now conditional: `"iceberg"` for `qa`/`production`, `"parquet"` otherwise.

**`dagster_assets.py` changes**

- Add `loader_file_format="parquet"` to every `dlt.run()` call to match the standalone `run_pipeline()` and produce the correct staging format for filesystem+Iceberg destinations.
- Set `storage_kind` and `kinds` to `"iceberg"` (qa/prod) or `"filesystem"` (local) rather than the named destination name (`"production"`). `data.destination.destination_name` returns the named destination, not the storage type, so catalog badges were misleading.
- Explicitly compute `dagster/table_name` from `pipeline.dataset_name` because dlt does not populate `schema_name` in `load_info` for filesystem+Iceberg destinations, causing `extract_resource_metadata` to always leave `table_name=None`.

### How can this be tested?

1. Trigger a materialization of one or more `edxorg_s3_consolidated_tables` assets in the Dagster UI.
2. After the run completes, verify in the asset metadata:
   - `dagster/column_schema` has a non-empty `columns` list reflecting the TSV file schema.
   - `jobs` contains at least one entry with a matching `table_name`.
   - `dagster/storage_kind` shows `iceberg` (not `production`).
3. Trigger a second materialization immediately after. Verify the run completes quickly (no files to process) — the `modification_date` incremental cursor should prevent reprocessing unchanged files.
4. Locally: `DAGSTER_ENVIRONMENT=dev python -m data_loading.defs.edxorg_s3_ingest.loads` should write Parquet (not Iceberg) without error.

### DIFF
--- a/dg_projects/data_loading/data_loading/defs/edxorg_s3_ingest/dagster_assets.py
+++ b/dg_projects/data_loading/data_loading/defs/edxorg_s3_ingest/dagster_assets.py
@@ -63,7 +63,7 @@ class EdxorgDltTranslator(DagsterDltTranslator):
             metadata={
                 **default_spec.metadata,
                 **table_name_meta,
-                **TableMetadataSet(storage_kind=_STORAGE_KIND),
+                **dict(TableMetadataSet(storage_kind=_STORAGE_KIND)),
             },
         )
 
--- a/dg_projects/data_loading/data_loading/defs/edxorg_s3_ingest/dagster_assets.py
+++ b/dg_projects/data_loading/data_loading/defs/edxorg_s3_ingest/dagster_assets.py
@@ -12,44 +12,59 @@ from dagster import (
     AssetKey,
     AssetSpec,
 )
+from dagster._core.definitions.metadata.metadata_set import TableMetadataSet
 from dagster_dlt import DagsterDltResource, DagsterDltTranslator, dlt_assets
 from dagster_dlt.translator import DltResourceTranslatorData
 from ol_orchestrate.lib.constants import EDXORG_DB_TABLES
 
 from .loads import (
+    dagster_env,
     edxorg_s3_pipeline,
     edxorg_s3_source,
     edxorg_s3_source_instance,
     table_format,
 )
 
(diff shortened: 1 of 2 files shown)


blarghmatey (commented):
Both issues addressed in 57676f81:

- `apply_hints` now includes `table_name=resource_name` explicitly, so the destination table name is a concrete hint rather than a `with_name()` fallback.
- `TableMetadataSet(storage_kind=...)` is now wrapped in `dict()` before spreading, matching the pattern used for `table_name_meta`.

PR #2318, opened 2026-06-17
add problem data to instructor reporting

KatelynGit:
https://github.com/mitodl/hq/issues/11464

add problem data to new instructor reporting table

### How can this be tested?
dbt build --select instructor_module_problems_report

### DIFF
--- a/src/ol_dbt/models/reporting/_reporting__models.yml
+++ b/src/ol_dbt/models/reporting/_reporting__models.yml
@@ -1369,3 +1369,31 @@ models:
   - name: user_latest_row_desc
     description: integer, row number indicating the latest row for each user based
       on application and course run dates
+
+- name: instructor_module_problems_report
+  description: A report showing learner problem data for later roll-up at the instructor level.
+  columns:
+  - name: user_email
+    description: string, email address of the user
+  - name: full_name
+    description: string, full name of the user
+  - name: courserun_readable_id
+    description: string, unique identifier for the course run formatted as course-v1:{org}+{course code}+{run_tag}
+  - name: section_title
+    description: string, title of the chapter (section) block within the course
+  - name: subsection_title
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/reporting/_reporting__models.yml
```diff
@@ -1370,3 +1370,33 @@ models:
   - name: user_latest_row_desc
     description: integer, row number indicating the latest row for each user based
       on application and course run dates
+
+- name: instructor_module_problems_report
```
rachellougee, line 1373: Can we add some tests in this new model? At minimum: not_null on user_email, courserun_readable_id, and problem_name. A unique combined key to prevent fan-out.

src/ol_dbt/models/reporting/instructor_module_problems_report.sql
```diff
@@ -0,0 +1,88 @@
...
+from enrollment
+inner join user
+    on enrollment.user_fk = user.user_pk
+inner join course_run
+    on enrollment.courserun_fk = course_run.courserun_pk
+inner join problem
+    on
+        course_run.courserun_readable_id = problem.courserun_readable_id
+        and course_run.platform = problem.platform
+left join problem_events
```
rachellougee, line 61: This left join produced a lot of empty rows where attempt, success, and grade are all null have no actionable value

rachellougee (approved):
dbt runs fine, I left two comments, both are worth addressing before merging

src/ol_dbt/models/reporting/instructor_module_problems_report.sql
KatelynGit: they want to see when learners don't complete problems- however I agree it's too many records I'm checking with the requestor- waiting for a reply

PR #2336, opened 2026-06-24
feat: auto-repair Iceberg snapshot pointer lag from Airbyte full-refresh pin bug

blarghmatey:

N/A — this is a self-contained reliability improvement discovered during incident investigation on 2026-06-24.

Adds automatic detection and repair of an Airbyte Iceberg bug where `current-snapshot-id` gets frozen after a full-refresh and subsequent incremental syncs stop becoming visible to Trino/dbt.

**The bug:** When an Airbyte connection is reset and performs a full-refresh, it creates a large snapshot and pins `current-snapshot-id` to it. Subsequent incremental syncs append new snapshots to the Iceberg history but do not advance the pointer. Trino and dbt always read the `current-snapshot-id` snapshot, so all downstream models see stale data even though newer records exist in the history.

**Observed impact:** On 2026-06-24, this affected 9 xPRO raw tables (`ecommerce_order`, `ecommerce_line`, `ecommerce_receipt`, `ecommerce_couponredemption`, `ecommerce_coupon`, `b2becommerce_b2border`, `b2becommerce_b2borderline`, `b2becommerce_contractpage`, `ecommerce_couponpaymentversion`). The pointer was frozen at the June 17 snapshot while 14 newer snapshots had accumulated — causing xPRO order data in `marts__combined__orders` to appear stale despite successful dbt rebuilds.

**New components:**

- **`packages/ol-orchestrate-lib/src/ol_orchestrate/lib/iceberg_maintenance.py`**
  - `SnapshotPointerLag` dataclass — represents a table with a stale pointer, with a `lag_hours` property
  - `find_snapshot_pointer_lag(glue_database, table_prefixes, min_lag_hours)` — scans Iceberg tables matching given prefixes and returns those where `current-snapshot-id` is not the latest snapshot and the gap exceeds the threshold
  - `advance_snapshot_pointer(glue_database, table_name)` — repairs a lagging table by writing a new Iceberg metadata JSON with the pointer advanced to the latest snapshot and updating the Glue `metadata_location` parameter

- **`dg_projects/lakehouse/lakehouse/sensors.py`** (new file)
  - `repair_lagging_snapshot_pointers` op — re-discovers lagging tables at run time and calls `advance_snapshot_pointer` for each; raises at the end if any failed (triggering the Slack run-failure alert)
  - `iceberg_snapshot_pointer_repair` job — wraps the op
  - `iceberg_snapshot_pointer_lag` sensor — runs hourly, scans the high-frequency Airbyte groups (xPRO, MITx Online, MIT Learn, Learn AI, OCW), yields a `RunRequest` to the repair job when new lag is detected; cursor-based deduplication prevents repeated repair runs for the same batch of stuck tables

- **`dg_projects/lakehouse/lakehouse/definitions.py`** — registers the sensor and repair job

### How can this be tested?

1. Deploy to QA and confirm the `iceberg_snapshot_pointer_lag` sensor appears in the Dagster UI in `RUNNING` state.
2. To simulate the bug: use `aws glue update-table` to set `metadata_location` on a raw table to point at an older metadata JSON (one where `current-snapshot-id` is behind the latest snapshot). Wait for the next sensor tick (≤ 1 hour) and confirm a `iceberg_snapshot_pointer_repair` run is triggered and completes successfully.
3. After the repair run, verify with `aws glue get-table --database-name ol_warehouse_production_raw --name <table>` that `metadata_location` now points at the new metadata file and `current-snapshot-id` matches the latest snapshot.

In production, the sensor will fire automatically if the condition recurs.

### Additional Context

The `advance_snapshot_pointer` repair is safe and additive — it writes a new metadata JSON file (UUID-named) rather than modifying the existing one, so the prior state is preserved. Glue is updated atomically via a single `update_table` call pointing at the new file.

The sensor scans only tables in `RAW_LAYER_GROUP_CONFIGS` prefixes (excluding `_default`) to limit Glue API calls. Low-frequency sync groups are excluded because pointer lag there is less urgent and would add noise.

### DIFF
--- a/dg_projects/lakehouse/lakehouse/sensors.py
+++ b/dg_projects/lakehouse/lakehouse/sensors.py
@@ -33,17 +33,6 @@ _RAW_DATABASE_BY_ENV: dict[str, str] = {
 }
 RAW_DATABASE = _RAW_DATABASE_BY_ENV.get(DAGSTER_ENV, "ol_warehouse_production_raw")
 
-# Explicit list of high-frequency (6 h) Airbyte source groups to scan.
-# The thirdparty groups (salesforce, zendesk) sync much less often and are
-# deliberately excluded to keep the sensor fast and the alert signal clean.
-_HIGH_FREQUENCY_PREFIXES = [
-    "raw__mitxonline__app",
-    "raw__xpro__app",
-    "raw__mitlearn__app",
-    "raw__learn_ai__app",
-    "raw__ocw__studio",
-]
-
 # Alert when the pointer lags behind the latest snapshot by more than this.
 # Two hours provides a buffer for the normal write window without masking a
 # genuine stuck-pointer condition (which gaps by days, not minutes).
@@ -61,7 +50,6 @@ def repair_lagging_snapshot_pointers(context):
     """
     lagging = find_snapshot_pointer_lag(
         glue_database=RAW_DATABASE,
-        table_prefixes=_HIGH_FREQUENCY_PREFIXES,
         min_lag_hours=_MIN_LAG_HOURS,
     )
 
@@ -144,7 +132,6 @@ def iceberg_snapshot_pointer_lag_sensor(
 ) -> SensorResult:
     lagging = find_snapshot_pointer_lag(
         glue_database=RAW_DATABASE,
-        table_prefixes=_HIGH_FREQUENCY_PREFIXES,
         min_lag_hours=_MIN_LAG_HOURS,
     )
 
--- a/dg_projects/lakehouse/lakehouse/sensors.py
+++ b/dg_projects/lakehouse/lakehouse/sensors.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import datetime
 import json
 import logging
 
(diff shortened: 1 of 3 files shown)


dg_projects/lakehouse/lakehouse/sensors.py
```diff
@@ -0,0 +1,175 @@
...
+)
+def iceberg_snapshot_pointer_lag_sensor(
+    context: SensorEvaluationContext,
+) -> SensorResult:
+    lagging = find_snapshot_pointer_lag(
+        glue_database=RAW_DATABASE,
+        min_lag_hours=_MIN_LAG_HOURS,
+    )
+
+    cursor_data: dict[str, list[str]] = json.loads(context.cursor or "{}")
```
quazi-h, line 138: The  or "{}"  guard only protects against  None /empty string — it doesn't protect against a non-JSON cursor value, which causes an unhandled  JSONDecodeError . Dagster cursors can end up corrupted if a previous sensor run crashes mid-write. A try catch here for json.JSONDecodeError would be good.

rachellougee (approved):
This seems to be working for me. I did an end-to-end test by introducing lag on one of my staging table's snapshots in my schema. The sensor detected the lag, launched the repair job, and the job advanced current-snapshot-id  from the stale snapshot to the latest snapshot successfully.    

Image

PR #2408, opened 2026-07-08
 fix(dbt): fix missing and duplicate order_id in enrollment_detail_report

rachellougee:

duplicate order issues in https://github.mit.edu/mitxonline/mitxonline-issues/issues/1426
missing order_id for xpro program enrollment in https://github.mit.edu/xpro/xpro-issues/issues/688 
duplicate enrollment in https://github.com/mitodl/hq/issues/12189

Fixes two issues after we switched to use dimensional models for enrollment_detail_report:   

-   Missing order_id for xPRO program-order enrollments (tfact_enrollment.sql):                                   
-   Duplicate rows when an enrollment matches more than one order (enrollment_detail_report.sql), affecting any platform (confirmed on both xPRO and mitxonline)                                                                       

### How can this be tested?

`dbt run --select +tfact_enrollment +tfact_order enrollment_detail_report --full-refresh `

- Confirmed that this query returns order_id= 81192 instead of blank order_id for xpro.

```
select order_id, unit_price,program_name
from "ol_data_lake_production".ol_warehouse_production_rlougee_reporting.enrollment_detail_report
where courserunenrollment_id='69514' and platform='xPro'

```

- Confirmed that this query returns only 1 row instead of duplicated rows for mitxonline. This addresses the first issue in https://github.mit.edu/mitxonline/mitxonline-issues/issues/1426

```
select  *
from "ol_data_lake_production".ol_warehouse_production_reporting.enrollment_detail_report
where courserunenrollment_id='998362' and platform='MITx Online'
```

### DIFF
--- a/src/ol_dbt/models/dimensional/_fact_tables.yml
+++ b/src/ol_dbt/models/dimensional/_fact_tables.yml
@@ -53,7 +53,15 @@ models:
         to: ref('dim_course_run')
         field: courserun_pk
   - name: program_fk
-    description: Foreign key to dim_program (null for course enrollments)
+    description: >
+      Foreign key to dim_program. Null for most course enrollments (courserun_fk is
+      the relevant key there), but populated for mitxpro course enrollments (enrollment_type
+      = 'course') that resulted from a program-purchase order, since xPRO creates
+      one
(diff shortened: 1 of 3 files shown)


KatelynGit (approved):
LGTM

PR #2410, opened 2026-07-09
fix(dimensional): coalesce nullable SCD2 change-detection keys

### LINKED ISSUE #2380: Fix NULL-unsafe SCD2 change detection in dim_course_run and dim_product
### Expected Behavior

SCD2 change detection treats a NULL attribute as unchanged when it was NULL before, producing a bounded number of versions per natural key.

### Current Behavior

`dim_course_run.sql:214` compares `existing.courserun_title = new.courserun_title` **without** `coalesce`, while all sibling attributes at lines 215-219 are coalesced. `courserun_title` is nullable (no `not_null` test), so `NULL = NULL` evaluates to UNKNOWN and every null-title run is treated as "changed" on **every** incremental run — creating a new current row and expiring the old one, producing unbounded version churn.

The same pattern exists in `dim_product.sql` on `product_price`.

### Possible Solution

Coalesce to a sentinel like the other columns. Consider a data test asserting one current row per natural key and bounded version growth.

---
**P1.** Part of #2375 (2026-07 dbt warehouse audit). Witan task: `tk-fix-null-unsafe-scd2-change-detection-in-dim-cou-aaf192`.

blarghmatey:
## What

Fixes NULL-unsafe SCD2 change detection in `dim_course_run` and `dim_product`.

- `dim_course_run.sql` compared `existing.courserun_title = new.courserun_title` **without** `coalesce`, while every sibling attribute in the same `not exists` block was coalesced. `courserun_title` is nullable (no `not_null` test), so `NULL = NULL` evaluates to `UNKNOWN`, the `not exists` guard never matches for null-title runs, and every incremental run inserts a fresh current row and expires the prior one → unbounded SCD2 version churn.
- Same pattern on `product_price` in `dim_product.sql` (nullable, no `not_null` test).

## Fix

Coalesce both comparisons to a sentinel matching the surrounding columns (`''` for the title, `-1` for the numeric price).

Both models already carry a `dbt_utils.unique_combination_of_columns` test filtered to `is_current = true`, which asserts one current row per natural key — so the current-row invariant is already guarded; this fix stops the historical-version churn.

Closes #2380

### DIFF
--- a/src/ol_dbt/models/dimensional/dim_course_run.sql
+++ b/src/ol_dbt/models/dimensional/dim_course_run.sql
@@ -277,7 +277,7 @@ with micromasters_courseruns as (
             existing.courserun_readable_id = courseruns_with_all_fks.courserun_readable_id
             and existing.platform = courseruns_with_all_fks.platform
             and existing.is_current = true
-            and existing.courserun_title = courseruns_with_all_fks.courserun_title
+            and coalesce(existing.courserun_title, '') = coalesce(courseruns_with_all_fks.courserun_title, '')
             and coalesce(existing.courserun_start_on, '') = coalesce(courseruns_with_all_fks.courserun_start_on, '')
             and coalesce(existing.courserun_end_on, '') = coalesce(courseruns_with_all_fks.courserun_end_on, '')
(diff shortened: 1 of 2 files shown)


rachellougee (approved):
LGTM

PR #2413, opened 2026-07-09
fix(dimensional): dedupe MicroMasters/edxorg certs and fix MM email-case join

### LINKED ISSUE #2381: Dedupe MicroMasters certificates in tfact_certificate and fix email-case user resolution
### Expected Behavior

A MicroMasters certificate earned on edX.org produces exactly one row in `tfact_certificate`, and MicroMasters users resolve to a `user_fk` regardless of email case.

### Current Behavior

`tfact_certificate` unions `int__edxorg__mitx_courserun_certificates` (lines 45-61) and `int__micromasters__course_certificates` (lines 63-79). MicroMasters certs are earned on edX.org and both rows resolve `courserun_fk` to the same edxorg run (lines 247-252). The surrogate keys differ (user_id-based vs email-based), so the defensive dedup **cannot** catch the double-count.

Additionally, lines 241-243 join raw MicroMasters `user_email` against `dim_user.email`, which is `lower()`-ed — so mixed-case MM emails silently fail `user_fk` resolution (and the warn-severity test hides it).

### Possible Solution

- Dedupe on `(user, courserun)` across sources with a documented precedence (or drop one source).
- `lower()` the fact-side email before joining.

### Additional Details

`dim_program` has the same micromasters-vs-edxorg duality — check it alongside this fix.

---
**P1.** Part of #2375 (2026-07 dbt warehouse audit). Witan task: `tk-dedupe-micromasters-certificates-in-tfact-certif-21659d`.

blarghmatey:
## Summary
- MicroMasters course certificates are earned on and issued by edX.org, so `int__edxorg__mitx_courserun_certificates` and `int__micromasters__course_certificates` both produce a row for the same (user, course run) pair in `tfact_certificate` — the micromasters row's `dim_course_run` join is explicitly remapped onto the `edxorg` platform. Their surrogate `certificate_id`s differ (user_id-based vs email-based), so the existing `certificate_key`-based defensive dedup couldn't catch the double-count.
- Added a `cross_source_deduped` CTE that collapses to one row per `(user_fk, courserun_fk, certificate_scope)` whenever both FKs resolved, preferring the canonical `edxorg` record — this only affects the edxorg/micromasters overlap since every other platform resolves to a distinct `courserun_fk`.
- Fixed a silent user-resolution bug: the micromasters user_email join compared raw-case email against `dim_user.email`, which is always `lower()`-ed. Mixed-case MM emails failed to resolve `user_fk` (masked by the warn-severity relationships test).
- Documented both behaviors in `_fact_tables.yml`.

## Test plan
- [x] `dbt compile --select tfact_certificate` succeeds against the `dev_local` duckdb target
- [x] `pre-commit` (sqlfluff, yamllint, etc.) passes on changed files
- [ ] CI dbt build/test run

Fixes #2381

### DIFF
--- a/src/ol_dbt/models/dimensional/tfact_certificate.sql
+++ b/src/ol_dbt/models/dimensional/tfact_certificate.sql
@@ -241,8 +241,10 @@ with mitxonline_certificates as (
     left join user_lookup as ul_micromasters
         on combined_certificates.platform = 'micromasters'
         -- dim_user.email is always lower()-ed; micromasters emails are not,
-        -- so a mixed-case email would otherwise silently fail to resolve user_fk
-        and lower(combined_certificates.user_email) = ul_micromasters.email
+        -- so a mixed-case email would otherwise silently fail to resolve user_fk.
+        -- lower() both sides defensively since it's unclear from this query alone
+        -- that dim_user.email is guaranteed lowercase.
+        and lower(combined_certificates.user_email) = lower(ul_micromasters.email)
     left join user_lookup as ul_bootcamps
         on combined_certificates.platform = 'bootcamps'
         and combined_certificates.user_id = ul_bootcamps.bootcamps_application_user_id
--- a/src/ol_dbt/models/dimensional/tfact_certificate.sql
+++ b/src/ol_dbt/models/dimensional/tfact_certificate.sql
@@ -240,7 +240,9 @@ with mitxonline_certificates as (
         and combined_certificates.user_id = ul_edxorg.edxorg_openedx_user_id
     left join user_lookup as ul_micromasters
         on combined_certificates.platform = 'micromasters'
-        and combined_certificates.user_email = ul_micromasters.email
+        -- dim_user.email is always lower()-ed; micromasters emails are not,
+        -- so a mixed-case email would otherwise silently fail to resolve user_fk
+        and lower(combined_certificates.user_email) = ul_micromasters.email
     left join user_lookup as ul_bootcamps
         on combined_certificates.platform = 'bootcamps'
         and combined_certificates.user_id = ul_bootcamps.bootcamps_application_user_id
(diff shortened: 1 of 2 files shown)


src/ol_dbt/models/dimensional/tfact_certificate.sql
```diff
@@ -240,7 +240,9 @@ with mitxonline_certificates as (
         and combined_certificates.user_id = ul_edxorg.edxorg_openedx_user_id
     left join user_lookup as ul_micromasters
         on combined_certificates.platform = 'micromasters'
-        and combined_certificates.user_email = ul_micromasters.email
+        -- dim_user.email is always lower()-ed; micromasters emails are not,
+        -- so a mixed-case email would otherwise silently fail to resolve user_fk
+        and lower(combined_certificates.user_email) = ul_micromasters.email
```
rachellougee:
```suggestion
       and lower(combined_certificates.user_email) = lower(ul_micromasters.email)
```
The comment refers to dim_user.email, but it's unclear which side is sourced from `dim_user`. There is no harm in converting both sides. But up to you.

rachellougee (approved):
The duplicates seem resolved, but it requires a `--full-refresh` for deployment. Otherwise historical duplicate would stick around.

src/ol_dbt/models/dimensional/tfact_certificate.sql
blarghmatey: Fixed in 9e46762a: now lower() both sides of the join since it's unclear from this query alone that dim_user.email is guaranteed lowercase, per your suggestion.

blarghmatey:
Addressed review feedback: lowered both sides of the MicroMasters email join per @rachellougee's suggestion (9e46762a) — thread resolved. Copilot/Sentry's incremental-run WHERE-clause-ordering flag was already fixed in d26564b3 before this pass. Re: the `--full-refresh` deployment note — noted, will run this migration with `--full-refresh` so historical MicroMasters/edxorg duplicates from before the dedupe logic don't stick around.

PR #2414, opened 2026-07-09
fix(dimensional): fix dim_instructor name-collision identity collapse

### LINKED ISSUE #2377: Fix dim_instructor name-collision identity collapse
### Expected Behavior

Two distinct instructors who share a common name on the same platform resolve to two distinct `instructor_pk` values.

### Current Behavior

`src/ol_dbt/models/dimensional/dim_instructor.sql:46-56` dedupes on `(instructor_name, platform)` using `max()` over the bio columns to collapse to one `instructor_pk` per name+platform. Two distinct instructors sharing a common name on the same platform get silently merged, and the surviving bio text is whichever sorts alphabetically highest — not a deliberate "latest" pick.

This is the same identity-collapse failure mode as `dim_user`'s NULL-email collapse (#/witan `tk-fix-dim-user-null-email-identity-collapse-244cd0`), triggered by a name collision instead of a NULL join key. `_course_catalog_dimensions.yml:5-8` documents the grain as intentional, so this is a real **design gap** — not just a doc fix.

### Possible Solution

Decide deliberately: add a natural key beyond name (e.g. a source instructor id), or explicitly accept and document the collision risk. Also spot-check:

- `dim_program.sql:135-146` and `dim_contract.sql:29-31` — left-join on cast/case-mapped platform keys with no fallback/error path if the mapping misses a case.
- `_dim_program.yml` is missing a `relationships` test on `platform_fk` (only `published_date_key` has one).

---
**P1.** Part of #2375 (2026-07 dbt warehouse audit). Witan task: `tk-fix-dim-instructor-name-collision-identity-colla-ad8db6`.

blarghmatey:
## Summary
- `dim_instructor` deduped on `(instructor_name, platform)` alone, silently merging two distinct instructors who happen to share a name on the same platform into one row, with the surviving bio text picked by alphabetical `max()` rather than any deliberate precedence.
- mitxonline and ocw both have a stable per-instructor natural key in their source data that wasn't being surfaced: `wagtail_page_id` (the CMS page identifying the instructor's bio) for mitxonline, and `course_instructor_uuid` for ocw (already present in `int__ocw__course_instructors`, just not selected downstream).
- Threaded `instructor_source_id` through `int__mitxonline__course_instructors` and `dim_instructor`, and changed the dedupe grain to `coalesce(instructor_source_id, instructor_name)`. Updated `bridge_course_instructor`'s joins to match — they previously joined `dim_instructor` by name alone and would otherwise fan out once the same name can map to more than one `instructor_pk`.
- mitxpro's CMS faculty entries have no stable id in source (just a name + description in a JSON array), so that platform keeps deduping on name alone — documented as an accepted, known limitation rather than fixed, since there's no natural key to recover it from.
- Updated `_course_catalog_dimensions.yml` to document the new grain and the mitxpro exception.

## Not addressed here
The task also flagged `dim_program.sql:135-146` and `dim_contract.sql:29-31` for spot-checking platform-mapping fallback paths, and a missing `relationships` test on `dim_program`'s `platform_fk` — left as follow-up, out of scope for this identity-collapse fix.

## Test plan
- [x] `dbt parse` (whole project) and `dbt compile --select dim_instructor bridge_course_instructor int__mitxonline__course_instructors marts__ocw_courses` succeed against the `dev_local` duckdb target
- [x] `pre-commit` (sqlfluff, yamllint, etc.) passes on changed files
- [ ] CI dbt build/test run

Fixes #2377

### DIFF
--- a/src/ol_dbt/models/dimensional/_course_catalog_dimensions.yml
+++ b/src/ol_dbt/models/dimensional/_course_catalog_dimensions.yml
@@ -4,14 +4,24 @@ version: 2
 models:
 - name: dim_instructor
   description: Instructor dimension containing faculty and course instructors from
-    all platforms. Grain is (instructor_name, primary_platform) — same name on different
-    platforms are kept as separate rows.
+    all platforms. Grain is (instructor_source_id, or instructor_name when no source
+    id exists, primary_platform). mitxonline (wagtail_page_id) and ocw (course_instructor_uuid)
+    have a stable per-instructor natural key and dedupe on it, so two distinct instructors
+    sharing a name on the same platform are kept as separate rows. mitxpro's source
+    data has no such key; its instructors still dedupe on name alone, so two distinct
+    mitxpro faculty sharing an exact name will collapse into a single row — an accepted,
+    documented limitation rather than a fix.
   columns:
   - name: instructor_pk
-    description: Primary key - surrogate key generated from (instructor_name, primary_platform)
+    description: Primary key - surrogate key generated from (instructor_source_id,
+      or instructor_name when no source id exists, primary_platform)
     tests:
     - unique
     - not_null
+  - name: instructor_source_id
+    description: Platform-native stable identifier for the instructor (wagtail_page_id
+      for mitxonline, course_instructor_uuid for ocw). Null for mitxpro, which has
+      no such identifier in its source data.
   - name: instructor_name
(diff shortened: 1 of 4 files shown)


rachellougee (approved):

rachellougee (approved):
Looks like the name collisions from OCW are addressed

Issue #2407, opened 2026-07-08
Add ol-dbt diff command to QA migrated mart/reporting models against pre-migration output


quazi-h:
### Description/Context
Part of #2072 (Migrate mart and reporting models to use the dimensional layer). As more sub-issues under this epic move from triage into actual dimensional-layer migrations, we need a repeatable, field-level QA safeguard to confirm a migrated model produces identical output to the model it replaces, before it's deployed to production.

Requirements:
- Build the pre-migration and post-migration versions of a model side-by-side against real production data, with zero manual data copying.
- Compare output with field-level precision (not just row counts or spot checks).
- Usable by any agent/contributor working a migration ticket under #2072, not a one-off local script.

This repo already has the infrastructure to build against real production data with no data copy: the `dev_local` DuckDB target reads production Iceberg tables directly via `ol-dbt local register` (Glue-backed views), and `override_ref`/`override_source` transparently fall back to those views when a model isn't built locally yet. What's missing is the comparison step itself.

### Plan/Design
- Add `dbt-labs/dbt_audit_helper` to `src/ol_dbt/packages.yml` for row/column-level relation comparison macros (`compare_relations`, `compare_column_values`, `compare_row_counts`).
- Add a new `ol-dbt diff` command in `src/ol_dbt_cli/ol_dbt_cli/commands/diff.py` (following the existing pattern in `impact.py`/`local_dev.py`), e.g. `ol-dbt diff --old <model> --new <model> [--target dev_local|dev_production]`, wrapping `dbt run-operation audit_helper.compare_relations` (and friends) and surfacing a pass/fail summary plus mismatched rows for inspection.
- Register the command in `src/ol_dbt_cli/ol_dbt_cli/cli.py` and add tests under `src/ol_dbt_cli/tests/test_diff.py`.
- Out of scope: evaluating third-party tools like Datafold's `data-diff` — same-engine comparison via `dbt_audit_helper` covers this need without extra infrastructure.

blarghmatey:
Closed: delivered by PR #2425 (commit 03637811), which added `ol-dbt diff` at src/ol_dbt_cli/ol_dbt_cli/commands/diff.py, registered in cli.py, with dbt_audit_helper in packages.yml. PR #2469 later added local-snapshot and raw-relation support. The PR body said "Closes the intent of #2407", which GitHub does not auto-link, so this stayed open.

Issue #2072, opened 2026-04-06
[Epic] Migrate mart and reporting models to use the dimensional layer


blarghmatey:
## Context

The dimensional layer (`models/dimensional/`) serves as the architectural boundary between data preparation and data consumption. Mart and reporting models should be built exclusively on dimensional models (`dim_*`, `tfact_*`, `afact_*`, `bridge_*`) rather than directly referencing staging or intermediate models.

A comprehensive review found that the majority of pre-existing mart and reporting models bypass the dimensional layer and reference `int__*` or `stg__*` models directly. These models need to be migrated to use the appropriate dimensional models as their upstream sources.

## Goal

All mart (`models/marts/`) and reporting (`models/reporting/`) models should reference only:
- Models in `models/dimensional/`
- Other mart/reporting models (that themselves follow this rule)

## Scope

Sub-issues track individual models to be migrated. Models are grouped by layer.

### Mart Models (bypass dimensional layer entirely)

All existing mart models reference `int__*` intermediate models directly. Each needs to be rewritten to use the appropriate dimensional model(s) as its upstream source.

### Reporting Models (direct non-dimensional references)

Several reporting models reference `int__*` or `stg__*` models directly (some also reference mart models that will be fixed in the mart migration).

## Dependencies

Some migrations require dimensional models that do not yet exist (e.g., `tfact_grade`, `tfact_program_certificate`). Those are tracked in the dimensional warehouse TODO epic. This migration epic is blocked by the TODO epic for those cases.

## Definition of Done

- All mart and reporting models reference only dimensional or other mart/reporting models
- No mart or reporting model contains a `{{ ref('int__*') }}` or `{{ ref('stg__*') }}` call
- All migrated models compile and pass their dbt tests

quazi-h:
Some open questions that came up when I was building context and planning how to prioritize the tickets with copilot. Some of these seem like things that copilot can figure out just by analyzing our dbt models and maybe querying the openmetadata mcp server so I'll do another iteration to see if it can answer some it its own questions:

**Q1: `int__combined__*` models** 
— Several heavy marts (e.g., `marts__combined_course_enrollment_detail`, `marts__combined__users`) reference `int__combined__course_runs`, `int__combined__courserun_enrollments`, `int__combined__users`.
	- Are these expected to map to `dim_course_run`, `tfact_enrollment`, `dim_user`? Or are there cases where the combined intermediates contain logic not yet in the dimensional layer?                                                        
[...]

PR #2432, opened 2026-07-10
feat(superset): refresh assets, fix 2 orphaned datasets, enable --dbt-dir gate

blarghmatey:
## What & why

Enables the **Dataset → dbt-model chain** check (`ol-superset validate --dbt-dir`) as a hard CI gate, and fixes the two orphaned datasets that were blocking it. Also refreshes the checked-in Superset assets from a fresh production export.

The `--dbt-dir` check was deliberately left off when the `superset_validate` workflow shipped (in #2430) because 2 datasets errored. Investigating those (with AWS Glue) showed **both were real production data-quality bugs**, not validator false positives.

## The 2 orphaned datasets (verified against AWS Glue)

| Dataset | Glue table state | Verdict |
|---|---|---|
| `discussion_detail_report` | created **2026-03-16**, never updated since (`__dbt_tmp-…` S3 path) | **~4-month-stale**; feeds a **live** dashboard |
| `video_engagements_report` | created **2025-08-25**, never updated since | **~11-month-stale**; **0 charts** |

- **`discussion_detail_report`** is on the **"Data Detail Discussion"** chart → **"Learner Engagement"** dashboard, so it's live — but its physical table was built **once** from the never-merged PR #1970 branch and never refreshed, so the dashboard has been showing ~4-month-old data. Fix: **virtualize** the dataset (live SQL over `tfact_discussion_events` + `int__combined__course_runs`) so it renders fresh data every load. *(Materializing a reporting model was rejected — it would `ref('int__combined__course_runs')` directly, violating the #2072 dimensional-layering gate.)*
- **`video_engagements_report`** is a dead duplicate: the model was renamed plural→singular (`video_engagement_report`), the live singular dataset already exists, and this one has 0 chart refs. Fix: **delete**.

## Changes (amid the bulk prod re-export — 4 meaningful files)

- **`.github/workflows/superset_validate.yaml`** — run `ol-superset validate --dbt-dir ../ol_dbt` (gating); also trigger on `src/ol_dbt/**` so a dbt model rename/removal that orphans a dataset is caught.
- **`discussion_detail_report_…yaml`** — virtualized (physical → SQL).
- **`video_engagements_report_…yaml`** — deleted.
- **`scripts/drop_stale_warehouse_tables.sh`** — drops the 2 orphaned Glue tables + their S3 data (dry-run by default, freshness-guarded, fetches locations live from Glue).

## Verification

`ol-superset validate --dbt-dir ../ol_dbt` now **exits 0** (was 2 errors); 217 datasets, 21 non-blocking warnings.

## ⚠️ Manual prod follow-ups (sync is upsert-only)

1. **Delete the `video_engagements_report` dataset in Superset prod** — sync never deletes, so removing the file alone won't remove it from prod (it'd reappear on the next export).
2. **After this merges + the virtualization is live**, run `scripts/drop_stale_warehouse_tables.sh --apply` to drop the 2 stale Glue tables + S3 data.

## Follow-up idea (from review discussion)

Sync being upsert-only means repo↔prod drift (assets in prod but not tracked, or vice-versa) goes unnoticed. A `drift`/prune reporting mode for `ol-superset` would surface this — worth a separate task.

Refs #2407, #2072.


blarghmatey:
Split into two reviewable PRs so the meaningful changes aren't buried under the bulk prod re-export (Copilot couldn't review >20k lines):
- #2433 — bulk Superset asset re-export from production (mechanical).
- #2434 — virtualize discussion_detail_report, delete dead video dataset, enable `--dbt-dir` gate, cleanup script (the 4 meaningful files).

Both are green; superseding this PR.

PR #2433, opened 2026-07-10
chore(superset): refresh asset export from production

blarghmatey:
## What

Bulk re-export of Superset assets to sync the repo with current **live production** state (27 dashboards, 174 charts, 218 datasets). Mechanical export diff — no behavioral changes.

Split out from #2432 so the meaningful changes are reviewable on their own. The two datasets that need real fixes and the `--dbt-dir` validation gate are in the companion PR (see below); this PR **excludes** those two datasets to avoid overlap.

## Review note

This is large (Copilot can't review >20k lines) but purely mechanical — it's the diff `ol-superset export` produced against prod. Spot-check rather than line-by-line.


blarghmatey:
Removed the MicroMasters Finance dashboards (`MM_Fin_*`) the export pulled in — 6 dashboards + 42 charts we don't want to maintain. Verified no other dashboard references those charts; `ol-superset validate` stays green (21 dashboards / 132 charts, down from 27 / 174).

Heads-up: sync is upsert-only + export is additive, so these still exist in prod and a future export will re-add them unless deleted in prod or excluded from export (tracked by the drift-detection task `tk-ol-superset-drift-detection-sync-is-upsert-only--ca253c`).

PR #2470, opened 2026-07-22
Change course filter name on enrollment detail dashboard

KatelynGit:
https://github.com/mitodl/hq/issues/12070

Change course filter name on enrollment detail dashboard from "course" to "course readable ID"

### How can this be tested?
check qa and yaml


KatelynGit:
this is such a small change it probably doesn't need review

PR #2497, opened 2026-07-31
fix(dim_user): derive user_pk from durable ids instead of email

rachellougee:

This PR fixes `dim_user.user_pk` being derived from the mutable email, which silently re-keyed users, and orphaned user_fk in every incremental `tfact_` models, whenever someone changed their address.  

### How can this be tested?

dbt build --select dim_user

One-time full refresh of those models after the deployment
```
 dbt build --full-refresh --select tfact_certificate tfact_enrollment tfact_grade \                                                                               
    tfact_order tfact_payment tfact_problem_events tfact_studentmodule_problems   
```

### DIFF
--- a/src/ol_dbt/models/dimensional/_dim__models.yml
+++ b/src/ol_dbt/models/dimensional/_dim__models.yml
@@ -287,7 +287,11 @@ models:
     where we are able to see they are the same person.
   columns:
   - name: user_pk
-    description: string, primary key for this user dim table
+    description: string, primary key for this user dim table. The grain is still one
+      row per shared email, but the key's value is derived from durable identifiers
+      - the user_global_id when the group has one, otherwise the highest-ranked platform
+      account id in the group - so it no longer changes when a user edits their email
+      address.
     tests:
     - not_null
     - unique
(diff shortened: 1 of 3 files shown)


quazi-h (approved):
Changes look good to me. Claude had high praise for you when reviewing this. The fixed priority for resolving grouped user accounts makes sense to me and the tie break using id_source, id_source_user_id addresses the nondeterminism gap in the current implementation. The 7 incremental models that need a full refresh after deployment are mentioned in the PR notes and there are no other models across our other dbt layers that require the same treatment.

Two notes that came up while I was reviewing this with claude code:

1. There was a concern about any data consumers downstream using the reverse_etl role (which has SELECT privileges on the entire dimensional schema). So hightouch or anything else using the reverse_etl role _might_ have a dependency on dim_user but when I looked into hightouch, it doesn't look like we have any active syncs or models there for over a year. I can't think of anything else that might use the reverse_etl role.
2. There was a suggestion to add unit tests specifically for the ranking logic. Claude suggested 4 unit test cases to exercise 4 CTEs in dim_user that depend on that logic:

Test case: test_dim_user_identity_ranking_mitxonline_beats_edxorg
CTE(s) it exercises: combined_accounts (id_source assignment) → account_identity (priority ranking)
What it proves: mitxonline (rank 2) outranks edxorg (rank 3) for the same email, and both merge into one
  person
────────────────────────────────────────
Test case: test_dim_user_identity_stable_across_email_change
CTE(s) it exercises: combined_users (generate_surrogate_key)
What it proves: the actual bug fix — same durable id under a different email still resolves the same
  user_pk_source
────────────────────────────────────────
Test case: test_dim_user_identity_id_bearing_beats_rankless_no_id
CTE(s) it exercises: ranked_accounts (has_no_source_id ordering)
What it proves: an id-bearing account wins even with a numerically worse id_source_rank than an id-less one
────────────────────────────────────────
Test case: test_dim_user_identity_falls_back_to_email_when_no_id_anywhere
CTE(s) it exercises: account_identity (the else 'email' branch)
What it proves: the one disclosed residual gap — no durable id anywhere falls back to email-keyed hashing

Otherwise, I am approving these changes as everything else looks good.

PR #2504, opened 2026-08-05
add new zendesk reporting table

KatelynGit:
https://github.com/mitodl/hq/issues/12607

add new zendesk reporting table

### How can this be tested?
dbt build --select zendesk_report


KatelynGit:
temporarily using int layer until we build dimensional tables for tickets (this causes an error but I'm ignoring it on purpose)

PR #2505, opened 2026-08-05
Add zendesk chart

KatelynGit:
https://github.com/mitodl/hq/issues/12607

Add zendesk ticket chart

### How can this be tested?
review in qa and review the yaml


src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
```diff
@@ -0,0 +1,435 @@
...
+  extra: null
+  filterable: true
+  groupby: true
+  is_active: true
+  is_dttm: false
+  python_date_format: null
+  type: VARCHAR
+  verbose_name: null
+- advanced_data_type: null
+  column_name: ticket_updated_at
```
quazi-h, line 84: Should this be typed as a timestamp instead of a VARCHAR?

src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
```diff
@@ -0,0 +1,435 @@
...
+  extra: null
+  filterable: true
+  groupby: true
+  is_active: true
+  is_dttm: false
+  python_date_format: null
+  type: VARCHAR
+  verbose_name: null
+- advanced_data_type: null
+  column_name: ticket_created_at
```
quazi-h, line 71: Should this be typed as a timestamp instead of a VARCHAR?

src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
```diff
@@ -0,0 +1,435 @@
...
+  verbose_name: null
+currency_code_column: null
+database_uuid: 9a22a54c-8b2f-4c66-a866-3f23812ec929
+default_endpoint: null
+description: null
+extra: null
+fetch_values_predicate: null
+filter_select_enabled: true
+folders: null
+main_dttm_col: ticket_due_at
```
quazi-h, line 416: Should this be set to ticket_created_at?
I know the chart is currently set to filter by ticket_created_at ([line 55](https://github.com/mitodl/ol-data-platform/pull/2505/changes#diff-4bc888101a73a63d15e628d3b149be5f690c127b8e8ed07a4730c6c4c2d54ce8R55)) but was wondering why we didn't do the same for the dataset itself. Currently the ticket_due_at field is also a VARCHAR, not a timestamp (see other comments). I think tickets are much more likely to have no due date at all while all tickets should have a creation date.

quazi-h: I did some more digging on this, both ticket_created_at and ticket_updated_at are ISO-8601 strings (via a cast_timestamp_to_iso8601() macro), correctly reflected as VARCHAR/is_dttm: false in the dataset. This means the sort order for the charts will work correctly (order_by_cols: ticket_created_at desc) since ISO-8601 sorts correctly as text. The issue is once the chart is being used by a dashboard with a global date-range filter (when it is implemented into a weekly recap dashboard) then it might drop tickets without a ticket_due_at value. [According to our own documentation on that field, it is only populated when a ticket is type "task.](https://github.com/mitodl/ol-data-platform/blob/a20a7286540afa8e99f0dde8174a91304a8a88eb/src/ol_dbt/models/staging/zendesk/_stg_zendesk_models.yml#L116-L117)

quazi-h (commented):
Overall, this looks good. I left a few comments about the three date fields in the dataset. The only one that is blocking is which field is set as the "main_dttm_col", I think using "date_created_at" makes the most sense there.

Besides that the only major concern was whether there should be any RLS configured to control access to this data at a per-asset dataset/chart level. As-is the access control for this would be handled by src/ol_superset/policies/ol_rls_policies.json and a separate governance-roles JSON in ol-infrastructure (allowed_schemas/role sync) which is exactly how we've been handling things. Currently, there is no existing filter touching Zendesk/support data in [ol_rls_policies.json](https://github.com/mitodl/ol-data-platform/blob/a20a7286540afa8e99f0dde8174a91304a8a88eb/src/ol_superset/policies/ol_rls_policies.json) and all RLS is course scoped. This means access to int__zendesk__ticket is gated by Superset's role level filtering for whatever role is being used to query/explore access to our intermediate layer. It would probably be worthwhile to check the governance-roles config to make sure that the Customer Support team's role is scoped appropriately to prevent any sensitive PII from being reachable by those roles.

src/ol_superset/assets/dashboards/Enrollment_Detail_by_Learner_da9e03d3-e1bb-45b8-981f-208deca90e7a.yaml
```diff
@@ -5,6 +5,13 @@ dashboard_title: Enrollment Detail by Learner
 description: null
 metadata:
   chart_configuration:
+    '108':
```
quazi-h: This looks unrelated to the actual ticket, probably just reordered by a linter/normalization chore. I just wanted to make sure this was intentional/acceptable to include in this PR.

src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
KatelynGit: nope it's varchar in the table so it's good as is

src/ol_superset/assets/dashboards/Enrollment_Detail_by_Learner_da9e03d3-e1bb-45b8-981f-208deca90e7a.yaml
KatelynGit: o yea it is unrelated I can remove good catch

src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
KatelynGit: nope varchar is what it is in the table we're sourcing from

src/ol_superset/assets/datasets/Trino/int__zendesk__ticket_75791fe7-a03c-4cbe-a3f5-04dd48c58c2e.yaml
KatelynGit: I tried to delete this and got an error message "Warning
You cannot delete the last temporal filter as it's used for time range filters in dashboards." It also doesn't let me replace it. It's not actually filtering anything at the moment so I'll leave it as is

quazi-h (approved):
Looks good, thanks for explaining about how the yaml code is generated. I also confirmed no records are actually being filtered out, so these changes should be good to go.

PR #2582, opened 2026-08-19
docs(cohort-4): spec the ContentFile scoped-pull trigger contract

blarghmatey:

MIT Learn ETL → OL Data Platform migration, Cohort 4: https://github.com/mitodl/hq/issues/11513

Resolves the open design question tracked as `tk-define-contentfile-dagster-mit-learn-trigger-mec-8641eb`.

Adds `docs/design/contentfile_scoped_pull_contract.md`, specifying how Dagster tells MIT Learn that extracted ContentFiles are ready, now that Cohort 4 moves Tika extraction to the platform. Docs only — no code.

**The question the task asked was already answered, but not the one that mattered.** It framed the decision as "(a) webhook from platform to Learn vs (b) Learn polls Dagster asset metadata". (a) shipped some time ago and runs in production from three Dagster locations:

| Sender | Partitioned by |
|---|---|
| `dg_projects/openedx/openedx/assets/openedx.py:725` | `partitions_def` |
| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:340` | `course_and_source_partitions` |
| `dg_projects/canvas/canvas/assets/canvas.py:294` | `canvas_course_ids` (`DynamicPartitionsDefinition`) |

All three POST `{course_id, course_readable_id, content_path, source}` to `/api/v1/webhooks/content_files/`, and MIT Learn responds by firing `ingest_edx_run_archive` — i.e. *"go extract this archive yourself."* Cohort 4 inverts that. So the real question is what the message should mean once the platform does the extracting.

**New endpoint, not a reinterpretation.** All three senders share one path, so changing `process_create_content_file_request` would flip openedx, edxorg and Canvas simultaneously — no per-source rollback. Keeping them independent under the old path means branching per `ETLSource`, which is the same dual-meaning problem relocated. A new path lets sources migrate one at a time with the old one still serving whatever hasn't moved. Canvas migrates too, so the old path retires rather than living on.

**The part that needs genuinely new code is scoping, and it is the risky part.** `BaseWarehouseETLTask` (from mitodl/mit-learn#3566) has two modes and neither fits:

- `full_refresh=True` → prunes; "the only mode that ever sees deletes/unpublishes upstream"
- `full_refresh=False` → never prunes; "a partial pull must never be treated as the complete state of the source"

A per-course trigger is neither. Within one course it *is* the complete state — a removed file must be unpublished — but globally it is not, and pruning globally from a single-course pull would unpublish the entire corpus. §4 specifies a third mode whose prune predicate is scoped to the same key as the fetch, plus a published-row-count guard so a short read skips pruning instead of mass-unpublishing. This is the same full-sync hazard as `MIN_PODCASTS`/`MIN_EPISODES` in #2579, one level down.

**§5 records a constraint that has already caused an outage.** `openedx.py:196` documents it: per-partition `AutomationCondition` evaluation on a 3,500-partition asset produced O(N²) load against the LMS and saturated the run queue until no export ran at all. The spec turns that into a rule — per-partition conditions only where *evaluating* them is per-partition and cheap; otherwise one sweeping sensor batching into a single N-element POST. Hence the payload is a list even for one course.

**§8 flags a gap independent of the decision:** there is no OCW content-file sender anywhere in `dg_projects` and no OCW dg project. OCW's only Learn-facing model is the course-level `integrations__learn__ocw_courses`. Whatever is chosen above, OCW needs a new sending asset built.

### How can this be tested?

Not applicable — documentation only, no code paths. `pre-commit run --files docs/design/contentfile_scoped_pull_contract.md` passes: trailing-whitespace, end-of-files, added-large-files, merge-conflicts and detect-secrets all run on it and pass.

Every code reference in the doc was read from source at `origin/main` on 2026-08-19 rather than recalled: the three sender call sites and their `partitions_def`s, `learn_api.py:55`, `webhooks/urls.py`, `webhooks/views.py:162`/`:179`, the `BaseWarehouseETLTask` docstring, and the `openedx.py:196` note quoted verbatim in §5.

### Additional Context

Decisions this encodes, made by @tmacey: Option C (new endpoint, scoped pull), Canvas migrates with the others, and cadence is set per asset via automation conditions / freshness policies rather than a fixed cron.

Three questions are deliberately left open at the end of the doc rather than settled unilaterally — abort-vs-skip-prune on count mismatch, a batch-size ceiling, and whether the expected count belongs in the task signature or is re-read from the view.

Neither warehouse view (`integrations__learn__ocw_content_files`, `integrations__learn__content_files`) exists yet; §9 states the minimum columns they must expose.

### DIFF
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -93,10 +93,13 @@ created duplicate resources instead of updating them. **Verify the derivation on
 source before enabling.**
 
 **Canvas:** the sender computes `readable_id`, not Learn. Learn's own derivation is
-`f"{course_folder}-{course_code}"` (`learning_resources/etl/canvas.py:138`); the Dagster Canvas
-sender has `course_id` and `metadata["course_code"]` available but does not currently compose
-them into this form. The new Canvas sending asset must build `readable_id` this same way before
-sending, so it matches `LearningResource.readable_id` on receipt.
+`f"{course_folder}-{course_code}"` (`learning_resources/etl/canvas.py:138`), where `course_folder`
+is just `str(course_id)` — Learn's `canvas_course_folder()` derives it from the archive's S3 key,
+but its own docstring says that value *is* the Canvas course id, it's just reached via the archive
+path today. So the formula the new sender builds is `f"{course_id}-{course_code}"`, using the
+`course_id` and `metadata["course_code"]` it already has; it does not need `course_folder` as a
+separate lookup. The new Canvas sending asset must build `readable_id` this way before sending, so
+it matches `LearningResource.readable_id` on receipt.
 
 ## 3. Receiver behaviour
 
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -250,6 +250,21 @@ in `dg_projects`, and no OCW dg project** — OCW's only Learn-facing model is t
 of anything above; it cannot ride the existing mechanism.
 Tracked by `tk-ocw-text-extraction-assets-integrations-learn-oc-8ff9b4`.
 
+**What triggers it, and what happens to the existing Learn-facing webhook.** Today Concourse
+notifies Learn directly when an OCW course publishes, and Learn downloads from S3 and extracts it
+itself; after this change the platform does the extraction, so the platform — not Learn — needs to
+learn a course published. Narrowed to two candidates, not yet decided between them:
+- Concourse's publish step calls a new platform-side endpoint instead of (or alongside) Learn's.
+- A Dagster sensor watches the OCW S3 bucket directly for new/changed course exports.
+
+(Polling OCW Studio's `publish_date` column already in the warehouse was considered and set aside
+— the two options above stay closer to the current push-based, low-latency model.)
+
+Whichever is chosen, it needs to account for **unpublish**, which the existing Learn-facing webhook
+also handles and which happens immediately today. If that webhook is retired outright, the platform
+side needs an equivalent immediate signal for a course being unpublished — the §7 daily
+reconciliation sweep is not a substitute for that latency. Not designed here.
+
 ## 9. Warehouse views the tasks read
 
 | Task | View |
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -23,6 +23,12 @@ All three send a **pointer**: `{course_id, course_readable_id, content_path, sou
 MIT Learn fires `ingest_edx_run_archive` (or `ingest_canvas_course` for Canvas) — i.e. the message
 means *"a new archive is at this path, go extract it yourself."* **MIT Learn does the extraction.**
 
+**Known gap in the current Canvas sender:** `course_readable_id` is only set when Canvas's
+`sis_course_id` is truthy (`dg_projects/canvas/canvas/assets/canvas.py:283-286`) — Learn rejects a
+null value, so some Canvas courses never carry this field today. The new endpoint closes this gap
+rather than inheriting it: §2 has the Canvas sender compute `readable_id` itself from
+`course_id`/`course_code`, so it no longer depends on `sis_course_id` being set.
+
 Cohort 4 inverts that: the platform runs Tika extraction and publishes
 `integrations__learn__ocw_content_files` / `integrations__learn__content_files`, and MIT Learn
 *consumes* extracted text. The message has to change meaning from "go extract" to
@@ -86,6 +92,12 @@ where a dlt-derived id differed from MIT Learn's own derivation by a trailing sl
 created duplicate resources instead of updating them. **Verify the derivation on both sides per
 source before enabling.**
 
+**Canvas:** the sender computes `readable_id`, not Learn. Learn's own derivation is
+`f"{course_folder}-{course_code}"` (`learning_resources/etl/canvas.py:138`); the Dagster Canvas
+sender has `course_id` and `metadata["course_code"]` available but does not currently compose
+them into this form. The new Canvas sending asset must build `readable_id` this same way before
+sending, so it matches `LearningResource.readable_id` on receipt.
+
 ## 3. Receiver behaviour
 
 For each entry, enqueue a **scoped** pull rather than doing work in the request:
@@ -141,14 +153,25 @@ key for the receiver to filter on (see §9).
 
 This is the podcast full-sync hazard (`MIN_PODCASTS` / `MIN_EPISODES` in
 `assets/podcasts.py`) at a different granularity, and it deserves the same guard.
-`content_file_count` in the payload is that guard: the platform states how many rows it published,
-and the task **refuses to prune** if the warehouse returns materially fewer. A course legitimately
-going to zero files is rare enough to be worth an explicit override rather than a silent mass
-unpublish.
+`content_file_count` in the payload is that guard: the platform states how many rows it published
+for this course, and the task compares that against what it actually reads back from the warehouse
+view for the same scope key. Since both numbers come from the same publish event, any gap between
+them is a read-consistency skew (StarRocks/Iceberg catalog cache not yet refreshed, a dbt rebuild
+of the view in flight, etc.), not a legitimate content change — a course that really did shrink
+would already show the smaller count in `content_file_count` itself.
 
-**Open for the implementer:** whether `content_file_count` mismatch should abort the whole pull or
-merely skip pruning while still upserting. Skipping the prune is the safer default — it degrades
-to "stale extra files" rather than "missing files."
+**Decision: any mismatch, not just a "material" one, skips the prune** (still upserts the rows it
+did read). The task never aborts the pull outright on a count mismatch. This is the strict
+direction on purpose — the alternative (tolerating a small gap) risks unpublishing files that
+still exist but weren't visible yet in the read, which is exactly the podcast-migration failure
+mode above; the cost of being strict is that pruning stalls on any catalog-timing hiccup rather
+than clean-deleting the same run, so a lingering-stale-file backlog needs some other cleanup path
+(not designed here) rather than relying on this guard to eventually catch up.
+
+**Still open:** a course legitimately going to zero files has no explicit override path — as
+written, the strict-mismatch guard above also blocks that case (0 read vs. non-zero
+`content_file_count` is the largest possible "mismatch"), so a real empty-course prune needs its
+own signal, not decided here.
 
 ## 5. Cadence — per asset, not one global answer
 
@@ -237,9 +260,13 @@ Column names must match `learn_marts_contract.md` conventions.
 
 ## Open questions
 
-1. **Abort vs skip-prune on count mismatch** (§4). Recommend skip-prune.
+1. ~~Abort vs skip-prune on count mismatch~~ — decided (§4): any mismatch skips the prune, never
+   aborts.
 2. **Batch size ceiling** for the N-element form. Unbounded lists make one failed POST lose a whole
    sweep; a cap turns it into partial progress.
 3. **Does `expected_count` belong in the task signature or re-read from the view?** In the payload
    it is a cross-check between two systems; re-read from the view it only catches read failures,
    not publish failures.
+4. **Explicit override for a legitimate zero-file course** (§4). The strict count-mismatch guard
+   has no way to distinguish "real content lost" from "this course now legitimately has zero
+   files," so a genuine empty-course prune needs its own signal — not designed here.
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -115,6 +115,14 @@ An entry naming an unknown/unsupported `source` is logged and skipped, not 500'd
 source cannot reject an otherwise-valid batch. (Same resilience rule the `learning_resources`
 handler uses.)
 
+**Per-run lock.** Two scoped pulls for the same `readable_id` can be in flight at once — a course
+exported twice in quick succession, or a retried POST. Without coordination, an
+earlier-started/later-finishing pull can prune based on a result set that predates a
+later-started/earlier-finishing pull's upserts, unpublishing files the other pull just added, with
+nothing to correct it until the next export. **Decision: serialize scoped pulls per `readable_id`**
+(a lock keyed on the scope key) rather than a staleness check against a payload timestamp — a
+second pull for a run already in flight waits rather than racing it.
+
 These tasks subclass `BaseWarehouseETLTask` from
 [mit-learn#3807](https://github.com/mitodl/mit-learn/pull/3807)
 (`learning_resources/lib/warehouse.py`), which already provides the StarRocks connection,
@@ -223,6 +231,16 @@ Both paths run simultaneously; each sender moves independently.
 
 Because Canvas migrates, the `ETLSource.canvas` branch in the current handler is transitional, not
 permanent.
+
+**Self-healing after cutover.** This design is purely event-driven — nothing else re-triggers a
+pull. mit-learn's `import_all_*_files` Celery beat sweeps currently provide a backstop against a
+dropped webhook POST by periodically re-pulling everything regardless of what triggered it; once
+those retire, a dropped POST would leave that run stale indefinitely with no automatic recovery.
+**Decision: a daily reconciliation sweep on the Learn side**, not a platform-side re-announce —
+now that the "ETL" on Learn's end is just a SQL pull against the warehouse view rather than
+downloading and parsing an archive, a full sweep is cheap enough to run on a schedule without the
+O(N²) hazard §5 warns about (that hazard is specific to re-triggering the *extraction* side, not
+reading an already-materialized view).
 
 ## 8. OCW has no sender at all
 
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -257,12 +257,18 @@ Tracked by `tk-ocw-text-extraction-assets-integrations-learn-oc-8ff9b4`.
 | `SyncOCWContentFilesTask` | `integrations.integrations__learn__ocw_content_files` |
 | `SyncOpenEdXContentFilesTask` | `integrations.integrations__learn__content_files` |
 
-Neither model exists yet. Both must expose at minimum the scope key (`readable_id` of the owning
-course/run), `etl_source` (needed by the §4 prune predicate to disambiguate scope keys across
-sources), plus the `ContentFile` fields MIT Learn persists — `key`, `title`, `description`,
-`url`, `file_type`, `content`, `content_title`, `content_author`, `content_language`,
-`content_type`, `image_src`, `uid` — and a `last_modified` for the incremental (`since`) path.
-Column names must match `learn_marts_contract.md` conventions.
+Neither model exists yet. Both must expose at minimum the scope key, `etl_source` (needed by the
+§4 prune predicate to disambiguate scope keys across sources), plus the `ContentFile` fields MIT
+Learn persists — `key`, `title`, `description`, `url`, `file_type`, `content`, `content_title`,
+`content_author`, `content_language`, `content_type`, `image_src`, `uid`, `edx_module_id`,
+`source_path`, `file_extension`, `checksum` — and a `last_modified` for the incremental (`since`)
+path. Column names must match `learn_marts_contract.md` conventions.
+
+**Naming: `run_readable_id`, not `readable_id`.** This view is one row per file, so its own
+identity column is `key` (the `ContentFile` key). Calling the scope-key column `readable_id` here
+would clash with §2's `readable_id`, which is the payload's course/run-level field — despite both
+naming the same underlying value, this is one row per *file* using a column that reads like the
+row's own key. Name it `run_readable_id` to keep the two unambiguous.
 
 ## 10. Checklist before enabling any source
 
--- a/docs/design/contentfile_scoped_pull_contract.md
+++ b/docs/design/contentfile_scoped_pull_contract.md
@@ -15,8 +15,8 @@ Today three Dagster locations already POST to `/api/v1/webhooks/content_files/`:
 
 | Sender | Partitioned by |
 |---|---|
-| `dg_projects/openedx/openedx/assets/openedx.py:725` | `partitions_def` (per deployment, per course) |
-| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:340` | `course_and_source_partitions` |
+| `dg_projects/openedx/openedx/assets/openedx.py:731` | `partitions_def` (per deployment, per course) |
+| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:332` | `course_and_source_partitions` |
 | `dg_projects/canvas/canvas/assets/canvas.py:294` | `canvas_course_ids` (`DynamicPartitionsDefinition`) |
 
 All three send a **pointer**: `{course_id, course_readable_id, content_path, source}`. On receipt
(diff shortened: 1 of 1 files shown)


blarghmatey:
Constraint for §9, surfaced reviewing #2589 (https://github.com/mitodl/ol-data-platform/pull/2589#discussion_r3883728068): `integrations__learn__content_files` has to extract the text out of `raw_xml` itself for problem and html blocks.

MIT Learn's `documents_from_olx` gates on `VALID_TEXT_FILE_TYPES`, which includes `.xml`, and every match goes through Tika via `_extract_content` — so what Learn stores for a problem today is extracted text, not markup. The platform side never sees those files: `process_course_xml_blocks` collects only non-XML members into the static-asset bundle, and course XML travels as blocks with `raw_xml` intact. Nothing in the extraction assets touches them.

So unless the view derives text from `raw_xml`, problem and html text disappears at cutover even though both systems have the same bytes.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+| `full_refresh=False` | last watermark | **no** — "a partial pull must never be treated as the complete state of the source" |
+
+A per-course trigger needs a **third mode: scoped full refresh.** Within one course it *is* the
+complete state — a file removed from that course must be unpublished — but globally it is not, and
+pruning globally from a single-course pull would unpublish the entire corpus.
+
+So the prune predicate must be scoped to the same key as the fetch:
+
+> prune content files whose `run.learning_resource.readable_id == readable_id` **and** which are
+> absent from this result set — never anything outside that course.
```
mbertrand: I think this is keyed on the wrong identifier. The §2 example is a run key, Learn already passes `course_readable_id` as `run_id=` ([views.py#L168-L175](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/webhooks/views.py#L168-L175)), and `load_content_files` unpublishes on `ContentFile.run` ([loaders.py#L1081-L1093](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/etl/loaders.py#L1081-L1093)). As written this matches nothing on Open edX; with a course key it unpublishes every other run's files. The scoped prune needs `run__run_id` (currently sent by ol-data-platform as `course_readable_id`)  + `run__learning_resource__etl_source` (sent as `source`).

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+**Open for the implementer:** whether `content_file_count` mismatch should abort the whole pull or
+merely skip pruning while still upserting. Skipping the prune is the safer default — it degrades
+to "stale extra files" rather than "missing files."
+
+## 5. Cadence — per asset, not one global answer
+
+Cadence is chosen per sending asset via Dagster automation conditions / freshness policies, not a
+hardcoded cron. All four senders are already partitioned per course, so per-partition triggering
+needs no restructuring.
```
mbertrand: "All four" - that is including OCW which still needs to be built?

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+Today three Dagster locations already POST to `/api/v1/webhooks/content_files/`:
+
+| Sender | Partitioned by |
+|---|---|
+| `dg_projects/openedx/openedx/assets/openedx.py:725` | `partitions_def` (per deployment, per course) |
+| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:340` | `course_and_source_partitions` |
+| `dg_projects/canvas/canvas/assets/canvas.py:294` | `canvas_course_ids` (`DynamicPartitionsDefinition`) |
+
+All three send a **pointer**: `{course_id, course_readable_id, content_path, source}`. On receipt
```
mbertrand, line 22: +1 to Copilot. The extra bit: Learn's canvas readable id is `f"{course_folder}-{course_code}"` ([etl/canvas.py#L138](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/etl/canvas.py#L138)), and the Dagster sender has `course_id` and `metadata["course_code"]` but never composes them. The contract should state the formula and which side computes it, since the new payload makes `readable_id` required.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+receipt, not completion, matching the existing handlers.
+
+An entry naming an unknown/unsupported `source` is logged and skipped, not 500'd, so one bad
+source cannot reject an otherwise-valid batch. (Same resilience rule the `learning_resources`
+handler uses.)
+
+These tasks subclass `BaseWarehouseETLTask` from
+[mit-learn#3566](https://github.com/mitodl/mit-learn/pull/3566)
+(`learning_resources/lib/warehouse.py`), which already provides the Trino connection, `iter_rows`,
+and the watermark bookkeeping.
```
mbertrand: #3566 is closed now, the class landed in #3807 (`6c9617ddf`). It's StarRocks over pymysql, not Trino ([warehouse.py#L1-L11](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/lib/warehouse.py#L1-L11)), so §9's `integrations.<view>` shorthand won't work with `iter_rows`. And the `apply_async(kwargs=...)` above doesn't fit the class as written — `run()` discards `**kwargs` ([#L226](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/lib/warehouse.py#L226)) and `iter_rows` only filters on `last_modified` ([#L145-L154](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/lib/warehouse.py#L145-L154)). The doc should say which new code: a scope predicate on `iter_rows`, `run()` forwarding it, and scoped runs not touching the watermark.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+directly, with no run in between.
+
+**Rule:** a per-partition `AutomationCondition` is safe only where *evaluating* it is genuinely
+per-partition and cheap. Where evaluation sweeps globally, use a single sensor that batches into
+one N-element POST instead. Prove the evaluation cost before attaching a condition to a
+high-cardinality partitioned asset — this asset graph has already been taken down by getting it
+wrong.
+
+Existing pattern to build on: `_DBT_AUTOMATION_CONDITION = upstream_or_code_changes()` in
+`dg_projects/lakehouse/lakehouse/assets/lakehouse/dbt.py:36`.
```
mbertrand: It's [`dbt.py:39-41`](https://github.com/mitodl/ol-data-platform/blob/397680bca0661ed22aa2ca82417415f6e95a50d8/dg_projects/lakehouse/lakehouse/assets/lakehouse/dbt.py#L39-L41) now and reads `upstream_or_code_changes() if DBT_AUTOMATION_ENABLED else None`.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+| Task | View |
+|---|---|
+| `SyncOCWContentFilesTask` | `integrations.integrations__learn__ocw_content_files` |
+| `SyncOpenEdXContentFilesTask` | `integrations.integrations__learn__content_files` |
+
+Neither model exists yet. Both must expose at minimum the scope key (`readable_id` of the owning
+course/run) plus the `ContentFile` fields MIT Learn persists — `key`, `title`, `description`,
+`url`, `file_type`, `content`, `content_title`, `content_author`, `content_language`,
+`content_type`, `image_src`, `uid` — and a `last_modified` for the incremental (`since`) path.
```
mbertrand: `readable_id` here means the owning run, but [`learn_marts_contract.md:32`](https://github.com/mitodl/ol-data-platform/blob/397680bca0661ed22aa2ca82417415f6e95a50d8/docs/learn_marts_contract.md?plain=1#L32) defines it as the row's own upsert key, and this view is one row per file. I'd call it `run_readable_id` and keep `key` as the file identity. +1 on `etl_source`. The loader also writes `edx_module_id`, `source_path`, `file_extension`, `checksum` ([models.py#L1256-L1262](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/models.py#L1256-L1262)), so those probably belong here too.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+## 0. What this replaces, and why a new endpoint
+
+Today three Dagster locations already POST to `/api/v1/webhooks/content_files/`:
+
+| Sender | Partitioned by |
+|---|---|
+| `dg_projects/openedx/openedx/assets/openedx.py:725` | `partitions_def` (per deployment, per course) |
+| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:340` | `course_and_source_partitions` |
+| `dg_projects/canvas/canvas/assets/canvas.py:294` | `canvas_course_ids` (`DynamicPartitionsDefinition`) |
```
mbertrand: Line numbers have drifted — the calls are at [`openedx.py:731`](https://github.com/mitodl/ol-data-platform/blob/397680bca0661ed22aa2ca82417415f6e95a50d8/dg_projects/openedx/openedx/assets/openedx.py#L731) and [`openedx_course_archives.py:332`](https://github.com/mitodl/ol-data-platform/blob/397680bca0661ed22aa2ca82417415f6e95a50d8/dg_projects/edxorg/edxorg/assets/openedx_course_archives.py#L332), and the §5 docstring starts at `:194` not `:196`. Might be worth citing function names instead so this doesn't rot.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+---
+
+## Open questions
+
+1. **Abort vs skip-prune on count mismatch** (§4). Recommend skip-prune.
+2. **Batch size ceiling** for the N-element form. Unbounded lists make one failed POST lose a whole
+   sweep; a cap turns it into partial progress.
+3. **Does `expected_count` belong in the task signature or re-read from the view?** In the payload
+   it is a cross-check between two systems; re-read from the view it only catches read failures,
+   not publish failures.
```
mbertrand, line 311: One I'd add: what self-heals after cutover? This is purely event-driven, so a dropped POST leaves a run stale until its next export, and the `import_all_*_files` beat sweeps ([settings_celery.py#L266-L268](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/main/settings_celery.py#L266-L268)) are slated to go. Either Learn keeps a slow periodic pull over the view or the sender re-announces everything on a schedule. 

ALso: two pulls for the same run can overlap (a course exported twice in quick succession, or a retried POST). If the earlier one finishes last, its prune step unpublishes whatever the later one just added, and nothing fixes that until the next export. Probably wants a per-run lock, or a check that skips the prune if a newer pull for that run already completed.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+Because Canvas migrates, the `ETLSource.canvas` branch in the current handler is transitional, not
+permanent.
+
+## 8. OCW has no sender at all
+
+The three senders are openedx, edxorg and canvas. **There is no OCW content-file sender anywhere
+in `dg_projects`, and no OCW dg project** — OCW's only Learn-facing model is the course-level
+`integrations__learn__ocw_courses` (Cohort 1). So OCW needs a *new* sending asset built regardless
+of anything above; it cannot ride the existing mechanism.
+Tracked by `tk-ocw-text-extraction-assets-integrations-learn-oc-8ff9b4`.
```
mbertrand, line 254: Right now, when an OCW course is published, the Concourse pipeline sends a webhook to Learn ([views.py#L1302-L1345](https://github.com/mitodl/mit-learn/blob/bb93997bcb84ff93466fc4071a41b2c1f87b4948/learning_resources/views.py#L1302-L1345)), and Learn downloads the course from S3 and extracts the text itself. After this change the platform does the extraction, so the platform is what needs to hear that a course was published, not Learn. The doc does not say how that would happen. Options I can think of: Concourse sends the webhook to the platform instead, or Dagster watches the OCW Studio `publish_date` column that is already in the warehouse, or watches the OCW S3 bucket. Also worth saying whether the existing webhook to Learn goes away entirely — it also handles unpublishing a course, which happens right away today but would have to wait for the next sync if Learn only learns about it by pulling from the warehouse.

docs/design/contentfile_scoped_pull_contract.md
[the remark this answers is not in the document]
blarghmatey: Fixed in af64059d: scope key is run.run_id (sent today as course_readable_id), not LearningResource.readable_id — confirmed against process_create_content_file_request and ContentFile.objects.filter(run=course_run, ...) in mit-learn. Added etl_source alongside it since run_id alone collides across sources.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+## 9. Warehouse views the tasks read
+
+| Task | View |
+|---|---|
+| `SyncOCWContentFilesTask` | `integrations.integrations__learn__ocw_content_files` |
+| `SyncOpenEdXContentFilesTask` | `integrations.integrations__learn__content_files` |
+
+Neither model exists yet. Both must expose at minimum the scope key (`readable_id` of the owning
```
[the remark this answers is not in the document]
blarghmatey: Fixed in af64059d: added etl_source to the §9 minimum schema, required by the corrected §4 prune predicate.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+> prune content files whose `run.learning_resource.readable_id == readable_id` **and** which are
+> absent from this result set — never anything outside that course.
+
+This is the podcast full-sync hazard (`MIN_PODCASTS` / `MIN_EPISODES` in
+`assets/podcasts.py`) at a different granularity, and it deserves the same guard.
+`content_file_count` in the payload is that guard: the platform states how many rows it published,
+and the task **refuses to prune** if the warehouse returns materially fewer. A course legitimately
+going to zero files is rare enough to be worth an explicit override rather than a silent mass
+unpublish.
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 2f5d2918: content_file_count guard is now strict — any mismatch skips the prune (never aborts). Zero-file-course override left as an explicit open question (#4) since the strict guard as written blocks that case too.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,227 @@
...
+
+| Sender | Partitioned by |
+|---|---|
+| `dg_projects/openedx/openedx/assets/openedx.py:725` | `partitions_def` (per deployment, per course) |
+| `dg_projects/edxorg/edxorg/assets/openedx_course_archives.py:340` | `course_and_source_partitions` |
+| `dg_projects/canvas/canvas/assets/canvas.py:294` | `canvas_course_ids` (`DynamicPartitionsDefinition`) |
+
+All three send a **pointer**: `{course_id, course_readable_id, content_path, source}`. On receipt
+MIT Learn fires `ingest_edx_run_archive` (or `ingest_canvas_course` for Canvas) — i.e. the message
+means *"a new archive is at this path, go extract it yourself."* **MIT Learn does the extraction.**
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 2f5d2918: recorded the current gap in §0 (course_readable_id only set when sis_course_id is truthy) and closed it going forward — §2 now has the Canvas sender compute readable_id itself instead of depending on that field.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in af64059d: scope key is now run.run_id (+ etl_source), not LearningResource.readable_id. Your citations (process_create_content_file_request, ContentFile.objects.filter(run=course_run, ...)) confirmed the exact fix — thanks for the exact line refs.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Right, that was wrong. Fixed in af64059d: now says the three existing senders (openedx/edxorg/canvas); OCW's sender is called out separately in §8, and §8 now also narrows down what triggers it (47523a80).

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in 2f5d2918: the new Canvas sender computes readable_id itself as f"{course_folder}-{course_code}" (course_folder == course_id, per canvas_course_folder()), matching Learn's own derivation in etl/canvas.py:138. §0 also now documents the existing webhook's course_readable_id gap this replaces.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in af64059d: link updated to the merged mit-learn#3807 (StarRocks, not Trino). Also flagged that iter_rows/run() don't support a scope predicate yet — that's new mit-learn-side code, not designed in this doc.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in af64059d — now points at dbt.py:39.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in 29fdda16: renamed the view's scope-key column to run_readable_id (keeping key for file identity), and added edx_module_id, source_path, file_extension, checksum to the minimum schema.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in af64059d — openedx.py:731, openedx_course_archives.py:332, openedx.py:194.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Landed in 18b0a7de: daily reconciliation sweep on the Learn side for self-heal (cheap now that it's a SQL pull, not a re-triggered extraction), and a per-run lock keyed on readable_id to stop the prune race.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Narrowed in 47523a80: Concourse calling a new platform-side endpoint, or a Dagster sensor on the OCW S3 bucket — not decided between the two yet. Polling OCW Studio's publish_date was considered and set aside. Flagged that whichever is chosen still needs to cover unpublish, which the daily sweep is too slow to replace.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,272 @@
...
+Getting this wrong is the failure mode from the podcast migration (`integrations__learn__podcasts`),
+where a dlt-derived id differed from MIT Learn's own derivation by a trailing slash and would have
+created duplicate resources instead of updating them. **Verify the derivation on both sides per
+source before enabling.**
+
+**Canvas:** the sender computes `readable_id`, not Learn. Learn's own derivation is
+`f"{course_folder}-{course_code}"` (`learning_resources/etl/canvas.py:138`); the Dagster Canvas
+sender has `course_id` and `metadata["course_code"]` available but does not currently compose
+them into this form. The new Canvas sending asset must build `readable_id` this same way before
+sending, so it matches `LearningResource.readable_id` on receipt.
```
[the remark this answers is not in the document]
blarghmatey: Fixed in bd9e81e2: spelled out that course_folder is just str(course_id) — Learn's canvas_course_folder() docstring confirms that value is the Canvas course id, just reached via the archive's S3 key today. Formula is f"{course_id}-{course_code}", using fields the sender already has.

blarghmatey:
Investigated further — this needs a capa/OLX-aware tree-walk over the parsed block XML that splits student-safe question text from embedded `<solution>`/answer content, rather than generic Tika tag-stripping (which doesn't distinguish the two and is what today's Learn-side extraction already does, unnoticed). Traced the full blast radius across ol-data-platform, mit-learn, and learn-ai's AI tutor and wrote it up as an RFC: https://github.com/mitodl/hq/discussions/13220

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,314 @@
...
+source before enabling.**
+
+**Canvas:** the sender computes `readable_id`, not Learn. Learn's own derivation is
+`f"{course_folder}-{course_code}"` (`learning_resources/etl/canvas.py:138`), where `course_folder`
+is just `str(course_id)` — Learn's `canvas_course_folder()` derives it from the archive's S3 key,
+but its own docstring says that value *is* the Canvas course id, it's just reached via the archive
+path today. So the formula the new sender builds is `f"{course_id}-{course_code}"`, using the
+`course_id` and `metadata["course_code"]` it already has; it does not need `course_folder` as a
+separate lookup. The new Canvas sending asset must build `readable_id` this way before sending, so
+it matches `LearningResource.readable_id` on receipt.
```
mbertrand: Line 85 says `readable_id` must equal `ContentFile.run.run_id`, then this one says the Canvas value must match `LearningResource.readable_id`. For Canvas those differ: Learn creates the run with `run_id=f"{readable_id}+canvas"` ([canvas.py#L179](https://github.com/mitodl/mit-learn/blob/main/learning_resources/etl/canvas.py#L179)), so `f"{course_id}-{course_code}"` matches the resource but not the run, and a prune keyed on `run.run_id` would match nothing for Canvas. Same class of problem as the first round, just moved to Canvas. Either the sender appends `+canvas` or the receiver does, but the doc should pick one.

Also the §3 lock is keyed on `readable_id` alone, but §4 says `run_id` isn't unique across sources. Probably wants `(source, readable_id)`.

mbertrand (approved):
1 more suggested change, otherwise looks great

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+with `SyncOCWContentFilesTask` for `ocw`. Respond `200` once enqueued — the webhook acknowledges
+receipt, not completion, matching the existing handlers.
+
+An entry naming an unknown/unsupported `source` is logged and skipped, not 500'd, so one bad
+source cannot reject an otherwise-valid batch. (Same resilience rule the `learning_resources`
+handler uses.)
+
+**Per-run lock.** Two scoped pulls for the same `readable_id` can be in flight at once — a course
```
shanbady: Needs a bound. A blocking Celery task holds a worker slot; an N-course sweep starves the queue - same shape as §5's outage. A second pull reads the same view state, so discard-as-redundant or re-queue with countdown.

docs/design/contentfile_scoped_pull_contract.md
```diff
@@ -0,0 +1,314 @@
...
+
+`BaseWarehouseETLTask` has exactly two modes, and **neither is what this needs**:
+
+| mode | `since` | prunes? |
+|---|---|---|
+| `full_refresh=True` | `None` | yes — "the only mode that ever sees deletes/unpublishes upstream" |
+| `full_refresh=False` | last watermark | **no** — "a partial pull must never be treated as the complete state of the source" |
+
+A per-course trigger needs a **third mode: scoped full refresh.** Within one course it *is* the
+complete state — a file removed from that course must be unpublished — but globally it is not, and
```
shanbady, line 169: "Unpublish" isn't uniform. RESOURCE_FILE_ETL_SOURCES (etl/constants.py:121) soft-deletes; Canvas hard-deletes (canvas.py:257-262) and is excluded from cleanup_deleted_content_files (tasks.py:1018), so Canvas soft-deletes would never be reaped.

shanbady (changes requested):

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in 0bc152ee1: the Canvas sender now appends `+canvas` and sends `f"{course_id}-{course_code}+canvas"`, so `readable_id == run.run_id` holds for every source with no receiver-side branch. The §3 lock is now keyed on `(source, readable_id)` to match the §4 prune predicate.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in 0bc152ee1: the lock is acquired non-blocking, and a task that finds it held re-queues with a countdown under a max-retries cap, then falls back to the §7 daily sweep. I went with re-queue over discard because a re-export that lands mid-pull is not redundant with the in-flight read and would otherwise wait for the daily sweep.

docs/design/contentfile_scoped_pull_contract.md
blarghmatey: Fixed in 0bc152ee1: §4 now says the scoped prune keeps each source's current semantics, soft-delete for `RESOURCE_FILE_ETL_SOURCES` and hard-delete for Canvas. Moving Canvas to soft-delete would mean adding it to `RESOURCE_FILE_ETL_SOURCES`, which also gates search indexing (`learning_resources_search/tasks.py:669,1098`), so that stays out of scope here.

shanbady (approved):
👍

PR #2583, opened 2026-08-19
feat(openedx): collect static/ content files instead of discarding them

blarghmatey:

MIT Learn ETL → OL Data Platform migration, Cohort 4 (complex extraction / ContentFiles): https://github.com/mitodl/hq/issues/11513

**Bottom of a stack.** This layer only changes what the archive walker collects. The extraction assets that consume it, and the `integrations__learn__content_files` model above them, are the layers on top.

Cohort 4 moves ContentFile text extraction onto the data platform — the platform runs Tika and publishes extracted text, instead of MIT Learn re-extracting course archives itself. That work needs the course's actual content files, and `process_course_xml_blocks` could not supply them.

It skipped `drafts/`, `assets/` and `static/` **before** the non-XML branch, so their bytes were discarded along with their XML:

```python
if block_type in {"drafts", "assets", "static"}:
    continue
# NOTE: This check must come before the non-XML branch so that non-XML
# files in these directories (e.g. draft images, PDFs) are also excluded.
```

Per the [OLX reference](https://docs.openedx.org/en/latest/educators/olx/directory-structure.html), `static/` "contains the files used in a course, such as images or PDFs" — so the bundle was structurally unable to carry course content. The comment in the excluded branch even named PDFs as something it was dropping.

**One check was answering two different questions.** *"Is this a course block type?"* is correctly **no** for all three directories. *"Should these bytes be collected?"* is a separate question with a different answer per directory, and conflating them is what discarded the content. The fix separates them rather than loosening the predicate:

- **`drafts/`** — skipped entirely. It is the Studio draft workspace, so its contents are unpublished and must never reach the catalogue. Note the old uniform rule got this one right, but for the wrong reason; making the reason explicit is what stops a later edit from regressing it.
- **`assets/` and `static/`** — still never parsed as blocks, but their non-XML payload is now collected.
- **XML inside those two** — skipped rather than collected: course XML comes from the block directories, and a document extractor has nothing to do with an XML manifest.

**`assets/` is not in the documented export layout.** It does appear in real archives, and the original author classed it as file storage alongside `static/`, so it is treated the same on that reasoning. Calling this out because it is the one part of the change not backed by the OLX reference.

**What I could not verify.** The integration tests run against real course archives from `~/Downloads`, which are not present here (and are skipped in CI). So the change is verified against the OLX reference documentation and the synthetic fixture, not against a production archive. Worth someone running `pytest tests/lib/test_openedx_integration.py -m integration` with real archives before merge — that suite is exactly what would catch a wrong assumption about real-world layout.

I also dropped two claims I had initially written and could not substantiate: that the XML under `static/` is `assets.xml` (the documented asset manifest is `policies/assets.json`), and that subtitles are specifically `static/subs_*.srt.sjson` (widely-used convention, but the docs do not state it).

### Behavioural change worth flagging explicitly

This changes `CourseStaticAssetsBundle.data_version` for any course with static content, because the hash covers the collected bytes. `course_static_assets` carries `automation_condition=upstream_or_code_changes()`, so **affected partitions will re-materialize on the next run.**

### How can this be tested?

```
cd packages/ol-orchestrate-lib
uv run --with pytest-cov pytest tests/lib/test_openedx.py -q
```

Actually run on this branch:
- `tests/lib/test_openedx.py` — **40 passed**
- full `packages/ol-orchestrate-lib` suite — **197 passed, 80 skipped**
- `dg_projects/openedx` suite (the consumer of this function) — **25 passed**
- `pre-commit run` on the changed files — passes

Four new tests pin the behaviour that matters, rather than just the happy path:
- `..._collects_static_content_files` — PDFs, subtitles and images from `static/` are collected
- `..._drafts_never_collected` — draft content is withheld from **both** block parsing and file collection
- `..._file_storage_xml_is_neither` — XML under `static/` is neither a block nor a collected file
- `..._manifest_types_content_files` — the manifest carries the `mime_type` the extraction layer above dispatches on

The existing `..._structural_dirs_excluded` test is unchanged and still passes: those directories are still never block *types*.

### Additional Context

The upstream integration test previously asserted the old contract (`no files from excluded structural directories are present`); it now asserts only that draft content is withheld.


blarghmatey:
## Validated against a real course export

The "what I could not verify" caveat in the description is now resolved. I pulled `course-v1:MITx+14.310x+3T2021` (152 MiB) from `s3://ol-data-lake-landing-zone-qa/mitxonline/openedx/raw_data/course_xml/` and ran `process_course_xml_blocks` against it directly.

**The OLX reference was right, and the scale of what was being discarded is larger than I assumed.** Top-level file counts in the archive:

| dir | files |
|---|---|
| **static** | **1,672** |
| problem | 782 |
| vertical | 688 |
| html | 364 |
| video | 305 |
| drafts | 12 |
| assets | 1 |

`static/` is the **single largest directory in the export**, and on `main` every one of those 1,672 files is dropped. Its contents are exactly what Cohort 4 needs: 116+ PDFs, 302 `.srt` and 720 `.srt.sjson` subtitles, alongside images, `.R` scripts and CSVs.

Results on this branch:
- **1,874 files collected** (1,672 from `static/`, 182 from `html/`, plus `about/`, `tabs/`, `info/`, `policies/`)
- **2,028 XML blocks still parsed** — block parsing is unaffected
- **`drafts/` withheld**, asserted explicitly; no draft file reached the bundle
- `manifest.file_count` agrees with the collected file count

Two details worth noting, both of which turned out to matter downstream:

**Filenames in `static/` contain dots.** Real examples: `14.310x Syllabus.pdf`, `14.310_Lecture_15_Segment_03-_Part1.srt`, `subs_-GQxFdhr_qU.srt.sjson`. Anything deriving a type by splitting on the *first* dot gets all three wrong. Last-suffix matching is load-bearing, and the layers above this one depend on it.

**`assets/` exists in a real export**, with one file — the description notes this directory is not in the documented OLX layout, and that is confirmed: it is present but nearly empty here.

The one thing still not covered: this is a single course from one deployment. `packages/ol-orchestrate-lib/tests/lib/test_openedx_integration.py -m integration` against a spread of archives remains worth running before merge.

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/openedx.py
```diff
@@ -494,9 +494,24 @@ def process_course_xml_blocks(  # noqa: C901, PLR0912, PLR0915
...
+    - ``drafts/`` is skipped entirely. It is the Studio draft workspace, so its
+      contents are unpublished and must not reach the catalogue.
+    - ``assets/`` and ``static/`` are never parsed as blocks, but their non-XML
+      contents ARE collected. Per the OLX reference, ``static/`` "contains the
+      files used in a course, such as images or PDFs"
+      (https://docs.openedx.org/en/latest/educators/olx/directory-structure.html),
+      which makes it the source of the documents ContentFile text extraction
+      consumes. Excluding it left the bundle structurally unable to carry course
+      content. ``assets/`` is not in the documented export layout but appears in
+      real archives; it is treated the same way on the same reasoning.
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in ba7e94b. Reproduced first: two archives differing only in a static file's name both hashed to 785c6c82..., so the S3 key was reused and every downstream data_version_changed() check missed the rename. The path is now hashed alongside the bytes, length-prefixed so a shifted path/content boundary can't collide ("ab"+"c" vs "a"+"bc"). Three tests: the rename, the shifted boundary, and that identical content still hashes stably.

packages/ol-orchestrate-lib/tests/lib/test_openedx.py
```diff
@@ -256,15 +269,65 @@ def test_process_course_xml_blocks_static_assets(sample_course_archive):
...
+    This is the input ContentFile text extraction consumes: uploaded PDFs and
+    subs_*.srt.sjson subtitles live in static/, and excluding them would make the
+    bundle structurally unable to carry course content.
+    """
+    archive_path, _ = sample_course_archive
+    _, bundle = process_course_xml_blocks(archive_path, "prod")
+    paths = [p for p, _ in bundle.files]
+
+    assert "static/syllabus.pdf" in paths, "Uploaded PDFs must be collected"
+    assert "static/subs_video1.srt.sjson" in paths, "Subtitles must be collected"
```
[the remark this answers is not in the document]
blarghmatey: Verified — the premise holds but the consequence doesn't. mimetypes does return application/octet-stream for .sjson, but nothing downstream dispatches on MIME: is_transcript() and transcript_text() both key on Path(...).suffix.lower(), which is also why subs_<id>.srt.sjson resolves as SJSON rather than SRT. The manifest mime_type is descriptive metadata, not a dispatch input. Adding a .sjson MIME mapping would also make _content_type() report application/json, which is now in SUPPORTED_CONTENT_TYPES — harmless only because the is_transcript check runs first. Not adding it; it would create a second, MIME-based path to a decision that is deliberately suffix-based.

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/openedx.py
```diff
@@ -588,6 +610,11 @@ def process_course_xml_blocks(  # noqa: C901, PLR0912, PLR0915
                     static_assets.append((relative_path, file_data.read()))
```
mbertrand: The bundle now holds all of static/ in memory as a list of byte strings, and static/ is where the bulk of a big export lives. I streamed the three largest QA archives to check: they are ~99% static/ — 1286 MiB of static files in `TEST+BACKUPTEST101+3T2019` (xpro), 733 MiB in `xPRO+AMX+R21` (a real course), 1167 MiB in `TestX+8.02.1x+1T2023`. 15 QA courses have archives over 500 MiB. Run pods request 1Gi / limit 8Gi, so a single run fits, but the re-materialization wave could put several GiB-scale runs on nodes that only reserved 1Gi each. Not blocking — but worth either bumping the memory request, or streaming the bundle to a temp file instead of `list[tuple[str, bytes]]` so it stays flat regardless of course size.

packages/ol-orchestrate-lib/tests/lib/test_openedx.py
```diff
@@ -569,3 +632,68 @@ def test_un_nest_course_structure_survives_a_library_reference():
...
+
+
+def test_data_version_distinguishes_a_shifted_path_content_boundary(tmp_path):
+    """Length-prefixing keeps a path/content split from being ambiguous.
+
+    Without it, concatenating path and bytes lets two different bundles hash
+    identically when the boundary moves -- "ab" + "c" and "a" + "bc".
+    """
+    one = _archive_with_static(tmp_path, "one", {"ab.txt": b"c"})
+    two = _archive_with_static(tmp_path, "two", {"a.txt": b"bc"})
```
mbertrand: This test can't actually fail — the `.txt` extensions break the ambiguity it guards against. Naive concatenation of `static/ab.txt`+`c` and `static/a.txt`+`bc` already differ, so it passes even without the length-prefixing. The genuinely ambiguous pair is extension-less:

```suggestion
    one = _archive_with_static(tmp_path, "one", {"ab": b"c"})
    two = _archive_with_static(tmp_path, "two", {"a": b"bc"})
```

(verified `static/ab`+`c` and `static/a`+`bc` concatenate to identical bytes)

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/openedx.py
```diff
@@ -509,9 +509,24 @@ def process_course_xml_blocks(  # noqa: C901, PLR0912, PLR0915
     Non-XML files (e.g. SRT subtitles, HTML content, PDFs) are collected and returned
     as a CourseStaticAssetsBundle containing the raw files, a content-addressed
     data_version (SHA-256 of all file contents sorted by path), and a manifest JSON
```
mbertrand: This docstring is out of date: it says the hash covers file contents only, but the path-hashing commit (b320b68) made it cover the file paths too, so a rename changes data_version now. The `data_version` field description on `CourseStaticAssetsBundle` says the same thing and needs the same fix.

packages/ol-orchestrate-lib/tests/lib/test_openedx.py
```diff
@@ -256,15 +269,65 @@ def test_process_course_xml_blocks_static_assets(sample_course_archive):
...
+    subs_*.srt.sjson subtitles live in static/, and excluding them would make the
+    bundle structurally unable to carry course content.
+    """
+    archive_path, _ = sample_course_archive
+    _, bundle = process_course_xml_blocks(archive_path, "prod")
+    paths = [p for p, _ in bundle.files]
+
+    assert "static/syllabus.pdf" in paths, "Uploaded PDFs must be collected"
+    assert "static/subs_video1.srt.sjson" in paths, "Subtitles must be collected"
+    assert "static/logo.png" in paths, "Uploaded images must be collected"
```
mbertrand, line 328: The fixture here has a static/ dir but no assets/ dir, so the claim that assets/ files are collected is only tested in the integration suite — and that suite is skipped in CI. Adding one file under assets/ to the fixture and asserting it appears in the bundle would get that covered by a test CI actually runs.

mbertrand (approved):
Left a few comments about things that might be good to address before merging but don't seem to be blockers.

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/openedx.py
blarghmatey: Fixed in fea9f44: process_course_xml_blocks now streams each member into the tar.gz as it walks the source archive and returns the path, so peak memory is one chunk regardless of course size. Both asset callers drop their re-bundling loop. Kept the write in source-archive order on purpose — a sorted second pass seeks backwards through the gzip stream, which costs a full re-decompression per member on exactly the GiB-scale archives you measured; the version stays order-independent by hashing per-file digests combined in path order.

packages/ol-orchestrate-lib/tests/lib/test_openedx.py
blarghmatey: Fixed in fea9f44: took the extension-less pair. Also worth flagging — the hash now combines fixed-width per-file digests rather than concatenating raw bytes, so that ambiguity class is structurally gone; the length prefix on the path stays and the test now reads as the property check it is.

packages/ol-orchestrate-lib/src/ol_orchestrate/lib/openedx.py
blarghmatey: Fixed in fea9f44: both the function docstring and the data_version field description now say the digest covers each file's path and contents, so a rename moves the version.

packages/ol-orchestrate-lib/tests/lib/test_openedx.py
blarghmatey: Fixed in fea9f44: the fixture now has assets/handout.pdf and test_process_course_xml_blocks_collects_static_content_files asserts it lands in the bundle, so the assets/ path is covered by a test CI runs.

PR #2586, opened 2026-08-19
feat(openedx): parse video transcripts out of course archives

blarghmatey:

MIT Learn ETL → OL Data Platform migration, Cohort 4: https://github.com/mitodl/hq/issues/11513

**Layer 3 of a stack. Base is #2584**, which is based on #2583. Review from the bottom.

Adds `openedx/processed_data/course_transcript_text`: one JSONL row per subtitle file, in the **same row shape as the document rows** from #2584, so the `integrations__learn__content_files` model above can union the two without reconciling schemas.

Registered per deployment alongside the document asset — verified that `mitx`, `mitxonline` and `xpro` each expose both `course_document_text` and `course_transcript_text`.

**Why this is a separate asset rather than part of #2584.** Transcripts parse in-process and need no external service. Folding them into the Tika asset would mean a Tika outage also stopped transcript extraction — two independent failure domains sharing one fate for no reason. The cost is that each asset downloads the bundle itself; a second download is cheaper than a coupled outage.

**The parsers reproduce mit-learn exactly, and its test cases are ported verbatim.** `text_from_srt_content` and `text_from_sjson_content` mirror `learning_resources/etl/utils.py`, and the two cases from `utils_test.py` are copied into this suite as the parity contract. This is the same approach used for the podcast duration parser in #2579: a cutover diff should reveal a difference in the *data*, never in how the two sides parse it.

**One quirk is preserved deliberately.** mit-learn's SRT sequence-number rule is a bare `\d+\n`, which also strips a caption line that happens to be only digits — a subtitle reading `1984` is removed as though it were an index. That is pinned by `test_srt_digit_only_caption_is_stripped`, with a comment saying why: diverging would turn every such caption into a spurious difference during parallel validation. Worth fixing, but in both repos together, not here.

**Suffix dispatch matters more than it looks.** Open edX names its subtitles `subs_<id>.srt.sjson` — that is **SJSON, not SRT**, despite the middle of the filename. Dispatch is on the final suffix, matching mit-learn's transformer map, and there is a test for exactly this case.

**Two accounting decisions carried over from #2584, for the same reasons:**
- A transcript parsing to empty text is **recorded**, not dropped — an empty caption track and a file that was never seen are different facts.
- A malformed transcript is **counted and skipped**, not raised — one bad file must not cost the whole course.

Note this layer has **no** health guard equivalent to #2584's `assert_extraction_healthy`. That guard exists because a Tika outage can fail every document at once; transcript parsing has no external dependency that can fail wholesale, so the analogous failure mode does not exist here. Called out because its absence is a decision, not an oversight.

### Refactor included

`read_bundle_members` and `bundle_data_version` are factored out of #2584's asset, since both assets now need them. This is why the diff touches `content_files.py`.

### How can this be tested?

```
cd dg_projects/openedx
uv run --with pytest-cov pytest openedx_tests/ -q
```

Actually run on this branch:
- `dg_projects/openedx` suite — **54 passed** (19 new)
- `uv run pre-commit run --all-files` — passes (ruff, mypy, and the rest)
- Code-location load check — all three deployment repositories resolve, each exposing both extraction assets

Not run: parsing against a real course archive. The parsers are exercised against mit-learn's own fixtures and hand-built cases; the first real run is the actual check on production subtitle files.

### Additional Context

A latent bug ruff and mypy caught that the tests could not: refactoring `read_bundle_members` out of #2584's asset body briefly left `tarfile` unused, the formatter removed the import, and nothing in the test suite exercises that code path — so it would have surfaced as a runtime `NameError` on the first real materialization. Fixed here, and worth knowing that the pure-function tests in this area deliberately do not cover the I/O path.

### DIFF
--- a/dg_projects/openedx/openedx/assets/content_files.py
+++ b/dg_projects/openedx/openedx/assets/content_files.py
@@ -22,6 +22,7 @@ import logging
 import mimetypes
 import tarfile
 from collections.abc import Callable, Iterable, Iterator
+from contextlib import contextmanager
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any
@@ -74,6 +75,53 @@ def is_transcript(relative_path: str) -> bool:
     suggests.
     """
     return Path(relative_path).suffix.lower() in TRANSCRIPT_SUFFIXES
+
+
+@contextmanager
+def open_bundle_members(
+    bundle_location: UPath, temp_files: list[Path]
+) -> Iterator[Iterator[tuple[str, Callable[[], bytes]]]]:
+    """Download the static-asset tarball and walk it without loading it all.
+
+    Yields an iterator of ``(relative_path, read_bytes)``. ``read_bytes`` is a
+    callable, so a member's bytes are only pulled into memory if the caller
+    actually wants that member -- both callers decide from the path alone
+    whether a file is theirs to handle.
+
+    That laziness is the point. A real 14.310x export is 152 MiB and is mostly
+    images and archives that neither asset touches: reading every member up
+    front held all of it resident alongside the downloaded tarball and the
+    extracted text, which is what could exhaust a worker on a media-heavy
+    course. Members are streamed one at a time rather than materialised as a
+    list, so a completed member's bytes become collectable immediately.
+
+    ``temp_files`` accumulates paths for the caller to clean up, so the download
+    is removed even when extraction raises.
+
+    Shared with the transcript asset: both read the same bundle, and each pays
+    its own download rather than one depending on the other's output. That is
+    deliberate -- it keeps a Tika outage from also stopping transcripts.
+    """
+    bundle_path = Path(
+        NamedTemporaryFile(delete=False, suffix="_static_assets.tar.gz").name
+    )
+    temp_files.append(bundle_path)
+    bundle_location.fs.get_file(str(bundle_location), str(bundle_path))
+
+    with tarfile.open(bundle_path, "r:gz") as bundle:
+
+        def _members() -> Iterator[tuple[str, Callable[[], bytes]]]:
+            for member in bundle:
+                if not member.isfile():
+                    continue
+
+                def _read(member: tarfile.TarInfo = member) -> bytes:
+                    extracted = bundle.extractfile(member)
+                    return extracted.read() if extracted else b""
+
+                yield member.name, _read
+
+        yield _members()
 
 
 def _content_type(relative_path: str) -> str:
@@ -349,37 +397,12 @@ def extract_course_document_text(
 
     temp_files: list[Path] = []
     try:
-        bundle_path = Path(
-            NamedTemporaryFile(delete=False, suffix="_static_assets.tar.gz").name
-        )
-        temp_files.append(bundle_path)
         context.log.info(
             "Downloading static asset bundle from %s", course_static_assets
         )
-        course_static_assets.fs.get_file(str(course_static_assets), str(bundle_path))
-
-        # Members are walked lazily and their bytes pulled only for the ones
-        # that survive filtering. A real 14.310x export is 152 MiB and is mostly
-        # images and archives that get skipped; reading every member up front
-        # held all of that resident alongside the tarball and the extracted
-        # text, which is what could exhaust a worker on a media-heavy course.
-        with tarfile.open(bundle_path, "r:gz") as bundle:
-
-            def _members(
-                bundle: tarfile.TarFile = bundle,
-            ) -> Iterator[tuple[str, Callable[[], bytes]]]:
-                for member in bundle:
-                    if not member.isfile():
-                        continue
-
-                    def _read(member: tarfile.TarInfo = member) -> bytes:
-                        extracted = bundle.extractfile(member)
-                        return extracted.read() if extracted else b""
-
-                    yield member.name, _read
-
+        with open_bundle_members(course_static_assets, temp_files) as members:
             rows, counters = build_document_rows(
-                _members(),
+                members,
                 course_id=course_id,
                 source_system=source_system,
                 extract=tika.extract_text,
--- a/dg_projects/openedx/openedx/assets/transcripts.py
+++ b/dg_projects/openedx/openedx/assets/transcripts.py
@@ -112,7 +112,11 @@ def transcript_text(relative_path: str, raw: bytes) -> str | None:
     the caller distinguishes the two by checking the suffix itself.
     """
     try:
-        content = raw.decode("utf-8")
+        # utf-8-sig, not utf-8: MIT Learn runs these through Tika, which does
+        # charset detection first, so a BOM'd transcript extracts fine there and
+        # would fail here. No BOMs appeared in ~20k srt/sjson files sampled
+        # across 49 production exports, so this is insurance rather than a fix.
+        content = raw.decode("utf-8-sig")
     except UnicodeDecodeError:
         log.exception("Transcript is not valid UTF-8: %s", relative_path)
         return None
@@ -139,9 +143,9 @@ def build_transcript_rows(
     ``members`` yields ``(relative_path, read_bytes)``. The bytes stay behind a
     callable so the many non-transcript files in an export are never read.
 
-    A transcript that parses to empty text is still recorded. An empty caption
-    track and a file that was never seen are different facts, and only one of
-    them is worth investigating.
+    A transcript that parses to empty text is still recorded, and so is one that
+    fails to parse. An empty caption track, an unparseable file and a file that
+    was never seen are three different facts, and only the row carries which.
     """
     rows: list[dict[str, Any]] = []
     counters = {"candidates": 0, "extracted": 0, "empty": 0, "failed": 0}
@@ -156,6 +160,24 @@ def build_transcript_rows(
 
         if text is None:
             counters["failed"] += 1
+            # A failed parse still gets a row. Emitting nothing makes it
+            # indistinguishable downstream from the file having been deleted
+            # from the course, which is what MIT Learn unpublishes on.
+            rows.append(
+                {
+                    "course_id": course_id,
+                    "source_system": source_system,
+                    "file_path": relative_path,
+                    "content_type": (
+                        "application/json"
+                        if relative_path.lower().endswith(SJSON_SUFFIX)
+                        else "text/plain"
+                    ),
+                    "size_bytes": len(raw),
+                    "content": None,
+                    "extraction_status": "failed",
+                }
+            )
             continue
 
         # Status is decided on the stripped text, so the content must be
@@ -192,6 +214,26 @@ def build_transcript_rows(
     return rows, counters
 
 
+def assert_parsing_healthy(counters: dict[str, int], course_id: str) -> None:
+    """Refuse to publish a course whose every transcript failed to parse.
+
+    Without this the asset publishes a JSONL of nothing but failure rows, which
+    a consumer that filters on status reads as "this course has no transcripts"
+    -- the failure mode that looks like success. The document asset guards the
+    same case; this is the total-failure half of it, which is the part that is
+    not Tika-specific. There is no success-rate floor here because parsing is
+    local: a partial failure means malformed files, not a sick service.
+    """
+    candidates = counters["candidates"]
+    if candidates and not (counters["extracted"] + counters["empty"]):
+        msg = (
+            f"Refusing to publish transcript text for {course_id}: all "
+            f"{candidates} transcripts failed to parse. Publishing would be "
+            "indistinguishable from 'this course has no transcripts'."
+        )
+        raise RuntimeError(msg)
+
+
 @asset(
     key=AssetKey(("openedx", "processed_data", "course_transcript_text")),
     group_name="openedx",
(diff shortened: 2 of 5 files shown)


blarghmatey:
## Validated against real subtitles

Ran `build_transcript_rows` over every subtitle in `course-v1:MITx+14.310x+3T2021` (pulled from the QA landing zone). **The parsers came out clean on the first run:**

```
{'candidates': 1208, 'parsed': 1208, 'empty': 0, 'failed': 0}
```

- **1,208 transcripts parsed, none failed, none empty**
- Split **904 SJSON / 304 SRT** — confirming both paths are exercised by a single real course, and that SJSON is the dominant format
- **0 rows still containing `-->`** — every timestamp stripped, which is the thing most likely to go wrong silently
- Longest: 14,483 chars from `static/subs_14.310x_Lect4_Seg3_Final.srt.sjson`, reading as clean prose

The suffix-dispatch decision is vindicated by the real filenames: `subs_-GQxFdhr_qU.srt.sjson` (SJSON despite `.srt` in the middle) sits in the same directory as `14.310_Lecture_15_Segment_03-_Part1.srt` (SRT, with dots in the stem) and `14.310x Syllabus.pdf`. Last-suffix matching gets all three right; splitting on the first dot gets all three wrong.

## This validation also caught a defect in the layer below

`mimetypes` maps `.srt` → `text/plain`, which is in Tika's supported set, and a live Tika returns such a file *verbatim* — timestamps and all. So before the fix, all 304 SRT files would have been published **twice** against the same `file_path`: once as unusable noise from the document asset, once parsed properly here. The `integrations__learn__content_files` model above would have unioned both.

Fixed in the layer below (`33768737`), since it is the document extractor that needs to know what is not its own. Overlap between the two assets is now measured at **0**:

```
document paths     60
transcript paths   1,208
intersection       0
```

`SRT_SUFFIX`, `SJSON_SUFFIX` and `is_transcript` consequently moved into `content_files.py`, and this layer imports them — which is why the diff here is smaller than when this PR was first opened.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,234 @@
...
+    for relative_path, raw in members:
+        if not is_transcript(relative_path):
+            continue
+
+        counters["candidates"] += 1
+        text = transcript_text(relative_path, raw)
+
+        if text is None:
+            counters["failed"] += 1
+            continue
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 14ad64d. null, [], "str", 42, {"text": null} and {"text": "not a list"} all parse cleanly and then raised AttributeError/TypeError past the JSONDecodeError guard, aborting the course. text_from_sjson_content now type-checks the parsed value and the text array, and skips non-string captions rather than raising. Parametrized over all eight shapes, plus a test that one bad file doesn't lose the others.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,234 @@
...
+                "course_id": course_id,
+                "source_system": source_system,
+                "file_path": relative_path,
+                "content_type": (
+                    "application/json"
+                    if relative_path.lower().endswith(SJSON_SUFFIX)
+                    else "text/plain"
+                ),
+                "size_bytes": len(raw),
+                "content": text or None,
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 14ad64d. Status is decided on the stripped text, so content is now nulled on that same basis (`text if has_text else None`) rather than on `text or None`. A whitespace-only transcript is now empty status with null content, matching the document-row contract it's unioned with.

dg_projects/openedx/openedx/assets/content_files.py
```diff
@@ -73,6 +73,42 @@ def is_transcript(relative_path: str) -> bool:
...
+
+
+def bundle_data_version(bundle_location: UPath) -> str:
+    """Recover the content hash the bundle object key was named with.
+
+    Extracted text is a pure function of the archive plus the extractor, so
+    versioning on the bundle means re-extracting an unchanged bundle produces an
+    identical result.
+    """
+    return str(bundle_location).rsplit("/", 1)[-1].split(".")[0]
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 133aa4d — same fix as the #2584 thread. bundle_data_version is gone; both assets now version on the SHA-256 of their emitted JSONL via hashlib.file_digest, so a parser or Tika change on an unchanged bundle produces a new DataVersion and upstream_or_code_changes() consumers see it move.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Reproduces mit-learn's ``text_from_srt_content`` exactly, including the
+    sequence-number rule being a bare ``\\d+\\n`` -- which also removes a
+    caption line that happens to be only digits. That is a real quirk, kept on
+    purpose: diverging here would make every such caption a spurious difference
+    during parallel validation. Fix it in both repos together or not at all.
+    """
+    content = _SRT_TIMESTAMP.sub("", content)
+    content = _SRT_SEQUENCE.sub("", content)
+    return _SRT_BLANK_LINES.sub(" ", content)
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 14ad64d. The mit-learn parity fixture happens to contain no blank line between blocks, so _SRT_BLANK_LINES — the regex most likely to diverge — was the one nothing covered. Added a standard multi-block SRT case asserting the timings are gone, both captions survive, and no double newline remains.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,234 @@
...
+    course_id = context.partition_key
+
+    temp_files: list[Path] = []
+    try:
+        members = read_bundle_members(course_static_assets, temp_files)
+        rows, counters = build_transcript_rows(
+            members, course_id=course_id, source_system=source_system
+        )
+
+        output_file = Path(
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 14ad64d: the status is now "extracted", and the counter key with it. You're right that a shared schema whose status vocabulary differs by source isn't shared. Test pins it against the document asset's value.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,261 @@
...
+    context: AssetExecutionContext, course_static_assets: UPath
+):
+    """Parse every subtitle file in the course bundle into plain text."""
+    source_system = context.resources.openedx.deployment
+    course_id = context.partition_key
+
+    temp_files: list[Path] = []
+    try:
+        with open_bundle_members(course_static_assets, temp_files) as members:
+            rows, counters = build_transcript_rows(
```
mbertrand: The PR body says the analogous failure mode to #2584's health guard doesn't exist here, but I think that's only true of the rate floor. The total-failure clause in `assert_extraction_healthy` isn't about Tika — it refuses to publish "all N candidates failed" as an empty JSONL that downstream can't tell apart from "this course has no transcripts". That can happen in-process: a UTF-16 subtitle set fails every file on the `utf-8` decode (one authoring tool for a whole course is plausible — see the comment on line 115), and I confirmed `build_transcript_rows` returns `rows=[]` with `failed=N` and nothing between here and the `Output` refuses the publish. Porting just that clause is a few lines:

```python
if counters["candidates"] and not (counters["extracted"] + counters["empty"]):
    msg = (
        f"Refusing to publish transcript text for {course_id}: all "
        f"{counters['candidates']} transcripts failed to parse."
    )
    raise RuntimeError(msg)
```

Same tradeoff #2584 accepted: a one-transcript course with one bad file blocks until someone looks, which was judged better than silently publishing empty. If you'd rather leave it out, the body's "the analogous failure mode does not exist" claim should at least be narrowed to the rate floor.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,261 @@
...
+    for relative_path, read_bytes in members:
+        if not is_transcript(relative_path):
+            continue
+
+        counters["candidates"] += 1
+        raw = read_bytes()
+        text = transcript_text(relative_path, raw)
+
+        if text is None:
+            counters["failed"] += 1
```
mbertrand: Same question I left on #2584: a failed parse leaves no row, which downstream can't distinguish from the file having been removed from the course — Learn keeps that signal today via `failed_keys` exempting failures from its unpublish pass. Not asking to settle it in this PR, just flagging that both assets feed the same union model, so whichever way the failed-row decision lands on #2584 should land here in the same change.

dg_projects/openedx/openedx/assets/transcripts.py
mbertrand: If every transcript fails to parse (e.g. all UTF-16, per my other comment), this still publishes an empty JSONL that downstream reads as "no transcripts". The total-failure clause in `assert_extraction_healthy` isn't Tika-specific, so I think it's worth porting just that part:

```python
if counters["candidates"] and not (counters["extracted"] + counters["empty"]):
    msg = f"All {counters['candidates']} transcripts failed to parse for {course_id}"
    raise RuntimeError(msg)
```

dg_projects/openedx/openedx/assets/transcripts.py
mbertrand: Same issue I flagged on #2584 — a failed parse leaves no row, so downstream can't tell it from a deleted file. Whatever we decide there should apply to both assets, since they feed the same union model.

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def transcript_text(relative_path: str, raw: bytes) -> str | None:
+    """Parse one transcript file, dispatching on its suffix.
+
+    Returns None when the file is not a transcript, or when parsing failed --
+    the caller distinguishes the two by checking the suffix itself.
+    """
+    try:
+        content = raw.decode("utf-8")
```
mbertrand: Learn runs these files through Tika before parsing, and Tika does charset detection first — so a non-UTF-8 or BOM'd transcript would extract fine in Learn but fail here with no row. To see if that matters in practice, I scanned the latest production exports for 49 courses (the largest video-heavy ones plus a random sample, ~20k srt/sjson files): all UTF-8, no BOMs. So this looks like a non-issue for our content. `raw.decode("utf-8-sig")` is a cheap guard if a BOM ever shows up, but I wouldn't block on it.

dg_projects/openedx/openedx/assets/transcripts.py
mbertrand, line 263: If every transcript in a course fails to parse, this still publishes an empty JSONL that downstream reads as "no transcripts". The total-failure clause in `assert_extraction_healthy` isn't Tika-specific, so I think it's worth porting just that part:

```python
if counters["candidates"] and not (counters["extracted"] + counters["empty"]):
    msg = f"All {counters['candidates']} transcripts failed to parse for {course_id}"
    raise RuntimeError(msg)
```

dg_projects/openedx/openedx/assets/transcripts.py
mbertrand, line 162: Same issue I flagged on #2584 — a failed parse leaves no row, so downstream can't tell it from a deleted file.

dg_projects/openedx/openedx/assets/content_files.py
```diff
@@ -76,6 +77,53 @@ def is_transcript(relative_path: str) -> bool:
...
+    its own download rather than one depending on the other's output. That is
+    deliberate -- it keeps a Tika outage from also stopping transcripts.
+    """
+    bundle_path = Path(
+        NamedTemporaryFile(delete=False, suffix="_static_assets.tar.gz").name
+    )
+    temp_files.append(bundle_path)
+    bundle_location.fs.get_file(str(bundle_location), str(bundle_path))
+
+    with tarfile.open(bundle_path, "r:gz") as bundle:
```
mbertrand, line 111: Could use a test with a real tar.gz — nothing exercises this path now, which is how the `tarfile` import almost got dropped (per your Additional Context note).

mbertrand (changes requested):

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Fixed in d717879: decoding as utf-8-sig. Took it as insurance on your scan result rather than as a fix — the comment records that no BOMs turned up in the ~20k files you sampled.

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Fixed in d717879: assert_parsing_healthy ports the total-failure clause and runs before the JSONL is written. Left out the rate floor deliberately — parsing is local, so a partial failure means malformed files rather than a sick service.

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Fixed in d717879: a failed parse now gets a row with extraction_status "failed" and null content, same as the document rows in #2584 (f2c0305).

dg_projects/openedx/openedx/assets/content_files.py
blarghmatey: Fixed in d717879: test_open_bundle_members_reads_a_real_tarball builds a tar.gz, walks it through open_bundle_members, and checks the directory entry is skipped and the download is registered for cleanup.

mbertrand (approved):
👍

PR #2589, opened 2026-08-20
Close the ContentFile parity gaps and route WebVTT to the transcript parser

blarghmatey:
Fourth in the Cohort 4 Open edX extraction stack. Stacked on #2586.

Two fixes with the same root cause: this asset gates by **MIME type**, MIT Learn gates by **file extension**. The two agreeing was never a property, only a coincidence — and measured against real course exports, they didn't.

## 1. `.json` and `.tex` were being lost

`.json` maps to `application/json`, which was missing from Tika's supported set, so **four files in a real `MITx+14.310x+3T2021` export were extracted by MIT Learn and silently dropped here**. `.tex` maps to `application/x-tex` and would go the same way; it didn't appear in either archive sampled.

Both are now supported, and the tests parametrize over MIT Learn's `VALID_TEXT_FILE_TYPES` so the next divergence fails a test rather than losing files. `fake_is_supported` now gates on the real `SUPPORTED_CONTENT_TYPES` instead of a hand-picked subset that would have kept passing while the real set drifted.

## 2. WebVTT moves to the transcript parser

MIT Learn has no VTT transformer — `_transform_content_by_type` registers only `.srt` and `.sjson` — so a `.vtt` goes to Tika and is stored exactly as returned: `WEBVTT` header, cue timings, speaker tags and all.

Nothing reads those. The only `-->` anywhere in the mit-learn codebase is the SRT stripping regex, and the three consumers of transcript text — OpenSearch indexing (as an English-analysed text field), embedding (a plain recursive character splitter, markdown-aware only), and LLM summarisation — are none of them cue-aware. MIT Learn already strips timings from SRT and SJSON, which is **100% of the transcripts in both archives sampled**, so no consumer could have depended on cue timings.

This is a deliberate divergence rather than a parity fix, and is documented as one at both the module docstring and `text_from_vtt_content`.

The parser handles cue identifiers, `STYLE`/`REGION`/`NOTE` blocks, cue settings (`line:90% align:middle`), inline tags including mid-cue `<00:00:01.000>` timestamps, HTML entities, CRLF, and a BOM. Structure is excluded by one rule — a block is a cue iff it has a `-->` line — which works because the spec forbids `-->` in `NOTE` bodies.

## Correction from the previous revision of this PR

The first version also added `text/csv`, on the theory that Open edX needed it to serve tutor problems. **That was wrong.**

`.csv` appears only in `VALID_TUTOR_PROBLEM_FILE_TYPES`, which exists because Canvas requires an explicit mapping of which assets to process for tutor — its export is structurally different. Open edX was the *first* tutor target and needs no such mapping: `problem/` is a first-class OLX block-type directory, so problem documents are already carried by the block path with their XML intact.

Verified in both exports:

```
mitxonline MITx+14.310x+3T2021     782 problem blocks   (raw XML carries the question text)
xpro DEMO+SysEng1xDEMO             122 problem blocks
```

Extracting `.csv` here would have published course data files (gradebooks and the like) as course text for no consumer. It's now explicitly excluded, with a test pinning that.

## Coverage, measured end-to-end

The earlier revision compared only files inside the static-asset bundle, which structurally *cannot* reveal files the pipeline never collects. `documents_from_olx` walks the whole OLX tree with `os.walk`, so that was the wrong universe. Re-measured against it:

| | 14.310x | xpro SysEng1x |
|---|---|---|
| MIT Learn extracts (whole tree) | 3567 | 1530 |
| carried as blocks | 2028 | 756 |
| carried as transcripts | 1208 | 239 |
| carried as documents | 328 | 532 |
| **not carried** | **3** | **3** |

The three uncovered in each are `assets/assets.xml` (a manifest) and the two course metadata files, all handled by `process_course_xml()`. Transcripts we parse that MIT Learn skips: 0.

## Also

`file_extension` is now published on every document row. It's a `ContentFile` field on MIT Learn's side already, carried through `documents_from_olx` metadata, so publishing it saves the consumer re-deriving what this asset knows — and lets a consumer apply its own extension allowlist without mapping back from a MIME string.

## Tests

93 pass in `openedx_tests`. New coverage: the parametrized allowlist parity check, `.csv` and `.vtt` exclusion from the document set, and ten WebVTT cases pinning what "parsed properly" means so the divergence stays intentional rather than drifting.

### DIFF
--- a/dg_projects/openedx/openedx/assets/content_files.py
+++ b/dg_projects/openedx/openedx/assets/content_files.py
@@ -14,6 +14,17 @@ Data flow:
 Only documents Tika can parse are handled here. Video transcripts are a
 separate asset: they need no external service, and pairing them with Tika would
 mean a Tika outage also stopped transcript extraction.
+
+What counts as a document here is aligned to MIT Learn's VALID_TEXT_FILE_TYPES,
+but the two systems reach that set differently -- MIT Learn filters by file
+extension, this asset by MIME type -- so agreement is a property that has to be
+asserted rather than assumed. The tests parametrize over MIT Learn's list to
+enforce it.
+
+Note that Open edX course *problems* do not arrive here at all. `problem/` is a
+first-class OLX block-type directory, so problem documents are carried by the
+block path with their XML intact; a real 14.310x export has 782 of them. This
+asset covers the file-storage side of the archive, not the block side.
 """
 
 import hashlib
@@ -62,9 +73,18 @@ MIN_DOCUMENTS_FOR_RATE_CHECK = 5
 # this exclusion every one of the `.srt` would be extracted twice: once as
 # unusable timestamp-laden "text" here, and once properly by the transcript
 # asset. The transcript parser owns them.
+#
+# `.vtt` joins them for a different reason. `mimetypes` maps it to `text/vtt`,
+# which Tika would also accept and return verbatim -- which is exactly what MIT
+# Learn stores today, since its `_transform_content_by_type` has transformers
+# only for `.srt` and `.sjson`. WebVTT cue timings and speaker tags are noise to
+# every consumer of this text (search indexing, embedding, summarisation), so it
+# is routed to the transcript parser instead of being published as "document
+# text" that happens to be two-thirds timestamps.
 SRT_SUFFIX = ".srt"
 SJSON_SUFFIX = ".sjson"
-TRANSCRIPT_SUFFIXES = frozenset({SRT_SUFFIX, SJSON_SUFFIX})
+VTT_SUFFIX = ".vtt"
+TRANSCRIPT_SUFFIXES = frozenset({SRT_SUFFIX, SJSON_SUFFIX, VTT_SUFFIX})
 
 
 def is_transcript(relative_path: str) -> bool:
@@ -122,6 +142,19 @@ def open_bundle_members(
                 yield member.name, _read
 
         yield _members()
+
+
+def file_extension(relative_path: str) -> str:
+    """Return a bundle member's lowercased final suffix, or "" if it has none.
+
+    Published on every row for two reasons. It is a `ContentFile` field on MIT
+    Learn's side, carried through `documents_from_olx` metadata, so publishing it
+    saves the consumer re-deriving what this asset already knows. And MIT Learn
+    selects files by EXTENSION where this asset gates by MIME type, so a consumer
+    that needs to apply its own allowlist can do so directly rather than mapping
+    back from a MIME string.
+    """
+    return Path(relative_path).suffix.lower()
 
 
 def _content_type(relative_path: str) -> str:
@@ -265,6 +298,7 @@ def build_document_rows(
                     "course_id": course_id,
                     "source_system": source_system,
                     "file_path": relative_path,
+                    "file_extension": file_extension(relative_path),
                     "content_type": content_type,
                     "size_bytes": 0,
                     "content": None,
@@ -314,6 +348,7 @@ def build_document_rows(
                 "course_id": course_id,
                 "source_system": source_system,
                 "file_path": relative_path,
+                "file_extension": file_extension(relative_path),
                 "content_type": content_type,
                 "size_bytes": len(file_bytes),
                 "content": text or None,
--- a/dg_projects/openedx/openedx/assets/content_files.py
+++ b/dg_projects/openedx/openedx/assets/content_files.py
@@ -157,13 +157,25 @@ def file_extension(relative_path: str) -> str:
     return Path(relative_path).suffix.lower()
 
 
+# Pinned to Python's builtin table rather than the module-level lookup, which
+# reads /etc/mime.types when the host has one. That file is present on a dev
+# machine and on ubuntu-latest but absent from the python:3.14-slim runtime
+# image, so the same file resolved to different MIME types in dev and in
+# production: `.rtf` was `application/rtf` locally and `text/rtf` in the
+# cluster, and only one of those was in SUPPORTED_CONTENT_TYPES. Every RTF MIT
+# Learn extracts today would have been skipped in production with no row and no
+# error, while the parity test passed in the one environment nobody runs
+# production in.
+_MIMETYPES = mimetypes.MimeTypes(filenames=())
+
+
 def _content_type(relative_path: str) -> str:
     """Guess a document's MIME type from its path.
 
     The bundle manifest carries the same value, but it is a sibling S3 object;
     deriving it here keeps this asset dependent on the archive alone.
     """
-    mime_type, _ = mimetypes.guess_type(relative_path)
+    mime_type, _ = _MIMETYPES.guess_type(relative_path)
     return mime_type or "application/octet-stream"
 
 
--- a/dg_projects/openedx/openedx/assets/transcripts.py
+++ b/dg_projects/openedx/openedx/assets/transcripts.py
@@ -10,18 +10,26 @@ Data flow:
             -> openedx/processed_data/course_transcript_text  (JSONL in S3)
                 -> integrations__learn__content_files         (dbt)
 
-The two parsers below reproduce mit-learn's
+The SRT and SJSON parsers below reproduce mit-learn's
 ``learning_resources/etl/utils.py:text_from_srt_content`` and
 ``text_from_sjson_content`` exactly, including their quirks. Their test table is
 ported verbatim into this project's tests so the parity is enforced rather than
 asserted -- a cutover diff should show a difference in the data, never in how
 the two sides parse it.
+
+The WebVTT parser is the one deliberate exception. MIT Learn has no VTT
+transformer, so it stores whatever Tika returns for a `.vtt`: the `WEBVTT`
+header, every cue timing, and the speaker tags. That is not text anyone wants
+indexed, and no consumer reads the timings, so `.vtt` is routed here and parsed
+properly rather than published as document text. See `text_from_vtt_content` for
+the evidence behind that call.
 """
 
 import json
 import logging
 import re
 from collections.abc import Callable, Iterable
+from html import unescape
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any
@@ -41,6 +49,8 @@ from upath import UPath
 from openedx.assets.content_files import (
     SJSON_SUFFIX,
     SRT_SUFFIX,
+    VTT_SUFFIX,
+    file_extension,
     is_transcript,
     open_bundle_members,
     output_digest,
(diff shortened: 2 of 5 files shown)


dg_projects/openedx/openedx/assets/content_files.py
```diff
@@ -173,6 +206,7 @@ def build_document_rows(
                     "course_id": course_id,
                     "source_system": source_system,
                     "file_path": relative_path,
+                    "file_extension": file_extension(relative_path),
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in e2beafb. Transcript rows now carry file_extension, so the union in integrations__learn__content_files no longer reads null for every transcript-derived row. Parametrized over .sjson, .srt and .vtt — the last mattering most, as you noted, since .vtt is newly routed here.

packages/ol-orchestrate-lib/src/ol_orchestrate/resources/tika.py
```diff
@@ -41,6 +64,9 @@
         "application/epub+zip",
         "application/xml",
         "text/xml",
+        "application/json",
+        "application/x-tex",
+        "text/x-tex",
```
mbertrand, line 75: Following up my earlier comment (https://github.com/mitodl/ol-data-platform/pull/2584#discussion_r3882740658): this adds `application/json` and both `.tex` spellings but not `text/rtf`, so the `.rtf` regression is still live. I checked in the actual prod base image —

```
docker run --rm python:3.14-slim-bookworm python -c \
  'import mimetypes, os; print(os.path.exists("/etc/mime.types"), mimetypes.guess_type("x.rtf"))'
False ('text/rtf', None)
```

— no `/etc/mime.types` there, `.rtf` maps to `text/rtf`, which isn't in this set, so every RTF Learn extracts today lands in `counters["skipped"]` in production with no row and no error. Dev machines and ubuntu-latest CI do have `/etc/mime.types`, resolve `.rtf` to `application/rtf`, and `test_every_extension_mit_learn_extracts_is_supported` passes — it calls `_content_type`, so it inherits the host's mime DB and only fails in the one environment nobody runs pytest in.

Two-part fix: add `text/rtf` here, and pin `_content_type` to the builtin table so dev == CI == prod and the test actually enforces prod behavior:

```python
_MIMETYPES = mimetypes.MimeTypes(filenames=())

def _content_type(relative_path: str) -> str:
    mime_type, _ = _MIMETYPES.guess_type(relative_path)
    return mime_type or "application/octet-stream"
```

(the kwarg is `filenames`, not `files` as I had it in the earlier comment). With the pin the builtin table only ever produces `application/x-tex`, so the both-spellings story simplifies too — keeping both in the set is cheap insurance though. Worth re-checking `text/rtf` against your local Tika the way you did for json/tex, since the pin changes the Content-Type header it's sent with.

dg_projects/openedx/openedx/assets/content_files.py
```diff
@@ -14,6 +14,17 @@
...
+What counts as a document here is aligned to MIT Learn's VALID_TEXT_FILE_TYPES,
+but the two systems reach that set differently -- MIT Learn filters by file
+extension, this asset by MIME type -- so agreement is a property that has to be
+asserted rather than assumed. The tests parametrize over MIT Learn's list to
+enforce it.
+
+Note that Open edX course *problems* do not arrive here at all. `problem/` is a
+first-class OLX block-type directory, so problem documents are carried by the
+block path with their XML intact; a real 14.310x export has 782 of them. This
+asset covers the file-storage side of the archive, not the block side.
```
mbertrand, line 27: Learn doesn't just keep the XML for these though — it runs them through Tika and stores the extracted text. So when Learn cuts over to this data, problem/html text only survives if the dbt model pulls text out of `raw_xml`, and I don't think that's written down anywhere yet. Maybe a note on the #2582 issue so it doesn't get lost. (Related: the `.xml` case in `test_every_extension_mit_learn_extracts_is_supported` can never fail, since no XML file ever reaches the MIME gate.)

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -178,6 +242,11 @@ def build_transcript_rows(
...
                 "source_system": source_system,
                 "file_path": relative_path,
+                # Both row sources are unioned into one model downstream, so a
+                # field present on document rows and absent here is a schema
+                # inconsistency, not a harmless omission -- every
+                # transcript-derived row would read as a null extension.
+                "file_extension": file_extension(relative_path),
                 "content_type": (
                     "application/json"
                     if relative_path.lower().endswith(SJSON_SUFFIX)
```
mbertrand: `.vtt` rows land in the `text/plain` arm, while the document asset derives content_type from mimetypes, which maps `.vtt` to `text/vtt` in both the host and builtin tables — so the union gets an inconsistent convention. One more branch fixes it:

```python
"content_type": (
    "application/json"
    if relative_path.lower().endswith(SJSON_SUFFIX)
    else "text/vtt"
    if relative_path.lower().endswith(VTT_SUFFIX)
    else "text/plain"
),
```

dg_projects/openedx/openedx/assets/transcripts.py
```diff
@@ -105,6 +115,58 @@ def text_from_sjson_content(content: str) -> str | None:
     return " ".join(caption for caption in captions if isinstance(caption, str))
 
 
+_VTT_BLOCK_SEPARATOR = re.compile(r"\r?\n[ \t]*\r?\n")
+_VTT_TAG = re.compile(r"</?[^>]*>")
+
+
+def text_from_vtt_content(content: str) -> str:
```
mbertrand, line 122: `text_from_vtt_content` never returns None, so a garbage `.vtt` records as `empty` where a malformed sjson records as `failed`. Defensible — there's no cheap VTT validity check — but worth a sentence in the docstring since the two parsers differ here.

dg_projects/openedx/openedx/assets/transcripts.py
mbertrand, line 1: `failed` is counted but the asset publishes unconditionally — see previous comment: https://github.com/mitodl/ol-data-platform/pull/2586#discussion_r3883209322
If it's not landing there, this alignment branch is the natural place.

dg_projects/openedx/openedx/assets/content_files.py
```diff
@@ -263,6 +296,7 @@ def build_document_rows(
                     "course_id": course_id,
                     "source_system": source_system,
                     "file_path": relative_path,
+                    "file_extension": file_extension(relative_path),
```
mbertrand, line 313: This PR adds another field to the row shape — seems like a good moment to also settle whether a failed extraction gets a row, before the dbt model freezes the schema. See previous comment: https://github.com/mitodl/ol-data-platform/pull/2584#discussion_r3882739250

mbertrand (changes requested):

packages/ol-orchestrate-lib/src/ol_orchestrate/resources/tika.py
blarghmatey: Fixed in 9db0364: text/rtf added, and _content_type pinned to mimetypes.MimeTypes(filenames=()). Reproduced the split on this host — module-level guess_type gives application/rtf from /etc/mime.types, the pinned table gives text/rtf. test_the_mime_lookup_does_not_depend_on_the_host now pins that so the parity test enforces prod behaviour rather than the host's. Kept both .tex spellings and application/rtf for callers reading the host table. One thing I could not check: text/rtf against a live Tika — no local instance here, so that half is unverified.

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Fixed in 9db0364: transcript rows now go through _transcript_content_type, which spells .vtt as text/vtt, and test_transcript_rows_use_the_same_content_type_spelling pins the union's three cases.

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Fixed in 9db0364: the text_from_vtt_content docstring now names the asymmetry — no failure return, so a garbage .vtt records as empty where a malformed sjson records as failed, because there is no cheap VTT validity check and "no --> anywhere" is indistinguishable from a file that is only headers and comments.

dg_projects/openedx/openedx/assets/transcripts.py
blarghmatey: Landed in #2586 (d717879): assert_parsing_healthy ports the total-failure clause and runs before the JSONL is written.

dg_projects/openedx/openedx/assets/content_files.py
blarghmatey: Settled in #2584 (f2c0305) and #2586 (d717879): a failed extraction gets a row with extraction_status "failed" and null content, on both row sources, before the dbt model freezes the schema.

dg_projects/openedx/openedx/assets/content_files.py
blarghmatey: Recorded on #2582 as a §9 constraint: https://github.com/mitodl/ol-data-platform/pull/2582#issuecomment-5458050352 — integrations__learn__content_files has to pull the text out of raw_xml for problem/html, since Learn Tika-extracts those today and no XML member is ever in the bundle. The .xml parity case is fixed in 9db0364: excluded by name as _NEVER_IN_THE_BUNDLE with the reason, since it could not fail.

dg_projects/openedx/openedx_tests/test_content_files.py
```diff
@@ -398,3 +404,118 @@ def test_output_digest_is_stable_for_identical_output(tmp_path):
...
+
+    The two systems filter by different mechanisms -- MIT Learn by extension,
+    this asset by MIME type -- so agreement between them is a coincidence
+    unless it is asserted. `.json` was genuinely lost before this check
+    existed: four files in a real course-v1:MITx+14.310x+3T2021 export. `.tex`
+    did not appear in the archives sampled, but maps to `application/x-tex`,
+    which was missing from the supported set for the same reason.
+    """
+    content_type = _content_type("problem_set" + extension)
+    assert content_type in SUPPORTED_CONTENT_TYPES, (
```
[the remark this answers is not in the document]
blarghmatey: Confirmed the mechanics — .xls/.xlsx map to application/vnd.ms-excel and the spreadsheetml sheet type, both in SUPPORTED_CONTENT_TYPES, and MIT Learn's documents_from_olx gates on VALID_TEXT_FILE_TYPES, which excludes them. So this asset does publish a superset of what Learn extracts. Not adding an extension gate: those two MIME types are in the shared Tika set that OCW also uses, so unlike .csv this cannot be fixed at the set level, and an asset-local allowlist would mean two filters to keep in sync with Learn instead of one. Every row carries file_extension so a consumer applies its own allowlist — test_rows_carry_the_extension_consumers_filter_on pins that.

dg_projects/openedx/openedx_tests/test_content_files.py
```diff
@@ -398,3 +404,118 @@ def test_output_digest_is_stable_for_identical_output(tmp_path):
     two.write_text('{"content": "same"}\n')
 
     assert output_digest(one) == output_digest(two)
+
+
+# Verbatim from mit-learn learning_resources/constants.py. Reproduced rather
+# than imported because mit-learn is a separate deployment with no shared
+# package; these tests are what catches the two drifting apart.
+VALID_TEXT_FILE_TYPES = frozenset(
```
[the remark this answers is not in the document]
blarghmatey: Not doing this here. Sourcing the list from a shared artifact needs a package or a published contract that does not exist between the two deployments, and a CI job reaching into mit-learn is a cross-repo mechanism that belongs in its own change rather than in this branch. The copied snapshot plus the parity test is what is affordable now, and it has already earned itself: it caught the missing .json and .tex types, and this round the .rtf host/prod split.

mbertrand (approved):
👍

PR #2604, opened 2026-08-25
feat(feedback): implement feedback summaries asset and job

### LINKED ISSUE #2542: Conversation summarization asset + skip rule (the one per-record LLM cost)
### Description/Context

Generate `conversation_summary` on `afact_feedback_conversation` from the assembled redacted turns.

It serves two purposes: it is what a human reads in a cluster listing instead of scrolling a thread, and it is a candidate for `embedding_input` — the lever the 2026 support-ticket-clustering literature identifies as the largest single contributor to cluster quality on short, noisy text.

**This is the one per-record LLM call in the design.** Everything else is per-cluster (hundreds of calls) or per-corpus. Treat the cost as a first-class requirement, not an afterthought.

### Plan/Design

- **Skip rule** — never summarize `turn_count = 1` or conversations under the length threshold; the raw text already *is* the summary. `summary_model_version` stays null and `embedding_input` becomes `concatenated_turns`. This makes ORA and the edX plugin free by construction.
- **Incremental** — key on `feedback_conversation_pk` and only summarize new or changed conversations. With declarative automation on the upstream, a naive full re-run re-pays for the entire corpus on every refresh.
- **Checkpoint** — a failed run must not re-pay for work already done.
- **Measure first** — the low-hundreds-of-dollars estimate is unvalidated; sample-price it before the backfill.
- **PII** — generate from Presidio-redacted text only. The summary inherits that classification and is governed like `feedback_text`, not like an aggregate.

**Fallback** if the measured cost lands materially above estimate: summarize only the conversations inside clusters a human is actually reading, rather than the whole corpus.

Spec: [`feedback_ml_approach.md`](https://github.com/mitodl/ol-data-platform/blob/worktree-feedback-aggregation-spec/docs/design/feedback_ml_approach.md) §A.1, [`feedback_dimensional_model.md`](https://github.com/mitodl/ol-data-platform/blob/worktree-feedback-aggregation-spec/docs/design/feedback_dimensional_model.md) §5b

rachellougee:

Close https://github.com/mitodl/ol-data-platform/issues/2542

Builds the `feedback_summaries` Dagster asset 
  - Reads `int__feedback__conversation`, calls an LLM to generate `conversation_summary` per multi-turn conversation
  - **Skip rule**: single-turn conversations, and multi-turn ones under 500 characters, are not summarized
  - **Incremental, and change-awareness**: only summarizes new or changed conversations (a ticket that gains a new comment gets re-summarized, not skipped forever).                                                                                               
  - Supports `Anthropic`, `OpenAI`, and `AWS Bedrock` as the LLM backend (needs to confirm with @blarghmatey if Bedrock is configured on production)                                                                                                                      
  - Fixes a PII-redaction bug where a URL could be misread and redacted as a person's name.                                                                              

### How can this be tested?

`int__feedback__conversation` must exist in your dev schema. If it doesn't, copy it from production:       
```                                                                                                                                          
  create table ol_data_lake_production.ol_warehouse_production_<suffix>_intermediate.int__feedback__conversation                                                 
  as                                                                                                                                                             
  select * from ol_data_lake_production.ol_warehouse_production_intermediate.int__feedback__conversation;  
```

- Run the unit test suite: 
 ```                                                                                                                                   
  cd dg_projects/ml                                                                                                                                              
  uv run pytest ml_tests/ -q 
```                                                                                                                

- Run the asset locally with your Parley key at https://platform.parley.mit.edu/my-keys            
```                                                                   
  export DAGSTER_ENV=dev                                                                                                                           
  export ANTHROPIC_API_KEY =...                                                                                                                                   
  export ANTHROPIC_BASE_URL=https://parley.api.mit.edu

  uv run dagster dev -m ml.definitions    
``` 
                                                                                                                                                                 
 - In the Dagster UI, materialize `feedback_summaries` with a small sample_limit via Launchpad config:     
     
 ```                                                  
  ops:                                                                                                                                                           
    feedback_summaries:                                                                                                                                          
      config:                                                                                                                                                    
        sample_limit: 100      
```                                                                                                                                  

 Expected log line: `Summarized 100 new conversations (0 already summarized, 100 total upstream).`   
                                                                                                                                              
- (Optional) Turn on the automation sensor at http://127.0.0.1:3000/locations/ml.definitions/sensors/feedback_summaries_automation_sensor. 
 The sensor is off by default and this is only needed once Bedrock/API access is confirmed working and you want it to run automatically. 

Query the output table 
```
select * from ol_data_lake_production."ol_warehouse_production_rlougee_intermediate".feedback_summaries
```  

### Additional Context

For deployment step, do a Full refresh — feedback_redacted  to fixes historical mis-redactions:                                                                                             
```                                                                                                                                                  
  ops:                                                                                                                                                           
    feedback_redacted:                                                                                                                                           
      config:                                                                                                                                                    
        full_refresh: true      
```


dg_projects/ml/ml/lib/summarize.py
```diff
@@ -0,0 +1,164 @@
...
+            all of them; conversation_summary itself stays null for skipped rows,
+            and summary_model_version is the "was this LLM-generated" signal.
+    """
+    rows = df.to_dicts()
+    summaries: list[str | None] = []
+    model_versions: list[str | None] = []
+    embedding_inputs = []
+    for row in rows:
+        if needs_summary(row):
+            summaries.append(client.summarize(row["conversation_text"]))
```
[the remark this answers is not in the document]
rachellougee: `tfact_feedback` already joins the redacted text source on main branch, this branch just hadn't rebased yet.  Not an issue after rebasing main in.

dg_projects/ml/ml/assets/feedback_summaries.py
```diff
@@ -0,0 +1,105 @@
...
+            "candidates instead of re-hitting already-summarized rows."
+        ),
+    )
+
+
+@asset(
+    code_version="feedback_summaries_v1",
+    group_name="feedback",
+    key=AssetKey(["intermediate", "feedback_summaries"]),
+    deps=[AssetKey(["intermediate", "int__feedback__conversation"])],
```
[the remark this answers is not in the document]
rachellougee: the dbt changes will be a separate PR, not in this PR.

dg_projects/ml/ml/lib/summarize.py
```diff
@@ -0,0 +1,234 @@
...
+# so switching client_class to "openai"/"openai_compatible" requires overriding
+# this to a matching id (e.g. "gpt-4o-mini") -- there is no one id valid everywhere.
+SUMMARY_MODEL_VERSION = os.environ.get("SUMMARY_MODEL_VERSION", "claude-haiku-4-5")
+
+SUMMARY_PROMPT = (
+    "Summarize the following support conversation from the requester's point of "
+    "view. Focus on the problem reported and its resolution if one is present. "
+    "Do not include names or contact details.\n\n{conversation_text}"
+)
+
```
blarghmatey, line 54: This is fine for now, but it would be good to move this into an Opik managed prompt - https://www.comet.com/docs/opik/development/prompt-library/overview

We have an instance of Opik deployed at opik.ol.mit.edu (as well as CI and QA instances) and learn-ai is currently wired into it. We should also use it to instrument any other LLM calls that we are making throughout this project/repository.

dg_projects/ml/ml/lib/summarize.py
```diff
@@ -0,0 +1,234 @@
...
+                    "content": SUMMARY_PROMPT.format(
+                        conversation_text=conversation_text
+                    ),
+                }
+            ],
+        )
+        return message.content[0].text
+
+
+class OpenAISummaryClient:
```
blarghmatey, line 91: We'll also want to update this to be compatible with the Azure OpenAI endpoints

dg_projects/ml/ml/lib/summarize.py
rachellougee: I will write Opik into this after other steps are settled. There are too many steps to wire that in now:  Vault secret and project setup, plus this Opik deployment sits behind Keycloak OIDC, not a static API key, so real auth might needs a provisioned Keycloak client before it can work at all

dg_projects/ml/ml/lib/summarize.py
rachellougee: Addressed in https://github.com/mitodl/ol-data-platform/pull/2604/changes/521a8ec3769f72421c86ba9610a6aa16bee6893e

quazi-h (commented):
Reviewed the redaction change and the new summarize path. The first of these is a redaction regression I'd normally raise as **request-changes**, but I'm leaving this as a comment so you can push back or correct me rather than being gated on it.

Checked against the existing Copilot and @blarghmatey threads to avoid repeating anything.

dg_projects/ml/ml/lib/redact.py
```diff
@@ -30,24 +29,37 @@ def _get_anonymizer() -> AnonymizerEngine:
...
+        (result.start, result.end)
+        for result in results
+        if result.entity_type in EXCLUDED_ENTITIES
+    ]
+    # Containment, not any overlap: a PII match that only partially overlaps an
+    # excluded span (e.g. a name immediately followed by a URL) extends beyond it
+    # and is still real PII outside that span -- a one-character overlap must not
+    # exempt the whole match. Only a result fully inside an excluded span is the
+    # "NER misread a URL/date" case this is meant to catch.
+    filtered_results = [
```
quazi-h, line 73: Containment is right for the NER case in the comment above, but it also exempts **pattern-derived** PII that legitimately sits inside a URL.

Presidio's `UrlRecognizer` regex ends in `(?:/[^\s()<>"']*)?`, so a URL span swallows the path *and* query string. With the pinned `presidio-analyzer 2.2.362`:

```
"...see https://mitxonline.mit.edu/support?email=jane.doe@mit.edu&phone=617-555-0143"
   URL    (29, 105)
   EMAIL  (70,  86)  <- fully contained -> dropped
   PHONE            <- fully contained -> dropped
```

Both get written unredacted. Before this change `URL` wasn't in `analyze(entities=...)`, so no URL span existed and both were redacted. `https://openlearning.mit.edu/u/jane.doe@mit.edu` fails the same way.

Support text pasting a reset link or profile URL is exactly this asset's traffic, and the `full_refresh` discussed on your other thread would spread it across history.

Suggested narrowing - exempt NER-derived types only:

```python
NER_DERIVED = {"PERSON", "LOCATION", "NRP"}

filtered_results = [
    result
    for result in results
    if result.entity_type not in EXCLUDED_ENTITIES
    and not (
        result.entity_type in NER_DERIVED
        and any(start <= result.start and result.end <= end
                for start, end in excluded_spans)
    )
]
```

Related: Copilot's note on `test_redact.py:43` means the URL test can't catch this today - the fake only replaces the literal `PII`, which the URL input doesn't contain, so it passes either way.

dg_projects/ml/ml/lib/summarize.py
```diff
@@ -0,0 +1,234 @@
...
+            messages=[
+                {
+                    "role": "user",
+                    "content": SUMMARY_PROMPT.format(
+                        conversation_text=conversation_text
+                    ),
+                }
+            ],
+        )
+        return response.choices[0].message.content
```
quazi-h, line 120: `summarize` is annotated `-> str`, but `message.content` is `Optional[str]` - null on a refusal or content-filter finish.

The `None` reaches `summaries.append(...)` while `model_versions` and `embedding_inputs` still populate, so the row lands with a null summary but a non-null `summary_model_version`. Since `filter_unsummarized` keys off `turn_count` + `summary_model_version`, that row is **never retried**, and downstream treats it as successfully summarized.

Treating a falsy response as a failure (`continue`, same as the exception path) would cover it.

dg_projects/ml/ml/lib/summarize.py
```diff
@@ -0,0 +1,234 @@
...
+    conversation_refs: list[str] = []
+    turn_counts: list[int] = []
+    summaries: list[str | None] = []
+    model_versions: list[str | None] = []
+    embedding_inputs: list[str] = []
+    for row in rows:
+        if needs_summary(row):
+            try:
+                summary = client.summarize(row["conversation_text"])
+            except Exception:
```
quazi-h: This handler is per-row with no failure counter, so a *total* LLM outage - rotated Vault key, missing Bedrock IAM policy, bad `SUMMARY_MODEL_VERSION` - materializes the asset **green with zero rows**.

The log then reads `Processed 0 conversations (0 LLM calls, 0 skipped...)`, which is indistinguishable from the normal steady state. Given the PR body notes Bedrock access on production is unconfirmed, that's a plausible first production outcome.

A `failed` count in the asset metadata, or failing the run when every attempted row failed, would make it visible. (Distinct from Copilot's `feedback_summaries.py:96` point - that's about re-paying for a batch on retry; this is about the failure being invisible.)

dg_projects/ml/ml/lib/redact.py
rachellougee: Fixed, but not with NER_DERIVED = {"PERSON", "LOCATION", "NRP"}, that reintroduces a different bug.                                                                                  
                                                                                                                                                                 
 The point of exempting URLs is to preserve urls, like https://micromasters.mit.edu/dedp/learners/ and                                               
  https://mitxonline.mit.edu/records/shared/03bf36fe-77e1-4fd5-a115, as these are helpful feedback refererences. NER_DERIVED would redact the whole URL in the first one and break the url for the second one.

 After some testing, ALWAYS_REDACT_EVEN_IN_URL = {"EMAIL_ADDRESS", "PHONE_NUMBER"} is sufficient. A broader denylist defeats the point of keeping the URL in the first place. That said, if that raise concern, then I would just remove url completely.

quazi-h: Agreed, and your fix is better than what I suggested. I'd missed that the built-in `EmailRecognizer` allows `/ ? = &` in the local part, so `NER_DERIVED` would have over-matched from the URL itself and eaten the whole link — exactly the thing the exemption exists to prevent.

The strict pattern plus `{EMAIL_ADDRESS, PHONE_NUMBER}` is the right trade. Please don't drop the URL exemption; those reference links are worth keeping, and the broader pattern recognizers really do false-positive on UUIDs and course IDs.

One note, no action needed: the new test asserts on `_STRICT_EMAIL_RECOGNIZER` directly and monkeypatches `_get_analyzer`, so it never exercises the `remove_recognizer("EmailRecognizer")` line — if that name were ever wrong it would silently no-op and leave the loose recognizer in place. I ran it against presidio to confirm the name resolves and the recognizer is actually removed, so it's correct as written.

Resolving. Thanks for the thorough fix on the other two as well — the `raise Failure` on 100% failure is exactly what I was after.

quazi-h (approved):
Approving pending Opik thread resolution (deferring to Tobias). All of my feedback has been addressed, and the redaction fix looks great.

rachellougee:
Thanks for the review and feedback.

I made 3 changes to improve local testing (https://github.com/mitodl/ol-data-platform/pull/2604/changes/3e56bf245c3fe60f6f8659507bc78545a4c67efb) and tested them. No need to re-review, so I'm going to merge this

As for Opik: once the basic pipeline is confirmed in production, I'll open a separate PR to move the prompt into an Opik-managed prompt. I'd rather keep dependencies to a minimum for now.

blarghmatey:
Flagging a mismatch between this PR's Bedrock client and the IAM grant that authorizes it.

[`ml/resources/llm.py`](https://github.com/mitodl/ol-data-platform/blob/1f9700680e894c90c8e6956b2e27b8ffdcd93b22/dg_projects/ml/ml/resources/llm.py#L6) instantiates [`AnthropicBedrockMantle`](https://github.com/mitodl/ol-data-platform/blob/1f9700680e894c90c8e6956b2e27b8ffdcd93b22/dg_projects/ml/ml/resources/llm.py#L64) for `client_class="bedrock"`. That client talks to AWS's newer "Bedrock Mantle" service (`bedrock-mantle.{region}.api.aws`, IAM prefix `bedrock-mantle`) — a separate service from classic Bedrock (`bedrock-runtime`, IAM prefix `bedrock`), with its own IAM actions/resource model (`project` ARNs, not `foundation-model`/`inference-profile`).

The Dagster IRSA role this asset runs under is shared across every code location, so the IAM side needs a repo-wide grant either way (mitodl/ol-infrastructure#5750). We initially drafted that grant against `bedrock-mantle`, then reconsidered: per [AWS's own docs](https://docs.aws.amazon.com/bedrock/latest/userguide/models-endpoint-availability.html), Bedrock Mantle currently lacks Guardrails, cross-Region inference, and intelligent prompt routing, and AWS recommends the classic `bedrock-runtime` endpoint for new applications. #5750 now grants classic `bedrock:InvokeModel`/`InvokeModelWithResponseStream` instead (unscoped by model vendor, since `SUMMARY_MODEL_VERSION` is a runtime env var here).

With that IAM change, `client_class="bedrock"` needs to instantiate `AnthropicBedrock` (classic) instead of `AnthropicBedrockMantle` — as written today, production would get `AccessDeniedException` since the code and the grant point at two different AWS services. `AnthropicBedrock` takes the same `aws_region` kwarg your `get_client()` already passes, so the fix should be a one-line swap of the class plus a matching update to `supported_client_class`.

cc @rachellougee

rachellougee:
https://github.com/mitodl/ol-data-platform/pull/2639 is created to fix the Bedrock client issue

PR #2617, opened 2026-08-28
add missing records to combined products

KatelynGit:
`###` What are the relevant tickets?
https://github.com/mitodl/hq/issues/12649

add missing records to combined products

### How can this be tested?

### DIFF
--- a/src/ol_dbt/models/marts/combined/marts__combined__products.sql
+++ b/src/ol_dbt/models/marts/combined/marts__combined__products.sql
@@ -55,6 +55,23 @@ with mitxonline_product as (
     select * from {{ ref('int__edxorg__mitx_product') }}
 )
 
+, edxorg_paid_product as (
+    select
+        courserun_readable_id
+        , price
+        , upgrade_deadline
+        , row_number() over (
+            partition by courserun_readable_id
+            order by
+                if(courserun_mode = 'verified', 0, 1)
+                , price desc nulls last
+                , courserun_mode
+                , upgrade_deadline desc nulls last
+        ) as _row_num
+    from edxorg_product
+    where courserun_mode not in ('audit', 'honor')
+)
+
 , edxorg_runs as (
     select * from {{ ref('int__edxorg__mitx_courseruns') }}
 )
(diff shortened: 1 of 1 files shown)


quazi-h (commented):
One finding below. I'd normally raise it as **request-changes**, but leaving it as a comment so you can push back or tell me the two-platform listing is intentional.

Skipping the `list_price` / `upgrade_deadline` point - Copilot already covered it on line 214 and I agree with its suggested fix, no need for a second copy.

src/ol_dbt/models/marts/combined/marts__combined__products.sql
```diff
@@ -206,6 +206,39 @@ with mitxonline_product as (
...
+        , edxorg_runs.course_topics as topics
+        , edxorg_runs.courserun_instructors as instructors
+        , if(edxorg_runs.courserun_is_published, true, false) as is_live
+        , if(
+            edxorg_runs.micromasters_program_id is not null
+            , 'MicroMasters Credential'
+            , 'Certificate of Completion'
+        ) as certification_type
+    from edxorg_runs
+    left join mitxonline_product_view
```
quazi-h: The `mitxonline_product_view` anti-join can almost never match, so this guard is close to dead.

`edxorg_runs.courserun_readable_id` is old-form `{org}/{course}/{run}` (`int__edxorg__mitx_courseruns.sql:66`); `mitxonline_product_view.product_readable_id` is `course-v1:{org}+{course}+{run}` / `program-v1:...`. Checked against production data in a local DuckDB copy of the warehouse:

```
edxorg      1516 runs -- 1516 old-form,    0 new-form
mitxonline  4513 runs -- 4510 new-form,    3 old-form

exact id overlap: 3
  MITx/14_73x/1T2014
  MITx/14.73x/2013_Spring
  MITx/14.73x_1/1T2015
```

So for ~1513 of 1516 runs the `is null` filter on line 238 always passes. (I compared against `stg__mitxonline__...courses_courserun` rather than the full CTE, which also unions programs - so 3 is approximate, but the format mismatch is the point.)

This is inherited from the existing `edxorg_product_view`, but it now guards a much bigger row set - every edX.org run, not just verified-seat ones - so a run genuinely on both platforms gets listed twice.

If dedup is intended, normalizing one side with `{{ format_course_id(..., convert_to_old_format=false) }}` before joining would fix it. If the two-platform listing is intended, dropping the join and the filter would make that explicit.

KatelynGit: this should be fixed now

KatelynGit:
pre-commit.ci autofix

src/ol_dbt/models/marts/combined/marts__combined__products.sql
```diff
@@ -195,14 +212,52 @@ with mitxonline_product as (
...
             , 'Certificate of Completion'
         ) as certification_type
-    from edxorg_product
+    from edxorg_paid_product
     inner join edxorg_runs
-        on edxorg_product.courserun_readable_id = edxorg_runs.courserun_readable_id
+        on edxorg_paid_product.courserun_readable_id = edxorg_runs.courserun_readable_id
+        and edxorg_paid_product._row_num = 1
+    left join mitxonline_product_view
+        on edxorg_paid_product.courserun_readable_id = mitxonline_product_view.product_readable_id
```
quazi-h, line 220:
```suggestion
        on  {{ format_course_id('edxorg_paid_product.courserun_readable_id') }}
            =  {{ format_course_id('mitxonline_product_view.product_readable_id') }}
```
I think the normalization fix is needed here as well but this PR should be all set after that

quazi-h (approved):
LGTM!

Issue #2543, opened 2026-08-10
Embedding asset + embedding-model bake-off on a labeled Zendesk sample


blarghmatey:
### Description/Context

Embed each conversation once and store the vector on `afact_feedback_conversation` as an Iceberg `ARRAY<float>`, with `embedding_model_version` / `embedding_dim` / `embedding_input` stamped.

Scale: ~198K conversations for the MVP (~0.2–2.5 GB of vectors) — comfortably in-memory for batch clustering, no vector service needed.

### Plan/Design

**Model choice is an evaluation, not a pick.** Use MTEB only to *narrow*, then benchmark the shortlist on our own corpus:

- **Label set** — Zendesk `ticket_tags` (~2,354 distinct) + `group_name` as a free, noisy reference; optionally hand-label a few hundred for a cleaner set.
- **Metric is the task, not the MTEB average** — run UMAP+HDBSCAN on each candidate's vectors and score silhouette + agreement with the tag reference (ARI/NMI) + human coherence of the top clusters. Retrieval precision@k as a secondary.
- **Sweep** Matryoshka dims (256/512/1024) and cost/latency as tie-breakers.
- **Shortlist** (re-check MTEB standings at eval time) — Fenic-native managed: `gemini-embedding-001`, Cohere `embed-v4`, OpenAI `text-embedding-3-large`; self-hosted open: Qwen3-Embedding, BGE-M3, NV-Embed-v2. Starting default so implementation is not blocked on the bake-off: `gemini-embedding-001` or Cohere `embed-v4` at 512-dim, with `text-embedding-3-large` as the safe baseline.

**Also a scored arm:** `embedding_input = 'summary'` vs `'concatenated_turns'`. Adopt the summary as the input only where measured lift justifies its per-conversation LLM cost.

Redaction is upstream and mandatory — vectors are computed on redacted text only, independent of the fact that provider egress is acceptable. The choice stays reversible: a better model later is a rebuild of a derived table under a new version stamp, with `tfact_feedback` untouched.

Spec: [`feedback_ml_approach.md`](https://github.com/mitodl/ol-data-platform/blob/worktree-feedback-aggregation-spec/docs/design/feedback_ml_approach.md) §B

PR #2622, opened 2026-08-31
feat(feedback): add feedback embeddings asset and related embedding

rachellougee:
<!--- If it fixes an open issue, please link to the issue here. -->
<!--- Closes # --->
<!--- Fixes # --->
<!--- N/A --->
https://github.com/mitodl/ol-data-platform/issues/2543

<!--- Describe your changes in detail -->
  - Adds a `feedback_embeddings` Dagster asset that embeds each conversation once, using the LLM summary                                                                                                                             
  - Defaults to `text-embedding-3-large`.                                                                                                                                                   
  - Adds a separate embedding_llm resource (client_class="openai"), since Anthropic/Bedrock has no embeddings API.                                               
  - Writes to an intermediate `feedback_embeddings` table only; wiring the vector into afact_feedback_conversation is a separate PR, once this asset has run and populated the table

### How can this be tested?
<!---
Please describe in detail how your changes have been tested.
Include details of your testing environment, any set-up required
(e.g. data entry required for validation) and the tests you ran to
see how your change affects other areas of the code, etc.
Please also include instructions for how your reviewer can validate your changes.
--->
- Run the unit test suite: 
 ```                                                                                                                                   
  cd dg_projects/ml                                                                                                                                              
  uv run pytest ml_tests/ -q 
```                                                                                                                

- Run dagster locally
```
  cd dg_projects/ml                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
  export DAGSTER_ENV=dev                                                                                                                                                                                            
  export OPENAI_API_KEY =...                                                                                                                                
  export OPENAI_BASE_URL=https://parley.api.mit.edu/v1  
  export EMBEDDING_MODEL_VERSION=text-embedding-3-small                                                                                                                
                                                                                                                                                                                                                    
  uv run dagster dev -m ml.definitions       
```                                                                                                                                                                       
                                                                                                                                                                                                                    
  Then in the  Dagster UI:                                                                                                                                              
                                                                                                                                                                                                                    
  1. Materialize `feedback_summaries` first (skip if you already ran it while testing https://github.com/mitodl/ol-data-platform/pull/2604)                                             
  2. Materialize `feedback_embeddings` with a small sample_limit via Launchpad config:  
```                    
  ops:                                                                                                                                                                                                              
    feedback_embeddings:                                                                                                                                                                                            
      config:                                                                                                                                                                                                       
        sample_limit: 100              
```       
The log shows something like `Embedded 99 conversations (0 already embedded, 210 total upstream)`                                                                                                                                                                       
  3 Verify the output — check the materialization's logged row counts in Dagit, and/or query the 
```
select * from ol_data_lake_production."ol_warehouse_production_rlougee_intermediate".feedback_embeddings
```  
                                
  4. Re-run it a second time with no upstream changes — it should log 0 new embeddings, confirming the incremental filter works.                                                                                    
                                                                                                                                    

### Additional Context
<!--- optional - delete if empty --->
<!--- Please add any reviewer questions, details worth noting, etc. that will help in
assessing this change.  --->


<!--- Uncomment and add steps to be completed before merging this PR if necessary
### Checklist:
- [ ] e.g. Update secret values in Vault before merging
--->

### CODE REVIEW

File: dg_projects/ml/ml/lib/embed.py [hunk 1]
```diff
@@ -0,0 +1,348 @@
...
+    """Upsert one chunk directly into the real feedback_embeddings table.
+
+    Mirrors summarize.checkpoint_chunk: a crash after this call keeps everything
+    upserted so far, so the next run's ordinary filter_unembedded pass sees it as
+    already embedded, no separate recovery step needed. table_identifier is
+    "database.table", e.g. "ol_warehouse_production_intermediate.feedback_embeddings".
+    """
+    if chunk_df.height == 0:
+        return
+    table = catalog.load_table(table_identifier)
```
quazi-h: **This deadlocks the asset's first materialization.**

`catalog.load_table()` raises `NoSuchTableError` when the table is absent, and on a fresh deployment `feedback_embeddings` doesn't exist yet — `PolarsIcebergIOManager` only creates it in `handle_output`, which runs *after* the asset body returns. No dbt model creates it either.

The asset already assumes this: `feedback_embeddings.py:103` wraps the *read* of this same table in `contextlib.suppress(NoSuchTableError)`. The checkpoint write didn't get the same guard.

Reproduced against a real pyiceberg catalog — three embeddings computed and paid for, then:

```
NoSuchTableError: Table does not exist: intermediate.feedback_embeddings
embedding calls paid for before the crash: [3]
```

Because the exception escapes `embed_and_checkpoint`, the asset never returns, so the io_manager never creates the table, and every later run fails identically. It won't reproduce in a dev schema where the table already exists from earlier testing.

Minimal fix — 47/47 existing tests pass unchanged:

```python
try:
    table = catalog.load_table(table_identifier)
except NoSuchTableError:
    logger.info("%s does not exist yet; skipping checkpoint", table_identifier)
    return
```

That leaves the first run unprotected, which matters at the ~198K scale. If you'd rather keep checkpointing from run one, `catalog.create_table_if_not_exists(table_identifier, schema=chunk_df.to_arrow().schema)` also works — I verified the round-trip preserves `List(Float32)` and a second run upserts rather than duplicates — but it needs `_FakeCatalog` in the two checkpoint tests updated.

Same shape in `summarize.checkpoint_chunk` on #2604.

dg_projects/ml/ml/lib/embed.py
```diff
@@ -0,0 +1,348 @@
...
+    batch call -- retrying one at a time isolates it rather than dropping every
+    otherwise-fine row in the chunk along with it.
+
+    errors, if given, collects each failure's message -- lets a caller surface
+    *why* calls failed (e.g. in a Failure message) without changing this
+    function's return type.
+    """
+    try:
+        vectors = client.embed_batch([row["resolved_text"] for row in chunk])
+    except Exception:
```
quazi-h: The individual-retry fallback fires on *any* batch exception, not just the bad-row case the docstring describes.

Measured at the default `EMBEDDING_BATCH_SIZE=100`: a systemic failure costs **101 calls per chunk** — the batch, then all 100 rows replayed one at a time. On a 429 that's 100 extra requests aimed at an endpoint that just asked us to back off, and since the OpenAI SDK already retries with backoff (`DEFAULT_MAX_RETRIES = 2`), a surfaced `RateLimitError` means backoff was already exhausted.

Gating the fallback on exception type drops it to 1 call while keeping bad-row isolation intact — verified 4/5 rows still embedded around one poison row:

```python
except (openai.AuthenticationError, openai.PermissionDeniedError, openai.RateLimitError):
    raise
except Exception:
    ...  # existing per-row fallback
```

Trade-off worth naming: this makes a 429 fail the asset instead of being absorbed into `errors`. If you'd rather it degrade, the abort counter already covers that — happy to be argued out of it.

dg_projects/ml/ml/lib/embed.py
```diff
@@ -0,0 +1,348 @@
...
+        # afact_feedback_conversation's embedding_vector column both specify a
+        # single-precision Iceberg ARRAY<float> -- Float64 would double storage and
+        # commit a different physical type than requested.
+        pl.Series("embedding_vector", vectors, dtype=pl.List(pl.Float32)),
+        pl.lit(client.dim, dtype=pl.Int64).alias("embedding_dim"),
+        pl.lit(client.model_version, dtype=pl.String).alias("embedding_model_version"),
+    )
+
+
+def embed_conversations(
```
quazi-h: Nit: `embed_conversations` isn't used in production — the asset imports `embed_and_checkpoint`, and this is only reachable from `test_embed.py`.

It duplicates `embed_and_checkpoint`'s chunk loop without the `EMBEDDING_MAX_CONSECUTIVE_FAILED_CHUNKS` abort, so anything that switches to it silently loses the systemic-failure guard. Worth deleting, or a docstring line marking it as the deliberately un-checkpointed variant.

quazi-h (commented):
Three findings below, all as comments rather than request-changes so you can push back. The first one I'd normally raise as a blocker.

Copilot's five findings all look genuinely addressed (batching, `Float32`, `embedding_dim` in the incremental key, the `Failure` on total failure, `turn_count` for content change) - not restating any of those.

I reviewed this in a worktree at `b1f6f9f6` and reproduced findings 1 and 2 against a real sqlite-backed pyiceberg catalog rather than reasoning from the diff. Both suggested fixes below pass the existing 47 tests unchanged.

One note on the content-change thread rather than a new comment: Copilot asked for `turn_count` **or** a hash of `resolved_text`, and `turn_count` is what landed. That covers a conversation gaining a turn, but #2604 adds a `record_refs` config to `feedback_redacted` for re-redacting individual records - so `conversation_text` can change under a stable `turn_count`, leaving those embeddings silently stale. Probably a follow-up issue rather than this PR.

dg_projects/ml/ml/lib/embed.py
rachellougee: This is addressed. I noticed the same issue on the summarize branch and applied the same `create_table_if_not_exists` fix there too. There were a lot of changes in between, so it fell through the cracks. 

One note if you want to re-test: drop the `feedback_summaries` and `feedback_embeddings` tables in your schema first, as `feedback_conversation_pk` was added to both after they were created.

dg_projects/ml/ml/lib/embed.py
rachellougee: Fixed, but with `except openai.BadRequestError` instead of listing a denylist,  everything else drops the whole chunk as one error.

The denylist misses other systemic errors (APIConnectionError, InternalServerError, etc.), and raising on the three listed bypasses the abort counter, so one transient rate limit would kill the whole run.  Thank you for the suggestion though

rachellougee:
I believe all the feedback are addressed so merging this to unblock https://github.com/mitodl/ol-data-platform/pull/2648

Issue #2534, opened 2026-08-10
Feedback Aggregation & Clustering System


blarghmatey:
### Description/Context

Free-text feedback about our platforms and courses is scattered across systems that never talk to each other: Zendesk support tickets, edX discussion-forum posts, Learn AI tutor conversations, Open edX ORA peer-assessment feedback, and a forthcoming edX feedback plugin. Nobody can answer "what are the recurring problems people hit this month?" without reading tickets one at a time, and nothing distinguishes a one-off complaint from a systemic issue.

Four audiences feel this gap at three altitudes:

| Audience | Altitude | Core question |
|---|---|---|
| Customer support | operational | What recurring issues are coming in this week? |
| Engineering | operational | Which systemic platform defects does feedback point to? |
| Instructors / instructional designers | course-scoped | What are my learners saying about *my* course? |
| Department leadership | strategic | How are category and sentiment trends moving? |

**Design review:** [RFC #12210](https://github.com/mitodl/hq/discussions/12210) (Draft — supersedes [#10793](https://github.com/mitodl/hq/discussions/10793))
**Spec set:** https://github.com/mitodl/ol-data-platform/pull/2422 — [`README_feedback_aggregation.md`](https://github.com/mitodl/ol-data-platform/blob/worktree-feedback-aggregation-spec/docs/design/README_feedback_aggregation.md) is the index

### Plan/Design

Five decisions are settled and captured in the RFC and spec set:

1. **Ingress — contract-first (RFC Option 3).** Land feedback in the Learn AI Postgres DB as an *interim* path using the existing Airbyte → dbt scaffolding, but define a stable common event contract now and key the warehouse tables off stable **source** business identifiers, so the later cut over to the `ol-analytics-api`/StarRocks data bus is source-swap + backfill + parity rather than a rebuild. Building the data bus first was rejected on schedule risk — its write path does not exist and its sink topology is an unsettled platform-wide question.

2. **A fact pair, not one table.**
   - `tfact_feedback` — the *ingested* record, one row per human free-text utterance at **turn** grain (Zendesk public requester-authored comment, forum `*.created` post, tutor human turn, ORA submission, plugin event). Sparse-nullable conformed FKs are the source-flexibility mechanism. **Insert-only** — no post-insert write path exists.
[...]

Issue #2359, opened 2026-07-01
Retire legacy_openedx IRx export: consumer-side delivery options for maintained-asset facade


blarghmatey:
### Description/Context

We are retiring the legacy nightly Open edX export pipeline ([`dg_projects/legacy_openedx`](https://github.com/mitodl/ol-data-platform/tree/main/dg_projects/legacy_openedx)) that delivers data to the MIT Institutional Research (IRx) team for processing with [Simeon](https://github.com/MIT-IR/simeon). That flow extracts directly from the Open edX MySQL/MongoDB databases and is increasingly unmaintainable. All of the data it delivers is now produced by supported pipelines, so we plan to replace it with a thin **facade** (a new `dg_projects/irx_export` code location) that re-packages already-maintained assets and drops them in S3.

This issue enumerates the **consumer-side delivery options and their tradeoffs** so the IRx team can tell us which shape works best on their end before we build. Please comment with your preference and any Simeon constraints we should design around.

**What you receive today (legacy contract).** Per deployment (`mitx`, `xpro`, `mitxonline`), nightly at 00:00 UTC:

```
s3://mitx-etl-{...}-{env}/{deployment}/{YYYYMMDD}/
├── course_ids.csv
├── users_query.csv
├── studentmodule_query.csv
├── enrollment_query.csv
├── role_query.csv
├── role_users.csv
├── forum/                    (MongoDB dump, BSON)
└── courses/{course_id}.tar.gz
```

There is no manifest file; completion is signalled only via a Healthchecks.io ping. Source: [`legacy_openedx/ops/open_edx.py`](https://github.com/mitodl/ol-data-platform/blob/main/dg_projects/legacy_openedx/legacy_openedx/ops/open_edx.py).

**What the maintained pipelines already produce:**

- **Tabular extracts** → the `irx__` dbt models (Iceberg tables), which were written specifically to match the `edx-analytics-exporter` output that Simeon ingests (see [`models/external/README.md`](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_dbt/models/external/README.md)). These are a **superset** of the six legacy CSVs — they additionally include grades, certificates, ORA submissions/scores, peer assessments, cohorts, and teams.
- **Course exports** → `course_xml` tar.gz archives, one content-addressed object per course run.
- **Tracking logs** → normalized / JSON-ified event logs, daily partitioned.
[...]

PR #2626, opened 2026-08-31
Stage the forum-v2 MySQL tables for all three deployments

blarghmatey:
Part of https://github.com/mitodl/ol-data-platform/issues/2359 (retire `legacy_openedx`). Also unblocks the forum source in https://github.com/mitodl/ol-data-platform/issues/2534 (feedback aggregation, Phase 2).

Stages the forum-v2 MySQL tables so dbt can see them. They have been syncing from the edxapp MySQL every 12h since the Mongo→MySQL migration, but all 12 were `modeled: false`, so no source YAML was generated and the data has been sitting in the lake invisible to dbt.

- Adds 18 staging models — six tables × three deployments — plus their source entries, model docs, and the `modeled: true` flips in the three `ingestion/inventory/units/*__mysql.yml`.
- Six tables, not all 12: `forum_commentthread`, `forum_comment`, `forum_uservote`, `forum_mongocontent`, `forum_abuseflagger`, `forum_historicalabuseflagger`. The other six (`coursestat`, `edithistory`, `forumuser`, `lastreadtime`, `readstate`, `subscription`) have no consumer yet and stay unmodelled.
- Two consumers are waiting on this: the IRx forum export that replaces the legacy mongodump, and the feedback project, whose Phase 2 reads forum posts at turn grain. Columns are named to the warehouse convention rather than to either consumer's contract.
- No `deduplicate_raw_table`. These are `full_refresh_overwrite`, so there are no append duplicates — same as the other full-refresh openedx tables.

**`content_type_id` is carried through unresolved, deliberately.** Four of the six tables are Django generic foreign keys keyed on `(content_type_id, content_object_id)`. Resolving that id needs `django_content_type`, which is not ingested from edxapp for any deployment. It cannot be hardcoded — the value is assigned per deployment by migration order (mitx 688/695, xpro 606/613, mitxonline 570/577) — and it cannot be inferred by joining, because `forum_commentthread.id` and `forum_comment.id` both start at 1 and overlap heavily enough that an unfiltered join matches *both* tables and returns a plausible-looking wrong answer. The hazard is written into the column descriptions so the next reader does not rediscover it. Ingesting `django_content_type` is tracked separately and blocks the irx forum model.

**Bodies are carried through unredacted**, consistent with every other staging model (`auth_user` staging carries `user_email`) and with there being no redaction convention in this layer. `mit_irx` is granted on the `external` directory only, not on staging, so delivering `body` to IRx stays an explicit decision at the `irx__` layer rather than something that leaks through by default.

### How can this be tested?
Built and inspected locally on `dev_local` (DuckDB over prod Iceberg, zero-copy):

```
ol-dbt local register --database ol_warehouse_production_raw
dbt run -t dev_local --select fqn:*forum_commentthread fqn:*forum_comment \
  fqn:*forum_uservote fqn:*forum_mongocontent fqn:*forum_abuseflagger \
  fqn:*forum_historicalabuseflagger
# Done. PASS=19 WARN=0 ERROR=0 SKIP=0     (18 models + 1 project hook)
```

Row counts from that build (local views registered 2026-08-25, so slightly behind live):

| model | mitxresidential | mitxpro | mitxonline |
|---|---:|---:|---:|
| `forum_commentthread` | 7,027 | 104,892 | 48,924 |
| `forum_comment` | 10,882 | 79,656 | 73,832 |
| `forum_uservote` | 2,837 | 16,915 | 23,454 |
| `forum_mongocontent` | 17,758 | 170,911 | 103,595 |
| `forum_abuseflagger` | 6 | 21 | 170 |
| `forum_historicalabuseflagger` | 3 | 3 | 57 |

Spot checks on the built output:
- Timestamps render as ISO 8601, e.g. `2026-08-25T11:37:17.813042Z`.
- No nulls in the columns downstream work depends on: `forumthread_id`, `user_id`, `forumcomment_created_on` (0 of 73,832 each on mitxonline), `forumcontent_mongo_id` (0 of 103,595).
- `forumcontent_mongo_id` is uniformly 24 characters and fully distinct (103,595 of 103,595), consistent with a one-to-one ObjectId bridge.
- `forumuservote_value` is `1` for all 23,454 mitxonline rows — the Open edX forum UI offers no downvote, so any consumer computing a `down_count` will get 0.

Static checks: `dbt parse` clean, `dbt list` shows all 18 registered, `ol-dbt validate --model staging` reports 0 errors (its 4 warnings are pre-existing docs-coverage gaps on unrelated models).

### Additional Context
The first local build failed on all 12 timestamp columns — `to_iso8601` is Trino-only and does not exist in DuckDB. Switched to the dispatched `cast_timestamp_to_iso8601` macro, which is the portable form.

That is worth a separate note: 24 staging models still call `to_iso8601` directly, and they genuinely do not build on `dev_local` — confirmed by running `stg__mitxpro__openedx__mysql__courserun_enrollment`, which fails with `Catalog Error: Scalar Function with name to_iso8601 does not exist`. Not touched here, but it means local-dev coverage is narrower than it looks.

Reviewer questions:
- Column naming: `forumthread_*` / `forumcomment_*` with `user_id`/`courserun_readable_id` for the conformed keys. Two unrelated consumers read these, so I aimed at the warehouse convention rather than either contract — say if you would name them differently, it is cheaper to change now than after both consumers build on them.
- Not macro-ised into one templated model. Three copies matches every other openedx staging table; these should get swept up by the existing per-platform-clone task along with everything else rather than forking the pattern for six tables.

### DIFF
--- a/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
+++ b/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
@@ -2236,6 +2236,9 @@ models:
   columns:
   - name: forumthread_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: user_id
     description: int, foreign key to auth_user for the thread author
   - name: user_username
@@ -2285,6 +2288,9 @@ models:
   columns:
   - name: forumcomment_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: forumthread_id
     description: int, foreign key to forum_commentthread. This is the thread association;
       in the pre-migration Mongo data the equivalent field held an ObjectId string
@@ -2332,6 +2338,9 @@ models:
   columns:
   - name: forumuservote_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: user_id
     description: int, foreign key to auth_user for the voter
   - name: forumuservote_value
@@ -2353,10 +2362,16 @@ models:
   columns:
   - name: forummongocontent_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: forumcontent_mongo_id
     description: str, the 24-character hex ObjectId this row carried in the pre-migration
       Mongo forum database. Only content that existed at the migration has a row here,
       so content created since has no ObjectId at all
+    tests:
+    - unique
+    - not_null
   - name: forumcontent_type_id
     description: int, Django content-type id identifying which forum table content_object_id
       points at. The value is assigned per deployment by migration order, so it is
@@ -2373,6 +2388,9 @@ models:
   columns:
   - name: forumabuseflag_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: user_id
     description: int, foreign key to auth_user for whoever raised the flag
   - name: forumabuseflag_flagged_on
@@ -2393,6 +2411,9 @@ models:
   columns:
   - name: forumhistoricalabuseflag_id
     description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
   - name: user_id
     description: int, foreign key to auth_user for whoever raised the flag
   - name: forumhistoricalabuseflag_flagged_on
--- a/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
+++ b/src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
@@ -2230,3 +2230,179 @@ models:
     description: int, foreign key to users_user for the user who uploaded the image
   - name: image_description
     description: str, optional description of the image as entered in the CMS
+
+- name: stg__mitxonline__openedx__mysql__forum_commentthread
+  description: MITx Online open edX discussion forum threads
+  columns:
+  - name: forumthread_id
+    description: int, the auto-incremented ID for this table
+  - name: user_id
+    description: int, foreign key to auth_user for the thread author
+  - name: user_username
+    description: str, username of the thread author
+  - name: user_retired_username
+    description: str, placeholder username left behind when the author has been retired
+  - name: courserun_readable_id
+    description: str, Open edX Course ID in the format course-v1:{org}+{course code}+{run_tag}
+  - name: forumthread_title
+    description: str, the thread title as authored by the learner
+  - name: forumthread_body
+    description: str, free-text thread body authored by the learner
+  - name: forumthread_type
+    description: str, discussion or question
+  - name: forumthread_commentable_id
+    description: str, id of the discussion topic the thread was posted under
+  - name: forumthread_context
+    description: str, where the thread lives, e.g. course
+  - name: forumthread_group_id
+    description: int, cohort the thread is scoped to, null when visible to everyone
+  - name: forumthread_is_closed
+    description: boolean, whether the thread has been closed to further replies
+  - name: forumthread_closed_by_user_id
+    description: int, foreign key to auth_user for whoever closed the thread
+  - name: forumthread_close_reason_code
+    description: str, reason code recorded when the thread was closed
+  - name: forumthread_is_pinned
+    description: boolean, whether the thread is pinned to the top of its topic
+  - name: forumthread_is_visible
+    description: boolean, whether the thread is visible, used as a soft delete
+  - name: forumthread_is_endorsed
+    description: boolean, whether the thread has an endorsed response
+  - name: forumthread_is_anonymous
+    description: boolean, whether the thread was posted anonymously
+  - name: forumthread_is_anonymous_to_peers
+    description: boolean, whether the thread is anonymous to other learners but not
+      to staff
+  - name: forumthread_created_on
+    description: timestamp, when the thread was posted
+  - name: forumthread_updated_on
+    description: timestamp, when the thread was last edited
+  - name: forumthread_last_activity_on
+    description: timestamp, when the thread last saw any activity
+
+- name: stg__mitxonline__openedx__mysql__forum_comment
+  description: MITx Online open edX discussion forum responses and comments
+  columns:
+  - name: forumcomment_id
+    description: int, the auto-incremented ID for this table
+  - name: forumthread_id
+    description: int, foreign key to forum_commentthread. This is the thread association;
+      in the pre-migration Mongo data the equivalent field held an ObjectId string
+      rather than an integer
+  - name: forumcomment_parent_id
+    description: int, foreign key to forum_comment for the response this comment replies
+      to. Null for a top-level response to the thread
+  - name: user_id
+    description: int, foreign key to auth_user for the comment author
+  - name: user_username
+    description: str, username of the comment author
+  - name: user_retired_username
+    description: str, placeholder username left behind when the author has been retired
+  - name: courserun_readable_id
+    description: str, Open edX Course ID in the format course-v1:{org}+{course code}+{run_tag}
+  - name: forumcomment_body
+    description: str, free-text comment body authored by the learner
+  - name: forumcomment_depth
+    description: int, nesting depth, 0 for a direct response to the thread
+  - name: forumcomment_sort_key
+    description: str, ordering key used to sort sibling comments
+  - name: forumcomment_child_count
+    description: int, number of direct replies to this comment
+  - name: forumcomment_endorsement_data
+    description: str, JSON object holding the endorsing user and time. Empty object
+      when the comment is not endorsed
+  - name: forumcomment_is_endorsed
+    description: boolean, whether the comment has been endorsed by staff
+  - name: forumcomment_is_visible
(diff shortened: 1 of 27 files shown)


src/ol_dbt/models/staging/mitxresidential/_stg_mitxresidential__models.yml
```diff
@@ -545,3 +545,179 @@ models:
     description: string, the edX course ID formatted as course-v1:{org}+{course code}+{run_tag}
     tests:
     - not_null
+
+- name: stg__mitxresidential__openedx__mysql__forum_commentthread
+  description: Residential MITx open edX discussion forum threads
+  columns:
+  - name: forumthread_id
+    description: int, the auto-incremented ID for this table
```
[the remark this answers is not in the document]
blarghmatey: Added in 3110b48e5: unique + not_null on each new model's primary ID, plus on forum_mongocontent.forumcontent_mongo_id, which is the key IRx reconciles its history against and is 1:1 with the bridge row. Verified before asserting rather than after — all 42 tests pass on dev_local against production data.

src/ol_dbt/models/staging/mitxpro/_stg_mitxpro__models.yml
```diff
@@ -2367,3 +2367,178 @@ models:
     description: int, foreign key to users_user for the user who uploaded the image
   - name: image_description
     description: str, optional description of the image as entered in the CMS
+
+- name: stg__mitxpro__openedx__mysql__forum_commentthread
+  description: MITx Pro open edX discussion forum threads
+  columns:
+  - name: forumthread_id
+    description: int, the auto-incremented ID for this table
```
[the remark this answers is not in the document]
blarghmatey: Added in 3110b48e5: unique + not_null on each new model's primary ID, plus on forum_mongocontent.forumcontent_mongo_id. All 42 tests pass on dev_local against production data.

src/ol_dbt/models/staging/mitxonline/_stg_mitxonline__models.yml
```diff
@@ -2230,3 +2230,179 @@ models:
     description: int, foreign key to users_user for the user who uploaded the image
   - name: image_description
     description: str, optional description of the image as entered in the CMS
+
+- name: stg__mitxonline__openedx__mysql__forum_commentthread
+  description: MITx Online open edX discussion forum threads
+  columns:
+  - name: forumthread_id
+    description: int, the auto-incremented ID for this table
```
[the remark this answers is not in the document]
blarghmatey: Added in 3110b48e5: unique + not_null on each new model's primary ID, plus on forum_mongocontent.forumcontent_mongo_id. All 42 tests pass on dev_local against production data.

blarghmatey:
Addressed the review: 3 threads fixed and resolved. Added unique + not_null on each new model's primary ID, plus on forum_mongocontent.forumcontent_mongo_id (3110b48e5) — 42 tests, all passing on dev_local against production data. All checks green.

src/ol_dbt/models/staging/mitxresidential/_stg_mitxresidential__models.yml
```diff
@@ -545,3 +545,200 @@ models:
...
+  - name: forumuservote_id
+    description: int, the auto-incremented ID for this table
+    tests:
+    - unique
+    - not_null
+  - name: user_id
+    description: int, foreign key to auth_user for the voter
+  - name: forumuservote_value
+    description: int, +1 for an upvote and -1 for a downvote. Only upvotes occur in
+      practice, since the Open edX forum UI does not offer a downvote
```
quazi-h: **This invariant is false on mitxresidential.** Vote histograms from production:

| deployment | `+1` | `-1` |
|---|---:|---:|
| mitxonline | 23,430 | 0 |
| mitxpro | 16,915 | 0 |
| **mitxresidential** | 1,950 | **887** |

887 of 2,837 votes here are downvotes — 31%. The spot check in the PR description covered mitxonline only (all `1`), and that generalized into a platform-wide claim that now ships verbatim into all three deployments' docs. Residential runs a different Open edX version, so the "the UI does not offer a downvote" premise doesn't transfer.

Worth fixing because it's the readers you name — the IRx export and the feedback project — who would trust it. The PR description's "any consumer computing a `down_count` will get 0" needs the same correction.

Suggest either dropping the second sentence, or scoping it: `downvotes are present on mitxresidential only`.

src/ol_dbt/models/staging/mitxresidential/stg__mitxresidential__openedx__mysql__forum_uservote.sql
```diff
@@ -0,0 +1,18 @@
...
+
+with source as (
+    select * from {{ source('ol_warehouse_raw_data', 'raw__mitx__openedx__mysql__forum_uservote') }}
+)
+
+, cleaned as (
+
+    select
+        id as forumuservote_id
+        , user_id as user_id
```
quazi-h: Nit: `user_id as user_id` is a no-op alias. Nine occurrences across the three `forum_uservote` / `forum_abuseflagger` / `forum_historicalabuseflagger` copies; there's no other self-alias of that shape anywhere in `src/ol_dbt/models/staging/` on `main`.

quazi-h (commented):
Reviewed against production data rather than the diff — the raw forum tables are all registered in the local DuckDB copy, so the claims in the description were directly testable.

One substantive finding (inline, on `_stg_mitxresidential__models.yml`): the `forumuservote_value` description asserts that only upvotes occur, and mitxresidential has 887 downvotes. Details inline. There's an irony worth naming — the three-way copy is *why* this is a bug: the identical sentence is correct on two deployments and wrong on the third.

**Optional, take it or leave it:** no `relationships` test on `forum_comment.forumthread_id` → `forum_commentthread`, and no `not_null` on `user_id`. Both hold today on all three deployments (0 orphan `comment_thread_id`, 0 null `author_id`), and both waiting consumers join on exactly these columns. `relationships` is used sparingly in staging (27 occurrences layer-wide), so this is a suggestion rather than a convention gap — just cheap while the tests are already being touched.

---

**Verified and correct** — listing what I actually measured so none of it gets re-checked:

- **The no-`deduplicate_raw_table` decision holds.** This is the claim most likely to silently double row counts, so I checked it directly: 0 duplicate `id`s and exactly one `_airbyte_generation_id` across all 18 raw tables.
- **Column completeness is 1:1 against the real raw schema**, not just against the `sources.yml` this PR also authors. Every business column is mapped; only the four `_airbyte_*` columns are dropped, matching convention.
- **The `content_type_id` values are exactly as documented** — mitx 688/695, xpro 606/613, mitxonline 570/577; exactly two per deployment, no third content type.
- **The generic-FK ambiguity warning is real, not defensive boilerplate.** mitxresidential ct=688: all 6,885 `content_object_id`s resolve in *both* `forum_commentthread` and `forum_comment`. mitxpro ct=613: all 76,772 resolve in both. An unfiltered join really does return a plausible-looking wrong answer — writing the hazard into the column descriptions was the right call.
- **`forumcontent_mongo_id` `unique` + `not_null` holds on all three** (103,595 / 170,911 / 17,758 rows, zero nulls, uniform 24-char length), so the tests added in 3110b48e will pass.
- **Three-way drift is clean.** The 18 SQL models are byte-identical after normalizing deployment tokens; both YAML sets are identical modulo line-wrap position.
- **The `to_iso8601` note is exactly right.** 26 files repo-wide contain a bare call, 24 of them under `staging/` — matching your count. (Worth recording, since a naive grep inflates this: `cast_timestamp_to_iso8601` contains `to_iso8601` as a substring.)
- **The 6-of-12 scoping is consistent** across all three inventory files.

On your two reviewer questions: the column naming reads right to me — `forumthread_*` / `forumcomment_*` with `user_id` / `courserun_readable_id` for the conformed keys is the warehouse convention, and aiming at that rather than at either consumer's contract is the correct call for a staging model. Agreed on not macro-ising; three copies matches every other openedx staging table and forking the pattern for six tables would make the eventual sweep harder, not easier.

Copilot's `unique`/`not_null` finding was addressed in 3110b48e and all three threads are resolved; nothing here overlaps it.

src/ol_dbt/models/staging/mitxresidential/_stg_mitxresidential__models.yml
blarghmatey: Fixed in be453686e: dropped the upvote-only claim from mitxresidential's forumuservote_value description. Description now just states +1/-1 semantics without asserting a distribution that varies by deployment.

src/ol_dbt/models/staging/mitxresidential/stg__mitxresidential__openedx__mysql__forum_uservote.sql
blarghmatey: Fixed in be453686e: dropped the no-op `user_id as user_id` alias across all 9 occurrences (forum_uservote, forum_abuseflagger, forum_historicalabuseflagger x3 deployments).

blarghmatey:
Addressed the review: both threads fixed and resolved in be453686e — dropped the upvote-only claim from mitxresidential's forumuservote_value description, and removed the no-op `user_id as user_id` alias across all 9 occurrences. All checks green.

Issue #2544, opened 2026-08-10
Clustering, category-proposal and sentiment assets + the run-promotion loop


blarghmatey:
### Description/Context

The remaining ML stages, all writing generated columns on `afact_feedback_conversation`.

### Plan/Design

**Clustering** — UMAP (to ~5–15 dims) then HDBSCAN. HDBSCAN over k-means specifically because it needs no pre-set `k` (the number of systemic themes is unknown and grows) and because its **noise class** (`cluster_id = -1`) *is* the one-off-complaint bucket we explicitly want separated from systemic signal. At conversation grain `min_cluster_size` reads directly as "how many conversations before we call it systemic". Stays our own sklearn step — Fenic offers K-means only.

**Run promotion** — each run writes a `feedback_cluster_run` row (algorithm, params, `cluster_count`, `noise_count`, silhouette, `run_status`). An unpromoted run's assignments sit in `feedback_cluster_candidate` until approved; promotion copies them onto `afact_feedback_conversation` and drops the candidates. This is what lets a proposed run be compared against the live one.

**Category** — bootstrapped, not cold-start. Tag and group seeds give a usable taxonomy before any clustering runs; then LLM-label each cluster from centroid samples + dominant seed tags, one call **per cluster** (hundreds — trivial cost, and the critical distinction from per-record work). `category_status` stays `proposed` until a human approves. SCD-lite: relabeling changes `category_label`, never `category_slug`.

**Sentiment** — two tiers, cheapest first. Zendesk CSAT is grain-matched to the conversation, so it is an exact label for the rated ~6% rather than a value propagated across turns — which is what makes it a trustworthy validation set. For the rest, evaluate explicit+embedding-kNN vs explicit+local-classifier vs explicit+LLM on a labeled sample and pick by accuracy-vs-cost; default expectation is that embedding-kNN wins on cost. `sentiment_source` records which tier produced each row.

Human-in-the-loop is required, not optional: the LLM proposes and a human curates `dim_feedback_category`. That decoupling is what lets clustering re-run freely.

Spec: [`feedback_ml_approach.md`](https://github.com/mitodl/ol-data-platform/blob/worktree-feedback-aggregation-spec/docs/design/feedback_ml_approach.md) §§C, D, E

PR #2648, opened 2026-09-08
feat(feedback): add UMAP+HDBSCAN clustering asset for feedback conversations

rachellougee:

https://github.com/mitodl/ol-data-platform/issues/2544

  - Adds a `feedback_clustering` Dagster multi-asset that reduces `feedback_embeddings` vectors via UMAP and clusters them via HDBSCAN                                                                                                                                                          
  - Writes two new append-only intermediate tables:                                                                                                              
    - `feedback_cluster_run` — one row per run (algorithm, UMAP/HDBSCAN params, `cluster_count`, `noise_count`, `silhouette_score`, `run_status`)                
    - `feedback_cluster_candidate` — one row per conversation per run, keyed on `feedback_conversation_pk` (`cluster_run_id`, `cluster_id`, `cluster_probability`, plus `source_slug`/`conversation_ref` for debugging)    
 - Every run gets its own `cluster_run_id`, so multiple clustering runs can coexist and be compared before anything is promoted.                                                                                                
  - Does **not** touch `afact_feedback_conversation` or promote any run to live, that's a separate follow-up                                                                                                                    

### How can this be tested?

  - Run the unit test suite:             
  ```                                                                                                                        
  cd dg_projects/ml                                                                                                                                              
  uv run pytest ml_tests/test_cluster.py -q  
```                                                                                                                    
  - Run Dagster locally :   
 ```                                                                                                                
  cd dg_projects/ml                                                                                                                                              
  export DAGSTER_ENV=dev                                                                                                                                         
 uv run dagster asset materialize --select 'intermediate/feedback_cluster_run,intermediate/feedback_cluster_candidate' -m ml.definitions --config-json '{"ops": {"feedback_clustering": {"config": {"min_cluster_size": 5, "umap_n_components": 2, "umap_n_neighbors": 3, "embedding_model_version": "text-embedding-3-small", "embedding_dim": 1024}}}}' 

```                           
  - Verify the output                                          
  ```sql                                                                                                                                                         
  select * from ol_data_lake_production."ol_warehouse_production_rlougee_intermediate".feedback_cluster_run                                                     
  select cluster_id, count(*) from ol_data_lake_production."ol_warehouse_production_rlougee_intermediate".feedback_cluster_candidate group by cluster_id        
  ```
  - Re-run with a different `min_cluster_size/umap_n_neighbors` and confirm a distinct `cluster_run_id` shows up alongside the earlier run rather than overwriting it.         
  - Re-run with `embedding_input_filter: summary` vs `concatenated_turns` and compare cluster contents to confirm mixing both arms produces format-driven (not topic-driven) clusters.                                                                                                                                                  
         

### Additional Context

While testing this, I found that clustering both raw ticket text and AI-written summaries together makes HDBSCAN group by text format (raw vs. summary), not by actual topic. That's not useful, the whole point of clustering is to find real recurring issues, not to separate "summarized tickets" from "raw tickets.".

**Supporting statistics:**

Mixing arms produces a format-driven split, not a topic-driven one (mixed run `ea2e28c4`):

| Cluster | `concatenated_turns` (raw) | `summary` |
|---|---|---|
| Cluster 0 | 43 | 0 |
| Cluster 1 | 1 | 16 |

Each arm clusters well on its own, but mixed clustering scores far worse than either arm alone:

| Run | Arm(s) | Conversations | Clusters found | Noise | Silhouette |
|---|---|---|---|---|---|
| `ea2e28c4` | both (mixed) | 60 | 2 | 0 | 0.050 |
| `2ccae777` | summary only | 106 | 7 | 43 (41%) | 0.461 |
| `9199339e` | raw only | 193 | 9 | 53 (27%) | 0.457 |
                                                                                                                                                              
(0 noise on the mixed run isn't a good sign here, it means every conversation got forced into one of 2 loose clusters)  

The simple fix: stop skipping short/single-turn conversations from summarization, so every conversation gets the same consistent LLM-generated style regardless of length. But that's a change to feedback_summaries (#2542).  Based on cost estimates, it would only increase the one-time backfill by roughly $75-$100.


dg_projects/ml/ml/lib/cluster.py
```diff
@@ -0,0 +1,243 @@
...
+
+import logging
+import os
+import uuid
+from typing import Any
+
+import numpy as np
+import polars as pl
+from sklearn.cluster import HDBSCAN
+from sklearn.metrics import (
```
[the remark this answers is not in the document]
rachellougee: Verified — sklearn.cluster.HDBSCAN is native to scikit-learn since v1.3 (currently pinned >=1.3, installed 1.9.0 here), not the separate hdbscan package. Confirmed via `from sklearn.cluster import HDBSCAN` succeeding and the full test suite exercising it. No change needed.

dg_projects/ml/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "dagster-postgres>=0.27.13",
     "dagster-iceberg>=0.3.8",
     "httpx2 >= 2.2.0",
-    "scikit-learn>=1.2.2",
+    "scikit-learn>=1.3",  # sklearn.cluster.HDBSCAN was added in 1.3
```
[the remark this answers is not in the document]
rachellougee: Verified — sklearn.cluster.HDBSCAN is native to scikit-learn since v1.3 (currently pinned >=1.3, installed 1.9.0 here), not the separate hdbscan package. Confirmed via `from sklearn.cluster import HDBSCAN` succeeding and the full test suite exercising it. No change needed.

dg_projects/ml/ml/assets/feedback_clustering.py
```diff
@@ -0,0 +1,216 @@
...
+from ol_orchestrate.lib.iceberg_maintenance import get_glue_catalog
+from pydantic import Field
+from pyiceberg.exceptions import NoSuchTableError
+from pyiceberg.expressions import EqualTo
+
+if DAGSTER_ENV == "dev":
+    _schema_suffix = os.environ.get("DBT_SCHEMA_SUFFIX")
+    database_name = f"ol_warehouse_production_{_schema_suffix}_intermediate"
+else:
+    database_name = "ol_warehouse_production_intermediate"
```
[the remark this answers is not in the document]
rachellougee: No need to handle QA here, as there is no upstream QA support

dg_projects/ml/ml/assets/feedback_clustering.py
```diff
@@ -0,0 +1,218 @@
...
+from ol_orchestrate.lib.iceberg_maintenance import get_glue_catalog
+from pydantic import Field
+from pyiceberg.exceptions import NoSuchTableError
+from pyiceberg.expressions import EqualTo
+
+if DAGSTER_ENV == "dev":
+    _schema_suffix = os.environ.get("DBT_SCHEMA_SUFFIX")
+    database_name = f"ol_warehouse_production_{_schema_suffix}_intermediate"
+elif DAGSTER_ENV == "qa":
+    database_name = "ol_warehouse_qa_intermediate"
```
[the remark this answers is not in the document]
rachellougee: Fixed in b9b64061: removed the QA-specific branch so feedback_clustering now uses ol_warehouse_production_intermediate outside dev, matching feedback_embeddings and feedback_summaries.

dg_projects/ml/ml/lib/cluster.py
```diff
@@ -0,0 +1,243 @@
...
+from umap import UMAP
+
+JOIN_COLS = ["feedback_conversation_pk"]
+# Not part of the join key, but carried through for debugging without a join.
+DEBUG_COLS = ["source_slug", "conversation_ref"]
+
+CLUSTER_RUN_SCHEMA = {
+    "cluster_run_id": pl.String,
+    "algorithm": pl.String,
+    "embedding_model_version": pl.String,
```
blarghmatey, line 26: What is the "version" here? Should this just be a model name or URI?

blarghmatey: Looking at the code, I see that it's a string identifier for a model. It would be good for us to have a broader compatibility with embedding models beyond just OpenAI. This could include Gemini, OpenAI, Bedrock, as well as locally run open-weights embedding models.

blarghmatey (approved):

dg_projects/ml/ml/assets/feedback_clustering.py
```diff
@@ -0,0 +1,216 @@
...
+        embeddings_df = embeddings_df.sample(n=config.sample_limit, seed=RANDOM_STATE)
+
+    if embeddings_df.height < config.min_cluster_size:
+        msg = (
+            f"Only {embeddings_df.height} embedded conversations available "
+            f"(min_cluster_size={config.min_cluster_size}); not enough to form "
+            "even one cluster. Run feedback_embeddings first, or lower "
+            "min_cluster_size for a small local test."
+        )
+        raise Failure(msg)
```
quazi-h: **This `Failure` will be retried in a loop.**

`stop_run_retries` returns early unless the exception is a `PermanentFailure` (`failures.py:86`), so a bare `Failure` never stamps `dagster/retry_on_asset_or_op_failure` and the `run_retries` daemon re-runs it — each attempt re-scanning `feedback_embeddings` from S3 before failing the same way.

"Not enough embedded conversations to cluster" is permanent by definition; a rerun can't add rows. This is the case `ol_orchestrate.lib.failures` was written for, and its module docstring has the precedent: a correctly-classified unretryable failure that `run_retries` still re-ran **5,363 times over six days**.

```python
from ol_orchestrate.lib.failures import permanent_failure
...
raise permanent_failure(msg)
```

dg_projects/ml/ml/definitions.py
```diff
@@ -61,13 +62,20 @@
...
+    name="feedback_clustering_job",
+    selection=[feedback_clustering],
+)
+
+# Scoped to just these assets, independent of the ml code location's shared
+# default_automation_condition_sensor. Stopped by default so a fresh deploy
+# doesn't auto-run against an unverified LLM credential; enable in the UI
+# once the Bedrock/API path is confirmed working.
+feedback_pipeline_automation_sensor = AutomationConditionSensorDefinition(
+    name="feedback_pipeline_automation_sensor",
```
quazi-h: **Renaming this drops the sensor's stored on/off state.**

Pushing back on the Copilot suggestion this came from (`29776506`) — the clarity argument is fair, but the rename isn't free.

Dagster keys instigator state on `selector_id = create_snapshot_id(selector)`, and the selector carries the instigator *name* (`_core/scheduler/instigation.py:272,276`). `feedback_summaries_automation_sensor` is already on `main`, so it's a live sensor with state in the production instance. Under the new name there's no stored state, `DefaultSensorStatus.STOPPED` applies again, and the cursor is lost.

If that sensor is currently enabled in production, this deploy silently switches off summaries, embeddings *and* clustering — no error, no log line, and the comment directly above says it's meant to be turned on in the UI once the credential is confirmed.

Two ways out, both fine: keep the old name and just add `feedback_clustering` to `target=`, or keep the rename and make re-enabling it part of the deploy. Worth knowing which before this ships.

dg_projects/ml/ml/assets/feedback_clustering.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # context.run_id, not a fresh UUID: retry-stable so _clear_partial_run can
+    # repair a prior partial attempt instead of orphaning it.
+    cluster_run_id = context.run_id
```
quazi-h: **`context.run_id` isn't retry-stable, so this can never match anything.**

`create_reexecuted_run` (`dagster/_core/instance/runs/run_domain.py:446`) sets `parent_run_id = parent_run.run_id` on a **new** run — retries get a new id. Combined with the `run_retries` daemon being the retry path here (see the comment on line 168), the concrete failure is:

1. the `feedback_cluster_candidate` append succeeds
2. the `feedback_cluster_run` append fails on the Glue/S3 commit
3. the daemon retries with a different `run_id`
4. `_clear_partial_run` deletes zero rows

The first attempt's candidate rows stay in an append-only table with no matching run row — a phantom `cluster_run_id` that every `group by cluster_run_id` will pick up.

`context.run.root_run_id or context.run_id` is stable across re-executions and makes the repair actually fire.

Secondary: because this currently never matches, the `table.delete()` above it is a predicate evaluation across the data files of a growing append-only table, paid every run for nothing. `cluster_run_id` is neither partition nor sort key and run ids are random hex, so per-file bounds won't prune it. (Reasoned, not measured — unlike the rest of this review.)

quazi-h (commented):
Reviewed at `b9b64061` in a worktree — built the project and ran things rather than reading the diff, so the numbers below are measured unless I say otherwise. Baseline: 63 tests pass.

Three inline comments for the things I'd want resolved before merge. Everything else is here so it doesn't turn into seven threads.

Not restating Copilot: the `sklearn.cluster.HDBSCAN` findings are wrong and your rebuttal is right (1.9.0 is installed and the suite exercises it), QA routing is resolved, and the `embedding_input_filter` note is answered by defaulting to `summary`. One Copilot item never became a thread, though — it got suppressed into a review-details block, so it may not have been seen: `logger` at `cluster.py:73` is defined and never used.

---

**The silhouette docstring's failure model doesn't hold, and small clusters are silently dropped** — `cluster.py:119`

The docstring says an unlucky subsample "raises ValueError ... roughly 68% of the time" and that this returns `None`. I tested it: 3 clusters sized 4000/4000/15, `sample_size=200`, 300 seeds. **Zero ValueErrors**, even though ~211 of the 300 subsamples contained no member of the 15-point cluster at all. It returns a confident 0.947 every time, scored over the survivors.

At 198K rows with `SILHOUETTE_MAX_SAMPLES=5000`, that means clusters near `min_cluster_size` — the marginal-but-systemic issues this whole feature is looking for — are routinely excluded from the number used to compare runs, with nothing recording that it happened.

**The small-run guard gates on the wrong parameter** — `feedback_clustering.py:163`

The guard checks `height < min_cluster_size` and the message says to lower `min_cluster_size` for a small local test. The constraint that actually binds at small N is `umap_n_components` (UMAP's spectral init needs k < N). Following the message's own advice:

| height | n_components | min_cluster_size | result |
|---|---|---|---|
| 11 | 10 (default) | 5 | passes the guard, then `TypeError: Cannot use scipy.linalg.eigh for sparse A with k >= N` |
| 12 | 10 | 5 | ok |
| 15 | 10 | 15 (default) | ok — 0 clusters, 15 noise, `run_status="completed"` |

Worth folding `umap_n_components` into the same check.

**`run_status` can only ever be `"completed"`** — `cluster.py:241`

It's the only literal in the module, and the failure path raises before anything is written. So the column can't tell a failed run from one that never happened — and per the last row above, a run that clustered nothing is also `completed`. Either write a `failed` row or drop the column.

**`embedding_input` isn't carried onto candidate rows** — `cluster.py:219`

`cluster_embeddings`' own docstring argues for stamping provenance on the run row because `feedback_embeddings` is upserted in place and can't be reconstructed later. The per-row arm has exactly that problem. For a mixed-arm run, the per-cluster raw-vs-summary breakdown in your PR description can't be reproduced from these two tables — you'd have to join back to `feedback_embeddings`, which may have moved. One line: add it to `DEBUG_COLS`.

**Smaller things**

- `compute_cluster_agreement` (`cluster.py:146`) is only reachable from tests, and `CLUSTER_RUN_SCHEMA` has no `ari`/`nmi` column, so nothing it computes could be persisted anyway. Fine if it's staged for the promotion follow-up — just flagging that it's currently unreachable.
- `candidates_df.cast(CLUSTER_CANDIDATE_SCHEMA)` doesn't reorder columns, so the written order is `feedback_conversation_pk, source_slug, conversation_ref, cluster_run_id, ...` while the constant declares `cluster_run_id` first. Harmless for Iceberg, but the constant stops describing the table.
- `total_conversations` records only the post-filter count, so a `summary`-only run and a `concatenated_turns`-only run have no shared denominator to compare coverage against.

**Things I checked that are right**

`_clear_partial_run`'s `NoSuchTableError` guard is the first-materialization bug from #2622 handled properly — I ran it against a real sqlite-backed pyiceberg catalog and the absent-table, matching-rows and no-match paths all behave. Yielding candidates before the run row is also deliberately the right order.

dg_projects/ml/ml/lib/cluster.py
rachellougee: Yes, that's the motivation for this model configuration, but Gemini and bedrock each return a different shape and need their own adapter.

rachellougee: Added client_class='gemini'/'bedrock_embeddings' in LLMClientFactory, with GeminiEmbeddingClient/BedrockEmbeddingClient adapters in embed.py (Bedrock dispatches on model prefix since Titan/Cohere have different request shapes). openai_compatible already covers self-hosted/gateway routes like Parley. e5484f0

dg_projects/ml/ml/assets/feedback_clustering.py
rachellougee: Fixed: now raises permanent_failure(msg) instead of a bare Failure, so run_retries won't retry a condition a rerun can't fix. e5484f0

dg_projects/ml/ml/definitions.py
rachellougee: Agreed — kept feedback_summaries_automation_sensor's existing name and added feedback_clustering to its target, per your first option. e5484f0

dg_projects/ml/ml/assets/feedback_clustering.py
rachellougee: Fixed: cluster_run_id = context.run.root_run_id or context.run.run_id, per your suggested fix. e5484f0

rachellougee:
Leaving both as-is. compute_cluster_agreement is finished code for the future embedding-model bake-off (§B.1) — nothing calls it yet because that comparison hasn't run, not because it's dead weight. The total_conversations cross-arm denominator gap is real, but #2542's summarization-skip fix would likely collapse the arm split it's meant to compare, making it moot before it'd be used.

PR #2650, opened 2026-09-08
fix(dbt): resolve the raw dedup column from the inventory, not from Airbyte

### LINKED ISSUE #2443: edxorg staging models dedup on _airbyte_extracted_at, a column the dlt loader never emits
### Expected Behavior
The `deduplicate_raw_table()` macro should collapse duplicate raw rows per record key to the most recent load, for every staging model that calls it.

### Current Behavior
`deduplicate_raw_table(order_by='_airbyte_extracted_at', ...)` ([macro definition](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/macros/apply_deduplication_query.sql#L24-L39)) defaults `order_by` to `_airbyte_extracted_at` and does a `row_number() over (partition by ... order by {{ order_by }} desc)`.

The edxorg S3 tables (`raw__edxorg__s3__tables__*`) are no longer loaded by Airbyte — they're produced by the `edxorg_s3` dlt source ([`src/ol_dlt/ol_dlt/sources/edxorg_s3/__init__.py`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dlt/ol_dlt/sources/edxorg_s3/__init__.py#L160-L169), `write_disposition="merge"` with an explicit `primary_key`). dlt does not emit an `_airbyte_extracted_at` column — it emits `_dlt_load_id` / `_dlt_id`.

All 7 edxorg S3 staging models still call `deduplicate_raw_table` with the Airbyte-only column, so `order by _airbyte_extracted_at` is ordering by a column the underlying raw table doesn't have:

- [`stg__edxorg__s3__user.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__user.sql#L5)
- [`stg__edxorg__s3__user_profile.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__user_profile.sql#L5)
- [`stg__edxorg__s3__user_courseaccessrole.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__user_courseaccessrole.sql#L5)
- [`stg__edxorg__s3__courserun_enrollment.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__courserun_enrollment.sql#L5)
- [`stg__edxorg__s3__courserun_grade.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__courserun_grade.sql#L5)
- [`stg__edxorg__s3__courserun_certificate.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__courserun_certificate.sql#L5)
- [`stg__edxorg__s3__courseware_studentmodule.sql:5`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__courseware_studentmodule.sql#L5)

Separately, [`generate_base_model_enhanced.sql`](https://github.com/mitodl/ol-data-platform/blob/845a6745bd8e87bee0dbe7c8e02363bc6956ac9d/src/ol_dbt/macros/generate_base_model_enhanced.sql#L136) gates *auto-generated* base models' dedup step on `"_airbyte_extracted_at" in column_names` — so for any dlt-loaded raw table, codegen silently turns dedup off rather than erroring. The hand-written edxorg models above predate that codegen path and instead call the macro directly with a hardcoded (now-stale) `order_by`.

Net effect on the 7 models above: since Trino/Iceberg treats an unresolvable `order by` column reference as a compile error in some engines but can silently resolve to `NULL`/undefined ordering in others depending on how the query is planned, this needs to be confirmed against the live warehouse (see Possible Solution). At minimum it is dead/incorrect logic; at worst it produces nondeterministic "most recent" row selection since `dlt`'s `merge` write disposition already collapses to one row per primary key at write time, meaning post-merge duplicates (if any exist, e.g. from historical back-loads before the `merge` disposition was in place) are not being deduplicated on the intended recency signal.

### Steps to Reproduce
1. Note that `raw__edxorg__s3__tables__auth_user` and siblings are loaded via the `edxorg_s3` dlt source (merge + primary_key), not Airbyte.
2. Inspect the table's actual columns in the warehouse (Trino/Glue) and confirm whether `_airbyte_extracted_at` is present at all (e.g. leftover from a prior Airbyte-loaded generation of the table) or absent.
3. If absent: running any of the 7 staging models above should fail to compile, or resolve `_airbyte_extracted_at` to an ambiguous/undefined column depending on query engine behavior.
4. If present (leftover Iceberg column from before the dlt cutover): confirm whether it's stale/frozen (never updated by new dlt-written rows), which would make the `order by` deterministic-but-wrong (always preferring old Airbyte-era rows over newer dlt-written ones during the row_number partition).

### Possible Solution
Introduce a loader-agnostic macro, e.g. `raw_extracted_at()`, that resolves to `_airbyte_extracted_at` for Airbyte-loaded sources and `_dlt_load_id` for dlt-loaded sources (monotonic per-load; coarser than a per-row timestamp, but adequate for "latest load wins" since dlt's `merge` disposition already dedupes on primary key within a load). Point the 7 edxorg staging models above at the new macro, driven by a per-source loader flag, so future Airbyte→dlt cutovers are a config change rather than a per-model SQL edit.

This is called out as a blocking cross-cutting fix in the proposed [Airbyte → dlt migration RFC](https://github.com/mitodl/hq/discussions) (link TBD once posted) — filing separately since it's a live data-quality bug on an already-completed migration (edxorg), independent of whether/when the rest of the migration proceeds.

### Additional Details
Found during a 2026-07-10 evaluation of the broader Airbyte → dlt migration (dedup macro: `src/ol_dbt/macros/apply_deduplication_query.sql`; codegen gate: `src/ol_dbt/macros/generate_base_model_enhanced.sql:136`). `deduplicate_raw_table` is called by ~70 staging models total across the warehouse; edxorg is the only source that has actually completed an Airbyte→dlt cutover so far, making it the sole live instance of this coupling bug today — but every future per-source cutover under the proposed migration plan will hit the same issue unless the macro is fixed first.

blarghmatey:

Closes https://github.com/mitodl/ol-data-platform/issues/2443

`deduplicate_raw_table` defaulted `order_by` to `_airbyte_extracted_at`, a column only Airbyte writes. As sources move to dlt the column disappears and every staging model that dedups on it stops compiling.

- Resolves the ordering column per **raw table** through the ingestion inventory, replacing the Airbyte-only default.
- dlt- and dagster-loaded tables resolve to `none`, and the macro then emits a **pass-through** instead of a dedup.
- The macro has no default ordering column any more: a caller passes `raw_table=` or an explicit `order_by=`, and passing neither is a compile error.
- Fixes `generate_base_model_enhanced.sql`, which gated dedup on `"_airbyte_extracted_at" in column_names` and so generated dlt models with dedup silently switched **off**.
- Deletes the dead `deduplicate_query` macro (zero callers, hardcoded `_airbyte_emitted_at`).
- Makes `_dlt_load_id` the loader default for dlt units, so the dlt side has a real ordering key rather than a pass-through.

> **Prerequisites landed and validated.** #2651 turned on `add_dlt_load_id`; #2653 fixed the regression it caused (dlt declares the column non-nullable, and pyiceberg refuses to add a required column to an Iceberg table that already holds rows). QA `keycloak_ingest` now succeeds and its `raw__keycloak__app__postgres__client` carries `_dlt_load_id` as **optional**, 27 fields, 0 nulls across 72 rows.
>
> **Remaining precondition is per-unit, not global.** Each dlt unit needs one successful load on `a124ad58`+ before its staging models can order by the column. `keycloak` has had one; `edxorg/mysql` has not, which is exactly what its transitional `raw_metadata_column:` null encodes. That line is deletable in a follow-up once edxorg reloads.

This is not hypothetical. In `ol_warehouse_production_staging`, all 7 edxorg S3 staging models named in #2443 are frozen at 2026-06-24 13:10 — 76 days before this PR — while models that don't dedup on a missing column rebuilt normally.

<details>
<summary><b>Implementation details</b></summary>
<br>

**Why per raw table, and why through the inventory.** One dbt source mixes tables from units with different loaders — edxorg's raw namespace nests three — so this cannot be resolved per source. It also cannot come from dbt's `source.loader`: `_edxorg_sources.yml` declares `loader: airbyte` over 18 dlt-produced tables, and `generate.py` hardcoded that value into every generated sources file. `INGESTION_INVENTORY_SPEC.md` §3.3 rule 8 already guarantees a `raw_table` names exactly one unit, so the lookup is unambiguous even where prefixes nest.

**Dedupping dlt units on `_dlt_load_id` (second commit).** `LOADER_METADATA_COLUMNS[dlt]` resolves to `_dlt_load_id` instead of `none`. The column itself is turned on by **#2651**, which lands first so its first load can confirm dlt evolves an existing Iceberg table to add it — the one thing not testable locally. The value is `str(increasing_precise_time())`, which dlt guarantees increases over time for a given schema/destination/dataset, so "highest load id wins" is a valid latest-version ordering.

`add_dlt_id` stays off deliberately. On the pyarrow path `_dlt_id` is a random per-row value (`arrow.py` fills it with `generate_dlt_ids`); the deterministic `key_hash`/`row_hash` variants exist only in the relational normalizer, which `arrow.py` notes in an explicit `TODO`. So a row hash is not reachable from our backend, and would order nothing if it were.

**The one transitional declaration.** `edxorg/mysql` declares an explicit null `raw_metadata_column`, with the condition for removing it written beside it. Its tables predate the flag, and it is the **only** dlt unit whose tables feed a dedup caller — the same 7 models this issue was filed for. Making them order by a column no reload has written yet would reinstate exactly this bug. The other six dlt units have no dedup caller, so the new default is inert for them until they matter. Once an edxorg_s3 load runs on the new config, deleting that line restores the default; `merge` keeps pre-flag rows at a null load id, and NULLS LAST means freshly stamped rows win.

**Null ordering is now explicit.** Trino documents `NULLS LAST` as its default in both directions, but this ordering has to survive the StarRocks migration, and the case it protects is real — a unit that starts stamping the load id leaves earlier rows null, and nulls-first would let a stale row win the dedup.

**Why `none` was the original answer.** #2443 proposed falling back to `_dlt_load_id`. Today there is no such column present: dlt writes no metadata columns on these tables at all, neither `_airbyte_*` nor `_dlt_*`. Verified against production `raw__edxorg__s3__tables__auth_user` (28 columns, none beginning with `_`) and QA `raw__mitxonline__app__postgres__b2b_organizationpage` (7 fields, all business columns).

That absence is a **default we have not changed**, not a dlt limitation. In dlt 1.30.0 (`dlt/normalize/configuration.py`) the flags differ per normalizer:

```python
json_normalizer    = ItemsNormalizerConfiguration(add_dlt_id=True,  add_dlt_load_id=True)
parquet_normalizer = ItemsNormalizerConfiguration(add_dlt_id=False, add_dlt_load_id=False)
```

`ol_dlt.database` builds its resources with `backend="pyarrow"` (`src/ol_dlt/ol_dlt/database.py:244`), so every one of these sources takes the parquet path and gets both flags off. Turning them on is one config line — `dlt.config["normalize.parquet_normalizer.add_dlt_load_id"] = True`, the line dlt's own `sql_database_pipeline.py` template ships commented out.

So `_dlt_load_id` remains available as a future option, and this design accommodates it without a code change: a unit that enables the flag declares `raw_metadata_column: _dlt_load_id` and the macro picks it up. `none` is the right answer for the tables as they exist today, and it is also defensible on the merits — dlt's `merge` disposition dedups on the primary key within a load and `replace` leaves one row per key, so there is nothing to collapse. The macro's own comment scopes the dedup to Airbyte's "Incremental Sync - Append" mode.

**Why a declaration rather than a two-way switch.** Two Salesforce tables still carry the Airbyte v1 pair (`_airbyte_ab_id` / `_airbyte_emitted_at`). They declare `raw_metadata_column: _airbyte_emitted_at` in the inventory rather than being special-cased in code. That third convention is the reason resolution goes through a declaration at all.

**How the inventory reaches dbt.** dbt's Jinja cannot read YAML, so `ol-dbt inventory metadata-columns --write` renders the map into a committed macro, `_raw_metadata_columns.sql`. A generated macro rather than a var or seed is what makes every dbt invocation agree — local, CI and Dagster — with no runtime plumbing to forget. Without `--write` the command checks the file is current and exits 1 if not, and a unit test asserts the same thing; the failure to guard against is an inventory edit that lands without regeneration, after which dbt keeps deduplicating a cutover source on a column it no longer has.

**Removing the default.** The old default is how a model could acquire a dependency on Airbyte metadata without anyone writing it down, then keep it silently after its source moved. Caller counts have grown since the issue's 2026-08-13 measurement — 72 → 123 callers, 36 → 77 naming an `_airbyte_*` column — so taking over the default is what stops the seam widening further.

**Not in scope.** 5 models keep an explicit `order_by` naming `_airbyte_extracted_at` inside a compound expression (the four `*__tracking_logs__user_activity` models and `stg__edxorg__program_entitlement`, e.g. `'_airbyte_extracted_at desc, _ab_source_file_last_modified desc, "time"'`). All 5 read Airbyte-loaded units, so they are correct today, and a multi-column tie-breaker cannot reduce to the single column this map returns. They need handling at their own unit's cutover.

</details>

### How can this be tested?

```
uv run pytest src/ol_dbt_cli/tests/
uv run ol-dbt inventory validate
uv run ol-dbt inventory metadata-columns          # check mode; exits 1 when stale
cd src/ol_dbt && uv run dbt compile --select staging
```

What I ran and observed:

- **584 tests pass**, including 13 new ones covering the resolution order (table → unit → loader default), explicit-null overrides, and a staleness guard on the generated macro.
- **`dbt compile --select staging` exits 0**, zero compilation errors. That selection covers every model this PR touches; I did not compile the full 671-model project.
- **The compiled output splits the way it should**: 12 models now emit `most_recent_source as (select * from source)` with no `row_number()` — the 7 from #2443 plus edxorg `course_policy`, `courserun`, `course_video` and the two dagster-unit `course_xml_blocks` — while **97 keep `row_number() over (... order by _airbyte_extracted_at desc)` unchanged**.
- **`ol-dbt inventory validate`**: 43 units, 982 tables, 0 errors.
- **The staleness check was exercised in both directions**: flipping one unit's `loader` made it exit 1; restoring the file returned it to exit 0.
- **The `_dlt_load_id` config actually resolves**: read back through dlt's own provider as `normalize.parquet_normalizer.add_dlt_load_id = True`, with `extract.workers = 1` alongside it confirming the value came from this project's `config.toml` and not a stray one.

The 97/12 split is **identical before and after the second commit**, which is the point: enabling the load id changes no model's behaviour today. It gives the dlt units an ordering key to adopt as each one reloads.

To see the fix on the reported models specifically, compile one and confirm the pass-through:

```
cd src/ol_dbt && uv run dbt compile --select stg__edxorg__s3__user
grep -A2 "most_recent_source as" target/compiled/open_learning/models/staging/edxorg/stg__edxorg__s3__user.sql
```

### Additional Context

**On the production evidence.** The 7 edxorg models are frozen at 2026-06-24 13:10, all within 40 seconds of each other — one failed run. Controls checked the same morning (`stg__mitxonline__app__postgres__users_user`, `stg__mitxpro__app__postgres__ecommerce_company`) carry 2026-09-08 08:14–08:16 timestamps, so the warehouse itself was building normally. I did not read the production dbt run logs; the causal link is that their raw tables demonstrably lack the column the models order by, and the identical `COLUMN_NOT_FOUND` error was reproduced directly against the same macro path in QA.

**This does not unblock RFC 12711 step 8**, and reviewers of that work should know why. `mitxonline/app_postgres` declares `loader: airbyte`, which is true for production and false for QA — `MITXONLINE_APP_DLT_ENVIRONMENTS = {"dev","ci","qa"}` means QA is dlt-loaded today while production is still Airbyte. One unit, two loaders, one `loader:` field, so those tables still resolve to `_airbyte_extracted_at` and the QA pilot build still fails. That is not a defect here: the inventory is accurate about production, which is what it describes. It is that the schema cannot express a transition window, while RFC 12319 models the flip as instantaneous and step 8's QA-first pilot is a deliberate multi-week straddle. Resolving it is a schema/sequencing decision I did not want to make inside a bug fix.


rachellougee (approved):
dbt build run clean on mitxonline and mitxresidential, so I think they are good enough.

One note: add a step in `ingestion_inventory_ci.yaml` to enforce the metadata column is current, so if someone forgets to run `ol-dbt  inventory metadata-columns --write` after edit raw_metadata_column in the inventory, CI would catch it.

```
      - name: Check the generated metadata-column macro is current                                          
        run: uv run ol-dbt inventory metadata-columns                                                       
 ```

PR #2657, opened 2026-09-09
Update orders dashboard to include product name

KatelynGit:
https://github.com/mitodl/hq/issues/12281

Update orders dashboard to include product name

### How can this be tested?
review [qa](https://bi-qa.ol.mit.edu/superset/dashboard/33/) and yaml


src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
```diff
@@ -13,7 +13,7 @@ params:
   - order_id
   - product_type
   - product_readable_id
-  - order_created_on
```
quazi-h, line 16: I'm not sure if dropping this column was intentional (Not sure if this is what copilot's feedback above was referring to?). That column is still used as the sort key (line 118)

src/ol_superset/assets/datasets/Trino/orders_with_product_report_d00acae6-be70-4e6f-a399-1b3027a33563.yaml
```diff
@@ -0,0 +1,722 @@
...
+  extra: null
+  filterable: true
+  groupby: true
+  is_active: true
+  is_dttm: false
+  python_date_format: null
+  type: BIGINT
+  verbose_name: null
+currency_code_column: null
+database_uuid: 9a22a54c-8b2f-4c66-a866-3f23812ec929
```
quazi-h, line 696: Just want to confirm that this is supposed to be the QA Trino UUID? [src/ol_superset/WORKFLOWS.md](https://github.com/mitodl/ol-data-platform/blob/main/src/ol_superset/WORKFLOWS.md?plain=1#L270) defines the production Trino UUID as `8702691f-d666-4dac-943b-9382c02233e3`

quazi-h (commented):
Two small things I added inline comments for:
- Was dropping `order_created_on` intentional?
- Correct Trino Database UUID referenced?

And one more finding that might require a code change in the dbt model that this chart is now referencing: I think the swap between tables for this chart (from marts__combined__orders to orders_with_product_report) introduces a few dupes in the final chart. The newer model uses a [left join to marts__combined__orders](https://github.com/mitodl/ol-data-platform/blob/153a76e17d45e4edc745a9c1d6b92ede6f5c3296/src/ol_dbt/models/reporting/orders_with_product_report.sql#L61-L62). There is one course (course-v1:PRO+AIGE+R1) which is sold on both xPro and MITx Online and so it appears twice in the products table under the same product_readable_id and joining on that ID alone matches both, duplicating 55 orders (662,342 rows -> 662,397 rows). I believe the fix would just be to also join on platform for [that left join](https://github.com/mitodl/ol-data-platform/blob/153a76e17d45e4edc745a9c1d6b92ede6f5c3296/src/ol_dbt/models/reporting/orders_with_product_report.sql#L61-L62).

KatelynGit:
@copilot resolve the merge conflicts in this pull request

src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
KatelynGit: this is fixed

src/ol_superset/assets/datasets/Trino/orders_with_product_report_d00acae6-be70-4e6f-a399-1b3027a33563.yaml
KatelynGit: asked copilot to resolve

KatelynGit:
> Two small things I added inline comments for:
> 
> * Was dropping `order_created_on` intentional?
> * Correct Trino Database UUID referenced?
> 
> And one more finding that might require a code change in the dbt model that this chart is now referencing: I think the swap between tables for this chart (from marts__combined__orders to orders_with_product_report) introduces a few dupes in the final chart. The newer model uses a [left join to marts__combined__orders](https://github.com/mitodl/ol-data-platform/blob/153a76e17d45e4edc745a9c1d6b92ede6f5c3296/src/ol_dbt/models/reporting/orders_with_product_report.sql#L61-L62). There is one course (course-v1:PRO+AIGE+R1) which is sold on both xPro and MITx Online and so it appears twice in the products table under the same product_readable_id and joining on that ID alone matches both, duplicating 55 orders (662,342 rows -> 662,397 rows). I believe the fix would just be to also join on platform for [that left join](https://github.com/mitodl/ol-data-platform/blob/153a76e17d45e4edc745a9c1d6b92ede6f5c3296/src/ol_dbt/models/reporting/orders_with_product_report.sql#L61-L62).

@quazi-h I think all 3 items should be fixed now

src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
```diff
@@ -13,28 +13,32 @@ params:
   - order_id
   - product_type
   - product_readable_id
+  - product_name
   - order_created_on
-  - order_created_date
+  - expressionType: SQL
```
quazi-h, line 18: Just curious, what is the difference between a SQL expression vs what it was before - just passing it through as a normal column?

src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
```diff
@@ -13,28 +13,32 @@ params:
...
+    label: order_created_date
+    sqlExpression: SUBSTRING(order_created_on, 6, 2) || '/' || SUBSTRING(order_created_on,
+      9, 2) || '/' || SUBSTRING(order_created_on, 1, 4)
   - order_state
   - unit_price
   - discount
-  - order_total_price_paid
   - order_tax_amount
   - order_total_price_paid_plus_tax
+  - order_total_price_paid
```
quazi-h, line 27: Also wanted to make sure that the column position swapping here was intentional since I think it does change the order of the columns on the visible report.

quazi-h (approved):
Changes look good and the findings I brought up were addressed. I left two more comments but they are not blockers, was just curious.

src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
KatelynGit: I think this should be ok the positions are close enough that it's not too noticeable

src/ol_superset/assets/charts/Order_detail_fb906f1f-c07f-472f-9387-ad02f73cae55.yaml
KatelynGit: It was a custom SQL statement in superset but the result is the same i think we should be ok

PR #2661, opened 2026-09-10
fix(reporting): join orders_with_product_report to products by platform 

rachellougee:

na
https://pipelines.odl.mit.edu/runs/343c38e0-d642-41f4-a905-3f50ecf6eec7
```
865 of 1068 FAIL 59 dbt_expectations_expect_compound_columns_to_be_unique_orders_with_product_report_order_id__line_id__platform  [[31mFAIL 59[0m in 5.06s]

[31mFailure in test dbt_expectations_expect_compound_columns_to_be_unique_orders_with_product_report_order_id__line_id__platform (models/reporting/_reporting__models.yml)[0m

  Got 59 results, configured to fail if >10
```

Fix the `orders_with_product_report` production dbt test faiure                                                                                                                     
 - The join to `marts__combined__products` in `orders_with_product_report`.sql matched on `product_readable_id` only in https://github.com/mitodl/ol-data-platform/pull/2577. That column is not unique across platforms e.g. `course-v1:PRO+AIGE+R1` exists for both xPro and MITx Online — so a matching order line fanned out into two report rows.                                  
- Fix: add `platform` to the join condition, so each order line matches only the product row from its own platform.    

### How can this be tested?

dbt build --target dev_production --select orders_with_product_report

Compound-uniqueness test dropped from 59 failing rows to 2. The remaining 2 are a separate, pre-existing xPro data issue unrelated to this join, already before this fix, and still under the fail threshold

### DIFF
--- a/src/ol_dbt/models/reporting/orders_with_product_report.sql
+++ b/src/ol_dbt/models/reporting/orders_with_product_report.sql
@@ -60,3 +60,4 @@ select
 from orders
 left join products
     on orders.product_readable_id = products.product_readable_id
+    and orders.platform = products.platform


KatelynGit (approved):
ran the build and looks good to me, looks like a simple change

PR #2669, opened 2026-09-11
feat(b2b_analytics): add learner-grain MVs for the learner-records API

blarghmatey:
N/A. Upstream of the learner-records tenant specced in https://github.com/mitodl/ol-analytics-api/pull/55 (`docs/b2b-learner-records-design.md` §1, gaps 1, 3, 4, 5). Depends on https://github.com/mitodl/ol-infrastructure/pull/5835.

Every `mv_b2b_*` view is pre-aggregated, so nothing can back a per-learner API. This adds two StarRocks MVs in a new `b2b_learner_records` database (`models/b2b_learner_records/`). They're kept out of `b2b_analytics` so these PII-bearing views and the k-anonymized aggregate views can be granted separately. It also adds a dbt snapshot of MITx Online organization memberships.

- `mv_b2b_learner_enrollment`, at org × contract × course run × learner. It uses the funnel view's joins without the `group by`, plus `dim_user`, grade and certificate. `is_passing` and certificate state are emitted separately, and the API derives `completion_status` from both.
- `mv_b2b_learner`, at org × learner. It unions the membership roster (`bridge_user_organization`) with enrollments under the org's contracts, so assigned-but-unstarted seats appear. It emits `membership_source` (`roster` / `enrollment` / `both`). `courses_enrolled` and the enrolled-on dates count active enrollments only, which matches the API default `include_inactive=false`. `courses_passed`, `courses_certified` and `program_certificates_earned` count every enrollment, so reclaiming a seat doesn't take back a completion a partner was already sent.
- `snapshot_mitxonline_b2b_userorganization` (Trino, `dimensional` schema). `b2b_userorganization` has no timestamps and its raw table is replaced on every load, so a removed membership or a `keep_until_seen` / `is_manager` change otherwise leaves no record. `check` strategy on the mitxonline ids and flags, `hard_deletes: invalidate`. It reads the staging model so the history doesn't depend on the raw loader's metadata columns across the Airbyte to dlt move. History starts at its first run, and a membership added and removed between two runs is never recorded.

Choices a reviewer should check:
- The organization is attributed only through the contract that owns the course run, never through `organization_administration_report`'s free-text `organization_key` fallback.
- `learner_id` is `user_global_id`. Rows with a null one are filtered in the SQL rather than asserted by a dbt test: a failing test fails `starrocks_dbt_assets`, which skips the refresh of every StarRocks MV, and the tests read the previous refresh's rows. The grain tests run at `severity: warn` for the same reason. `user_pk` is emitted only as an internal join key, because it can still re-key.
- `record_updated_on` is the `updated_since` cursor. It covers enrollment, grade and certificate changes, plus program certificates in `mv_b2b_learner`. In `mv_b2b_learner` it's the max over every enrollment and program certificate, inactive and revoked included, so a deactivation or revocation moves it forward (mitxonline does both through `save()`). Every `greatest()` argument is coalesced to `''` because StarRocks' `greatest()` returns null if any argument is null. Roster changes, `dim_user` edits and course-run edits carry no change timestamp upstream, so they don't move it.
- `mv_b2b_learner` reads the dimensional sources directly instead of selecting from `mv_b2b_learner_enrollment`. [`materialized_view_relations`](https://github.com/mitodl/ol-data-platform/blob/main/dg_projects/lakehouse/lakehouse/lib/starrocks_dbt.py#L97-L110) refreshes MVs in sorted name order, so the rollup would refresh first and serve the previous refresh's data.
- `program_certificates_earned` counts a program certificate only if the program contains a course the learner enrolled in under the org's contracts. Otherwise a certificate earned outside the licence would be reported to the partner.
- Neither view filters to the org's current roster, and `keep_until_seen` roster rows are included. mitxonline creates a membership on every contract-run enrollment and code redemption, so an off-roster enrollment is mostly a learner removed from the org who kept the enrollment. `keep_until_seen` marks memberships mitxonline created itself, which SSO reconciliation never removes. It isn't a pending removal.

There are no consent or activity columns yet. The API returns `outcomes_shared: false` and null activity fields until those land upstream.

No Dagster change is needed. The refresh asset and the drift check both take schema-qualified relations from the manifest, and the drift query already filters on a list of schemas. The snapshot is built by `full_dbt_project`, which runs `dbt build` over everything not tagged `starrocks`.

### How can this be tested?
- `ol-dbt validate`: 0 errors. Pre-commit (sqlfluff, yamlfmt, yamllint) passes. `dbt parse` places the snapshot in the `dimensional` schema.
- I rendered both models against production dimensional Iceberg tables in DuckDB (`ol-dbt local register --database ol_warehouse_production_dimensional`, 2026-09-14). Neither model has been built on StarRocks yet. DuckDB's `greatest()` ignores nulls, so the null-propagation fix is checked only by reading; today 0 of 37,450 contract enrollments have a null `enrollment_created_on` or `enrollment_updated_on`. Results:
  - `mv_b2b_learner_enrollment`: 37,440 rows, all unique on (contract_pk, courserun_pk, user_pk). 0 null `user_global_id`, 0 null `record_updated_on`.
  - `mv_b2b_learner`: 17,347 rows, unique on (organization_key, user_pk), 0 null `user_global_id`. Membership: 10,661 both, 281 enrollment only, 6,405 roster only. `record_updated_on` is null for 6,349 learners, all roster-only with no enrollment under the org's contracts.
  - Compared with this PR's pre-review SQL on the same data: the enrollment view is identical row for row, including `record_updated_on`. `mv_b2b_learner` loses no rows and gains 236, all off-roster learners whose enrollments are all inactive. `courses_enrolled` is unchanged for every existing row. Completion counts only increased (41 learners on `courses_passed`, 40 on `courses_certified`, 18 on `program_certificates_earned`). 27 cursors moved forward and none moved backward.
  - The funnel reconciliation from the first revision (1,184 / 1,184 (contract, run) rows against `mv_b2b_enrollment_completion_funnel`) was not re-run.
- The snapshot has only been parsed. It hasn't run on Trino.
- After #5835 deploys and this merges, the next `starrocks_dbt_assets` run creates both MVs. Confirm with `DESC b2b_learner_records.mv_b2b_learner_enrollment` and `DESC b2b_learner_records.mv_b2b_learner`. The next `full_dbt_project` run creates `ol_warehouse_production_dimensional.snapshot_mitxonline_b2b_userorganization`.

### Additional Context
Findings the API tenant has to handle:
- 13,649 enrollment rows have a null `sso_organization_id` (first-revision count): their org has no Keycloak UUID, so an API filtering on it can't reach them. The existing MVs have the same gap.
- 101 enrollment rows carry a non-revoked certificate with `is_passing` false (first-revision count). `completion_status` has to treat an unrevoked certificate as `certified` without requiring `is_passing`.
- 4,929 mitxonline roster rows have `keep_until_seen` set, 3,458 of them in orgs with no Keycloak UUID, where SSO reconciliation never runs and the flag never clears.

### Checklist:
- [ ] Deploy https://github.com/mitodl/ol-infrastructure/pull/5835 (creates the `b2b_learner_records` database and its `app` grants) before merging

https://claude.ai/code/session_01U2joM6rRfCzWanQzAks1Yf


src/ol_dbt/models/b2b_learner_records/mv_b2b_learner.sql
```diff
@@ -0,0 +1,142 @@
+{{ config(
+    materialized='materialized_view',
```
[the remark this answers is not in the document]
blarghmatey: Fixed in f62117df8: both learner MVs now build into their own StarRocks database, b2b_learner_records (models/b2b_learner_records, +schema: b2b_learner_records), provisioned with the same app grants in https://github.com/mitodl/ol-infrastructure/pull/5835. That PR has to deploy before this one merges.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner_enrollment.sql
```diff
@@ -0,0 +1,63 @@
+{{ config(
+    materialized='materialized_view',
```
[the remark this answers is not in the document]
blarghmatey: Fixed in f62117df8, same change as the mv_b2b_learner thread: this MV is in b2b_learner_records now, and the database comes from https://github.com/mitodl/ol-infrastructure/pull/5835.

blarghmatey:
Addressed Copilot's review. The two schema threads are fixed and resolved (f62117df8, plus https://github.com/mitodl/ol-infrastructure/pull/5835). Copilot's four suppressed comments:

- `learner_id` nullable (both views): both now carry a `not_null` test on `user_global_id`, so a null fails the build instead of reaching a partner. Production has 0 nulls today.
- `record_updated_on` incomplete (both views): `mv_b2b_learner` now also includes program-certificate update times. For the rest, the documented scope is narrowed: the cursor covers enrollment, grade and certificate changes. Roster, `dim_user` and course-run edits carry no change timestamp upstream, so there is nothing to propagate yet. A partner picks those up from a full reload of the bulk export. Real change tracking, including a removal/tombstone for learners dropped from a roster, is filed as a follow-up (a dbt snapshot over the projected fields).

quazi-h (commented):
Five threads, each self-contained and independently resolvable. Two are
mechanical fixes (T1, T3), one is a correctness bug needing a small design
call (T2), and two are questions where a reply resolves the thread rather
than a code change (T4, T5).

Not re-raising the schema-placement point — f62117df8 settles it.

Things I checked that hold up: the YAML column lists match both SELECTs
exactly and in order (25 and 16), so the Dagster drift check stays valid;
`sorted()` refresh order really does put `mv_b2b_learner` before
`mv_b2b_learner_enrollment`, which justifies the rollup-not-select comment;
both models correctly filter `is_current = true` against SCD2 `dim_course_run`;
`tfact_certificate` program rows have a null `courserun_fk` and are deduped on
(user_fk, courserun_fk, certificate_scope), so the course-cert left joins can't
fan out; `bridge_organization_courserun` is unique per run, so contract
attribution can't fan out either; and the nightly job selects the whole
multi-asset, so the new group is still built.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner_enrollment.sql
```diff
@@ -0,0 +1,63 @@
...
+    e.enrollment_status,
+    g.is_passing,
+    g.grade_value,
+    g.letter_grade,
+    cert.certificate_issued_on,
+    cert.certificate_is_revoked,
+    -- Every mitxonline timestamp here is an ISO-8601 string from the same macro, so
+    -- greatest() compares them correctly. Grade and certificate floor to the
+    -- enrollment's own timestamp when absent.
+    greatest(
```
quazi-h: **Type:** bug · **Fix:** mechanical · **Sites:** 2 (+1 stale comment)

The comment above says grade and certificate "floor to the enrollment's own
timestamp when absent." That's true on DuckDB and false on StarRocks, and this
model only ever runs on StarRocks.

StarRocks' `greatest()` is NULL-propagating — any NULL argument makes the whole
expression NULL. DuckDB's ignores NULLs. So the "0 null `record_updated_on`"
validation in the PR description was run on the one engine that cannot observe
this.

The floor is `e.enrollment_created_on`, which is itself nullable:
`tfact_enrollment`'s incremental clause has an explicit `enrollment_created_on
is null` branch, and `_fact_tables.yml` puts no `not_null` on it. For a row with
a null `enrollment_created_on` and no grade or certificate, arguments 2 and 3
are both NULL, so StarRocks returns NULL for the whole expression. That row's
`record_updated_on` is null permanently, which makes it invisible to every
`?updated_since=` query the partner issues.

This repo already hit the Trino version of this and documented the fix at
`dim_user.sql:697-699` — coalesce to a `''` sentinel rather than to another
nullable column, "these are ISO8601 strings, so '' sorts below any real date."
Same shape applies here.

Sites:
- [ ] `mv_b2b_learner_enrollment.sql:42-46` (this hunk)
- [ ] `mv_b2b_learner.sql:25-29` — same three-argument construct in `active_enrollments`
- [ ] `mv_b2b_learner.sql:133-135` — the comment asserting `er.record_updated_on`
      is non-null whenever `pc` is; it rests on the construct above and needs
      rewording once this is fixed

Done when: neither `greatest()` can return NULL for any input, and the check
runs somewhere other than DuckDB (see the test-strategy thread on
`_b2b_learner_records__models.yml:66` — it changes when a test here is
meaningful at all).

Heads-up: `mv_b2b_learner.sql:136` has a separate, unrelated defect in this same
column. Fixing this one does not on its own make `record_updated_on` a sound
cursor.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner.sql
```diff
@@ -0,0 +1,149 @@
...
+    er.first_enrolled_on,
+    er.last_enrolled_on,
+    coalesce(er.courses_enrolled, 0)                                                    as courses_enrolled,
+    coalesce(er.courses_passed, 0)                                                      as courses_passed,
+    coalesce(er.courses_certified, 0)                                                   as courses_certified,
+    coalesce(pc.program_certificates_earned, 0)                                         as program_certificates_earned,
+    -- A program certificate row exists only alongside an active enrollment, so
+    -- er.record_updated_on is non-null whenever pc is; the coalesce only keeps
+    -- greatest() from going null when there is no program certificate.
+    greatest(
```
quazi-h: **Type:** bug · **Fix:** small design call, then mechanical · **Sites:** 2

`record_updated_on` is the partner's polling cursor, but it's a `max()` over a
*filtered* set. Any row that leaves the filter can't be represented — the max
can only move backwards or vanish.

Instance 1 — enrollment deactivation (`:42`, `enrollment_is_active = true`):
learner L in org O has active enrollments A (`record_updated_on` 2026-01-05) and
B (2026-06-01), so the MV shows 2026-06-01. On 2026-07-01 B is deactivated. B
leaves `active_enrollments`, `courses_enrolled` drops 2 → 1, and
`record_updated_on` moves *backwards* to 2026-01-05. A partner polling
`?updated_since=2026-06-01` never sees the change. If B was L's only enrollment,
`er` disappears entirely: `record_updated_on` goes null and all four counters go
to 0, invisible to every `updated_since` query.

Instance 2 — program-certificate revocation (`:79`,
`certificate_is_revoked = false`): revoking a program certificate drops the row
from `program_certificates`, so `program_certificates_earned` decrements *and*
`program_certificate_updated_on` stops contributing. Nothing advances the cursor,
and the partner keeps reporting a revoked certificate until a full reload.

The two certificate paths also disagree with each other. The course-cert join at
`:39-40` has no revoked filter, so a *course* certificate revocation does move
the cursor (it flows through `cert.certificate_updated_on` into
`active_enrollments.record_updated_on`). Same event class, opposite visibility.

The general fix is to derive the cursor from something that survives a row
leaving the filter — e.g. roll the timestamp up over the unfiltered join while
keeping the counters filtered, so deactivations and revocations still bump it.

Sites:
- [ ] `mv_b2b_learner.sql:25-29` and `:56` — the rollup feeding the cursor
- [ ] `mv_b2b_learner.sql:65-80` — `program_certificates`, same problem via the
      revoked filter
- [ ] `mv_b2b_learner_enrollment.sql` — worth confirming whether a row simply
      disappearing there has the same effect for `/enrollments` consumers

Done when: deactivating an enrollment or revoking a certificate makes that
learner's `record_updated_on` move *forward*.

Heads-up: `:42` is also the subject of a separate thread about whether the
counters should be active-only at all. If you change that filter, reconcile both
threads in one pass rather than editing the CTE twice.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
```diff
@@ -0,0 +1,178 @@
...
+      dim_user surrogate. Internal join key only -- it can still re-key, so the
+      API must not expose it. Join target for a future user_fk-keyed activity
+      fact.
+  - name: user_global_id
+    description: >
+      Keycloak user id (dim_user.user_global_id) -- the API's learner_id, stable
+      across email changes. The API contract requires it, so a null fails the
+      build rather than reaching a partner as an unserializable record.
+    tests:
+    - not_null
```
quazi-h: **Type:** bug (blast radius) + design · **Fix:** one decision covering three items · **Sites:** 3

Precondition governing every test in this file: a failing dbt test fails
`dbt build` → `invocation.stream()` raises `DagsterDbtCliRuntimeError` →
`looks_retriable()` (`lakehouse/lib/starrocks_dbt.py:78`, checked at
`dbt_starrocks.py:188`) returns False for a test failure, so it re-raises
immediately → `starrocks_dbt_assets` fails → `refresh_starrocks_analytics_mvs`,
which is `deps=[starrocks_dbt_assets]` (`starrocks_mv_refresh.py:17`), is
skipped. One bad row freezes all ten MVs, including the eight pre-existing
`b2b_analytics` aggregates that have nothing to do with this PR. No other model
in the StarRocks project has a test today, so this is a new failure mode for
that pipeline.

**1. Blast radius.** The `not_null` on `user_global_id` will take the aggregates
down with it. The PR already reports 13,649 rows whose org has no Keycloak UUID;
a *user* without one isn't a stretch. `severity: warn`, or scoping it so it
can't block the refresh asset, matches the stated intent without the collateral.

**2. It can't do what the description claims.** "A null fails the build rather
than reaching a partner" — but these are `refresh_method='manual'` MVs.
dbt-starrocks 1.12's `starrocks__get_create_materialized_view_as_sql` emits a
bare `refresh manual` with no initial population, and
`refresh_starrocks_analytics_mvs` exists precisely because nothing else fills
them. At test time the MV holds the *previous* cycle's rows — or nothing at all
right after a `--full-refresh`. The run's own data loads after the tests and is
never validated. Either the assertion belongs in the API/serializer layer, or it
has to run after the refresh.

**3. Missing grain tests.** Both models document a grain (org × contract ×
course_run × learner; org × learner) and the API depends on it, but the only
evidence is a one-off DuckDB count in the PR body.
`dbt_utils.unique_combination_of_columns` on `(contract_pk, courserun_pk,
user_pk)` and `(sso_organization_id, user_pk)` would make upstream grain drift
fail loudly instead of duplicating learner rows to a partner — and both
`tfact_enrollment` and `tfact_certificate` carry defensive dedups commented "no
upstream uniqueness guarantee," so drift is the expected case, not an exotic one.

These are one thread because (3) inherits (1): adding a uniqueness test at
default severity re-creates exactly the blast radius (1) is about. Whatever
severity policy you land on needs to cover both.

Sites:
- [ ] `_b2b_learner_records__models.yml:65-66` — `mv_b2b_learner_enrollment.user_global_id`
- [ ] `_b2b_learner_records__models.yml:137-138` — `mv_b2b_learner.user_global_id`
- [ ] `_b2b_learner_records__models.yml` — add the two grain tests at the chosen severity

Done when: no test in this file can skip the MV refresh, and both grains are
asserted somewhere that runs against populated data.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner.sql
```diff
@@ -0,0 +1,149 @@
...
+    join {{ source('dimensional', 'dim_course_run') }} cr
+        on boc.courserun_fk = cr.courserun_pk
+    join {{ source('dimensional', 'tfact_enrollment') }} e
+        on boc.courserun_fk = e.courserun_fk
+    left join {{ source('dimensional', 'tfact_grade') }} g
+        on e.user_fk = g.user_fk and e.courserun_fk = g.courserun_fk
+    left join {{ source('dimensional', 'tfact_certificate') }} cert
+        on e.user_fk = cert.user_fk and e.courserun_fk = cert.courserun_fk
+    where cr.is_current = true
+      and e.enrollment_is_active = true
```
quazi-h: **Type:** question, not a defect — a reply resolves this · **Fix:** none until you answer · **Sites:** 1

`enrollment_is_active = true` gates the whole `active_enrollments` CTE, so it
governs `courses_passed` and `courses_certified` (`:53-55`) just as much as
`courses_enrolled`.

So a learner who passed a course and earned a certificate, and whose enrollment
is later deactivated — unenrolled, refunded, seat reassigned to someone else —
loses that run from all three counters. `certificates_earned` on the API
(documented as `courses_certified + program_certificates_earned`) drops for an
achievement that actually happened and was already reported to the partner in an
earlier poll.

`courses_enrolled` being active-only reads as deliberate and matches the header
comment about `include_inactive=false`. It's `courses_passed` /
`courses_certified` I'd want confirmed — a completion count that goes down when
a seat is reclaimed is surprising, and partners usually read those as
cumulative.

If it is intended, a line in the two column descriptions saying completions are
scoped to currently-active enrollments would stop the next reader from filing
this again.

Heads-up: `:42` is also implicated in the `record_updated_on` thread at `:136`,
for a different reason. If you do change this filter, reconcile both there
rather than editing the CTE twice.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner_enrollment.sql
```diff
@@ -0,0 +1,63 @@
...
+        coalesce(cert.certificate_updated_on, e.enrollment_created_on)
+    )                                                                                   as record_updated_on
+from {{ source('dimensional', 'bridge_organization_courserun') }} boc
+join {{ source('dimensional', 'dim_contract') }} c
+    on boc.contract_fk = c.contract_pk
+join {{ source('dimensional', 'dim_organization') }} org
+    on c.organization_fk = org.organization_pk
+join {{ source('dimensional', 'dim_course_run') }} cr
+    on boc.courserun_fk = cr.courserun_pk
+join {{ source('dimensional', 'tfact_enrollment') }} e
```
quazi-h, line 58: **Type:** question, not a defect — a reply resolves this · **Fix:** none until you answer · **Sites:** 2

Different axis from the schema-placement point — f62117df8 settles that one.
This is about *which rows* these views emit, given that every row carries
`email` and `full_name`.

**1. No roster guard here (`:54-55`).** `tfact_enrollment` joins on course run
alone, so every enrollment in a contract's run is returned whether or not the
learner is on that organization's roster. The PR's own numbers show the set is
non-empty: 45 learners come back as `membership_source = 'enrollment'`.
Provisioning lag is the benign explanation and the header comment at
`mv_b2b_learner.sql:83-85` says so — but staff, test, and QA accounts enrolled
into a contract run land in the same bucket, and they'd be disclosed to the
partner by name and email. Worth an explicit decision rather than inheriting
one, particularly since the aggregate views this was patterned on are
k-anonymized and these are not.

**2. `keep_until_seen` roster rows (`mv_b2b_learner.sql:100`).**
`bridge_user_organization` is read unfiltered, so rows with
`userorganization_keep_until_seen = true` are included. Per
`_bridge_tables.yml:172-174` that flag means the user "is retained in the
organization until their SSO data is next observed" — a grace-period membership
pending re-confirmation, which may already be gone upstream in Keycloak. Those
learners are emitted as `membership_source = 'roster'`, counted as assigned
seats, with name and email, and nothing in the output distinguishes them from a
confirmed member. Either exclude them or surface the flag so the API can decide.

Sites:
- [ ] `mv_b2b_learner_enrollment.sql:54-55`
- [ ] `mv_b2b_learner.sql:100`

Done when: either both row sets are narrowed, or the model headers record that
emitting them is intended and who signed off.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner_enrollment.sql
blarghmatey: Fixed in 6a219c217: every greatest() argument in both views is coalesced to '', and record_updated_on is nulled only when all of its timestamps are missing. The comment at mv_b2b_learner.sql:133 is gone with the construct it described. Latent in today's data: 0 of 37,450 contract enrollments have a null enrollment_created_on or enrollment_updated_on.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner.sql
blarghmatey: Fixed in 6a219c217: record_updated_on is now the max over every contract enrollment and program certificate, inactive and revoked included, while the counters keep their filters. mitxonline deactivates through CourseRunEnrollment.deactivate_and_save and revokes through save(), so both move updated_on forward. mv_b2b_learner_enrollment keeps inactive and revoked rows, so it has no equivalent gap. Against production dimensional tables in DuckDB: 236 off-roster learners whose enrollments are all inactive now keep their row (17,111 to 17,347 rows), 27 cursors moved forward and none moved backward.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
blarghmatey: Fixed in 6a219c217: the not_null tests are replaced by filtering null user_global_id in both SELECTs, and both grains get unique_combination_of_columns at severity: warn (mv_b2b_learner keyed on organization_key, since sso_organization_id is null for orgs without a Keycloak UUID). One correction on (2): refresh manual without DEFERRED defaults to IMMEDIATE, so a newly created MV does queue an async refresh, but a plain build leaves an existing MV untouched and the tests still read the previous refresh.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner.sql
blarghmatey: Not intended. Changed in 6a219c217: courses_passed, courses_certified and program_certificates_earned count every enrollment under the org's contracts, and courses_enrolled stays active-only. Today 46 inactive contract enrollments are passing and 45 hold an unrevoked certificate.

src/ol_dbt/models/b2b_learner_records/mv_b2b_learner_enrollment.sql
blarghmatey: Answered with 6afeaa8bf, which adds snapshot_mitxonline_b2b_userorganization. b2b_userorganization has no timestamps and its raw table is replaced on every load, so until now a removed membership or a keep_until_seen change left no record. Both row sets stay in the views, and the model headers say why.

1. Off-roster enrollments: mitxonline creates a membership on every contract-run enrollment and code redemption (courses/api.py:266, b2b/views/v0/__init__.py:360), so staff and QA accounts land on the roster too and a roster guard would not exclude them. What it would hide is learners removed from the org who kept their enrollment; the snapshot records when that removal happened.
2. keep_until_seen is not a grace period before removal. The enrollment-code path is disjoint from the SSO path: mitxonline sets the flag on memberships it creates itself so SSO reconciliation does not delete them (b2b/api.py:1729 only removes keep_until_seen=False rows), and clears it once the org appears in the user's SSO data. 4,929 mitxonline roster rows carry it, 3,458 of them in orgs with no Keycloak UUID, where reconciliation never runs.

blarghmatey:
Two points in my comment above are superseded by 6a219c217 and 6afeaa8bf. The `not_null` tests are gone: null `user_global_id` is filtered in the SQL instead, because a failing test fails `starrocks_dbt_assets` and skips the refresh of every MV. The membership snapshot is in this PR (`snapshot_mitxonline_b2b_userorganization`), not a follow-up.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
```diff
@@ -0,0 +1,210 @@
...
+      enrolled before mitxonline created memberships on enrollment), or both
+      (on the roster with an active enrollment).
+  - name: is_organization_manager
+    description: Administers the organization in MITx Online (userorganization_is_manager).
+      False for enrollment-only learners.
+  - name: first_enrolled_on
+    description: Earliest active enrollment under the organization's contracts, ISO-8601.
+      Null when the learner has no active enrollment.
+  - name: last_enrolled_on
+    description: Latest active enrollment under the organization's contracts, ISO-8601.
```
quazi-h, line 183: Nit, non-blocking.

`first_enrolled_on` picked up a null clause in 6a219c217 and `last_enrolled_on` didn't, but they come from the same expression:

```sql
min(case when enrollment_is_active then enrollment_created_on end) as first_enrolled_on,
max(case when enrollment_is_active then enrollment_created_on end) as last_enrolled_on,
```

So both go null under exactly the same condition — a learner with no active enrollment. That's now a populated case rather than a theoretical one, since this PR deliberately brings the 236 off-roster all-inactive learners back into the view. Worth carrying the same sentence across.

The asymmetry predates this PR: the original had `Null for roster-only learners` on `first_enrolled_on` and nothing on `last_enrolled_on`. The completions change correctly reworded the first and just didn't touch the second.

quazi-h (commented):
Re-reviewed 6a219c217 and 6afeaa8bf against the five threads — all verified in the code, not just the commit messages. T1, T2, T3 and T4 are fixed, and the T5 pushback holds: I read keep_until_seen backwards, and your numbers (3,458 flagged rows in orgs where reconciliation never runs) are hard to explain under my reading. One correction I'll take: you're right that REFRESH MANUAL defaults to IMMEDIATE absent DEFERRED, so a new MV does queue a refresh at create time — my finding 6 had the mechanism wrong, though the conclusion you kept (an existing MV isn't refreshed by a plain build, so tests read stale rows) stands.

One documentation nit below, non-blocking.

quazi-h (approved):
PR looks good, my feedback has been addressed. I added another small nit about documentation for a specific field but that predates this PR anyways so not a blocker for these changes. Separately, I created a witan task to rewrite the userorganization_keep_until_seen column description in _bridge_tables.yml to accurately reflect the flag's definition (marker protecting mitxonline-created memberships from SSO reconciliation — instead of a grace period before removal) since that misled me during the initial review.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
blarghmatey: Fixed in 51595c70b: last_enrolled_on now carries the same null note as first_enrolled_on.

PR #2678, opened 2026-09-14
feat(feedback): implement continuous cluster assignment

### LINKED ISSUE #2544: Clustering, category-proposal and sentiment assets + the run-promotion loop

rachellougee:
<!--- If it fixes an open issue, please link to the issue here. -->
<!--- Closes # --->
<!--- Fixes # --->
<!--- N/A --->
Close https://github.com/mitodl/ol-data-platform/issues/2544

<!--- Describe your changes in detail -->
Implements automated cluster assignment.
- New `feedback_cluster_identity` asset to stabilize cluster identity across re-clustering runs, so LLM-generated category labels persist
- New `feedback_cluster_assignment` asset to automatically apply LLM-suggested categories to conversations, without waiting on human approval.   
- New dbt intermediate models `int__feedback__cluster_membership`, `int__feedback__category_proposal` each handling its Dagster-generated source table gracefully before it exists. 

### How can this be tested?

Verified end-to-end against real data in a personal dev schema.

```
cd dg_projects/ml
export DAGSTER_ENV=dev
export DBT_SCHEMA_SUFFIX=<your_suffix>
```

- Materialize `intermediate/feedback_cluster_identity` > `intermediate/feedback_cluster_assignment` >`intermediate/feedback_category_proposals` in dagster UI, or via CLI

```
  dagster asset materialize --select intermediate/feedback_cluster_identity -m ml.definitions                                 
  dagster asset materialize --select intermediate/feedback_cluster_assignment -m ml.definitions                               
  dagster asset materialize --select intermediate/feedback_category_proposals -m ml.definitions   
```

- Run dbt:

```
dbt build --target dev_production --vars 'schema_suffix: <your_suffix>'
   --select int__feedback__cluster_membership int__feedback__category_proposal dim_feedback_category afact_feedback_conversation
```

Verify the data
```
  -- Confirm conversations have cluster assignments and category resolution                                                
  select                                                                                                                      
      cluster_key                                                                                                             
      , cluster_assignment_method                                                                                             
      , cluster_similarity                                                                                                    
      , category_fk                                                                                                           
  from ol_warehouse_production_<your_suffix>_dimensional.afact_feedback_conversation                                                
  where cluster_key is not null                                                                                                                                                                                                               
                                                                                                                              
  -- Confirm LLM-proposed categories landed                                                     
  select                                                                                                                      
      cluster_key                                                                                                             
      , category_slug                                                                                                         
      , category_label                                                                                                        
      , category_source                                                                                                       
      , category_status                                                                                                       
  from ol_warehouse_production_<your_suffix>_dimensional.dim_feedback_category                                                      
  where category_source = 'llm_discovered';                                                                                   
```

### Additional Context
<!--- optional - delete if empty --->
<!--- Please add any reviewer questions, details worth noting, etc. that will help in
assessing this change.  --->


<!--- Uncomment and add steps to be completed before merging this PR if necessary
### Checklist:
- [ ] e.g. Update secret values in Vault before merging
--->


dg_projects/ml/ml/assets/feedback_category_proposals.py
```diff
@@ -0,0 +1,223 @@
...
+    io_manager_key="io_manager",
+    pool="feedback_category_proposals",
+    metadata={
+        "schema": intermediate_database_name,
+        # cluster_key alone, not (cluster_run_id, cluster_id): a cluster_key is
+        # stable across runs, so re-proposing for it (label drift, retry) replaces
+        # its one row rather than accumulating one per run that happened to
+        # re-examine it.
+        "write_mode": "upsert",
+        "upsert_options": {"join_cols": ["cluster_key"]},
```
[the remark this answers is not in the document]
rachellougee: Fixed in 5df25da2: added schema_update_mode='update' to the AssetOut metadata.

rachellougee:
The fixes are in commit 44da4d36 for copilot review findings.

I also did a full end-to-end continuous pipeline working with real LLM-generated categories, no approval step involved:                                                                                                                     
  - Real Zendesk tickets (12643, 257874, 110572, 90124, …) now have `category_fk` resolved to an LLM-discovered category        
  - Each ticket's `cluster_key` and `cluster_run_id` are populated, tracing back to the actual clustering run                     
  - Both categories (course-setup, micromasters-program) are visible and correctly labeled, with category_status = 'proposed' live on the fact table `afact_feedback_conversation` immediately

dg_projects/ml/ml/assets/feedback_cluster_assignment.py
```diff
@@ -0,0 +1,307 @@
...
+                "assigned_at": pl.Datetime(time_zone="UTC"),
+            }
+        )
+    )
+
+    # Push the join and the "needs (re-)placement" predicate into the lazy plan
+    # so only conversations that are new or newly re-embedded are ever collected
+    # -- an incremental run shouldn't materialize the whole corpus just to find
+    # the handful of rows it actually needs to place.
+    to_place_df = (
```
blarghmatey, line 263: Conversations assigned to a key that later gets retired (or merged away) are never re-placed unless they were in that run's candidate set. This filter only picks up rows with no membership or with `assigned_at < embedded_at`, and `_rewrite_from_run` only rewrites the run's candidates.

The default path hits this for a whole class of conversations, not an edge case. `feedback_clusters` defaults to `embedding_input_filter="summary"`, so the skip-rule conversations (single-turn/short, `embedding_input='concatenated_turns'`) are never candidates. They do get placed here between runs, since this path doesn't filter on arm. When the next re-cluster retires their key, they keep it permanently, and `afact_feedback_conversation` keeps showing the retired key and its label (neither the membership join nor `cluster_category` filters on `cluster_status`). The same applies to anything embedded while a clustering run was in flight, and to rows left out by `sample_limit`.

One option: after a rewrite, also re-place every membership row whose `cluster_key` is no longer active.

dg_projects/ml/ml/assets/feedback_cluster_assignment.py
```diff
@@ -0,0 +1,307 @@
...
+                "radius": pl.Float64,
+                "embedding_model_version": pl.String,
+                "embedding_dim": pl.Int64,
+            }
+        )
+    )
+    if clusters_df.height == 0:
+        return pl.DataFrame(schema=MEMBERSHIP_SCHEMA)
+
+    embedding_model_version = clusters_df["embedding_model_version"][0]
```
blarghmatey: After an embedding model or dimension change, `clusters_df` holds active clusters from both configs, and this takes model/dim from whichever row comes first.

How it gets there: `_active_cluster_members` in `feedback_cluster_identity.py` scopes to the run's model/dim, so a run on the new config sees no active members, resolves everything as `new`, and never retires the old-config keys. Both sets stay `active`.

Then here:
- if `[0]` is a new-config row, `nearest_active_cluster` compares new vectors against old centroids too, and a dimension mismatch raises `ValueError: matmul: ... mismatch in its core dimension` (checked locally)
- if `[0]` is an old-config row, the embeddings filter matches nothing once re-embedding finishes, so new conversations are silently never placed

`_rewrite_from_run` has the same problem at line 103: `embedding_dim` comes from an unfiltered `feedback_cluster` row (retired included), and `list.to_array(old_dim)` on new-dim vectors raises (also checked locally). Scoping both to the run's `embedding_model_version`/`embedding_dim` from `feedback_cluster_run`, and retiring other-config active clusters when identity processes a run, would cover it.

dg_projects/ml/ml/assets/feedback_category_proposals.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    catalog = get_glue_catalog()
+    cluster_run_id = latest_identity_processed_run(catalog, intermediate_database_name)
+    if cluster_run_id is None:
+        context.log.info(
+            "No identity-matched clustering run found; nothing to propose."
+        )
+        return pl.DataFrame(schema=dict(CATEGORY_PROPOSAL_SCHEMA))
+
+    new_cluster_keys = (
```
blarghmatey: This only considers `new`/`split`/`merged` keys from the latest identity-matched run, so the comment at line 134 about retrying keys from an earlier run that failed isn't what happens.

Example: `propose_categories` swallows a failed LLM call for key K minted in run R1. Run R2 continues K, so K is `continued` in R2's lineage and never shows up here again. The same thing happens if R2 is identity-processed before this asset runs for R1. K never gets an `llm_discovered` category, and `category_fk` stays on the tag-seed fallback.

Selecting active `cluster_key`s from `feedback_cluster` that have no row in `feedback_category_proposal` would pick these up regardless of which run minted them.

dg_projects/ml/ml/assets/feedback_cluster_identity.py
```diff
@@ -0,0 +1,425 @@
...
+    if continuity < CONTINUITY_FLOOR:
+        # Fail before writing either output -- feedback_cluster_assignment only
+        # rewrites from a run that has lineage rows, so leaving this run
+        # unresolved keeps it on the prior, still-good cluster_key set.
+        msg = (
+            f"Run {cluster_run_id} continuity {continuity:.2f} is below the "
+            f"floor {CONTINUITY_FLOOR:.2f}; a configuration change is needed, "
+            "not a rerun."
+        )
+        raise Failure(msg)
```
blarghmatey: The message says a configuration change is needed, not a rerun, but a bare `Failure` still gets rerun. `stop_run_retries` (`ol_orchestrate/lib/failures.py`) only tags the run for a `PermanentFailure`, so the daemon retries this up to `run_retries.max_retries: 3` times, and each retry recomputes the same continuity and fails again. `feedback_clusters.py` uses `permanent_failure(msg)` for this reason; same fix here.

dg_projects/ml/ml/assets/feedback_category_proposals.py
```diff
@@ -0,0 +1,246 @@
...
+
+    category_df = (
+        get_dbt_model_as_dataframe(
+            database_name=dimensional_database_name,
+            table_name="dim_feedback_category",
+        )
+        .select(["feedback_category_pk", "category_label"])
+        .collect()
+    )
+    afact_df = (
```
blarghmatey, line 182: `afact_df` and `conversation_df` load every conversation in the corpus (all ~198K `conversation_text` values) to sample at most 10 per cluster for the handful of keys that need a label. Filtering both lazily on `feedback_conversation_pk.is_in(membership_df["feedback_conversation_pk"])` before `.collect()` keeps it to the members of those clusters.

src/ol_dbt/models/intermediate/feedback/int__feedback__category_proposal.sql
```diff
@@ -0,0 +1,33 @@
...
+-- existing row (feedback_category_proposals only writes for new/split/merged
+-- keys); dim_feedback_category unions this in, always category_status='proposed'
+-- until a human corrects it.
+{% set proposal = dev_schema_source('feedback_intermediate', 'feedback_category_proposal') %}
+{# cluster_key replaces the old cluster_run_id/cluster_id compound key -- a table
+   written before this rework won't have it until the Dagster asset's own
+   schema-evolution step next runs. #}
+{% set proposal_columns = adapter.get_columns_in_relation(proposal.resolved_relation) | map(attribute='name') | list
+    if proposal.resolved_relation else [] %}
+{% set cluster_key_exists = 'cluster_key' in proposal_columns %}
```
blarghmatey: `feedback_category_proposal` doesn't exist on main, so `cluster_key_exists` (and the new `category_description_exists` on line 13) only protect dev schemas written by earlier commits on this branch. In prod both branches are dead code. If the null branch ever did fire, the `not_null`/`unique` tests on `cluster_key` would fail anyway. Dropping the stale dev tables and selecting the columns directly would be simpler, and each future column wouldn't need its own guard.

blarghmatey (commented):
Reviewed against the rev. 6 flow in `docs/design/feedback_dagster_asset_spec.md`. Two of these are correctness issues in how assignment and identity pick their scope: they trust "latest run" or "row `[0]`" instead of scoping by the run's embedding config and input arm. Details inline. Copilot's earlier findings addressed in 44da4d36b are not repeated here.

dg_projects/ml/ml/assets/feedback_cluster_identity.py
rachellougee: Fixed: uses permanent_failure instead of a bare Failure, matching feedback_clusters.py's pattern, so a bad-continuity run isn't retried up to max_retries.

dg_projects/ml/ml/assets/feedback_cluster_assignment.py
rachellougee: Fixed: both _rewrite_from_run and _incrementally_place_new_embeddings now take an explicit embedding_model_version/embedding_dim (from feedback_cluster_run for a just-processed run, or from the single distinct config among active clusters otherwise -- None if zero or more than one, in which case placement is skipped that tick). feedback_cluster_identity now also retires any active cluster from a different config than the run being matched, so a model/dim change doesn't leave two configs simultaneously active.

dg_projects/ml/ml/assets/feedback_cluster_assignment.py
rachellougee: Fixed as part of the same change: the incremental-placement predicate now also picks up any membership row whose cluster_key isn't in the current active set (retired, merged away, or from a prior config), not just rows with no membership or a stale embedded_at.

dg_projects/ml/ml/assets/feedback_category_proposals.py
rachellougee: Fixed: now selects every currently-active cluster_key with no row in feedback_category_proposal, instead of scoping to the latest run's new/split/merged lineage -- covers a key minted by an earlier run whose LLM call failed, or one continued past a later run before this asset got to it.

dg_projects/ml/ml/assets/feedback_category_proposals.py
rachellougee: Fixed: afact_df and conversation_df now filter on membership_df's feedback_conversation_pk before collect(), same pattern as the rewrite in the earlier review round.

src/ol_dbt/models/intermediate/feedback/int__feedback__category_proposal.sql
rachellougee: Fixed: dropped both column-existence guards and select the columns directly, since feedback_category_proposal doesn't exist on main at all -- there's no pre-rework schema to protect against.

rachellougee:
Addressed all 6 findings in https://github.com/mitodl/ol-data-platform/pull/2678/changes/03219effa549421938aecc2e3826701d864c51d0

blarghmatey (approved):

PR #2681, opened 2026-09-15
Fix orders with product reporting table

KatelynGit:
comments here https://github.com/mitodl/ol-data-platform/pull/2657

Fix orders with product reporting table add platform to join clause

### How can this be tested?
 dbt build --select orders_with_product_report


KatelynGit:
@copilot reverse the last commit

PR #2684, opened 2026-09-15
feat(b2b_learner_records): add a contract x course-run MV for /courses

blarghmatey:
N/A. Backs `GET /organizations/{organization_id}/courses` in the learner-records spec (https://github.com/mitodl/ol-analytics-api/pull/55). The endpoint is https://github.com/mitodl/ol-analytics-api/pull/60. Stacked on https://github.com/mitodl/ol-data-platform/pull/2669, so the base is `feat/b2b-learner-grain-mvs`.

Adds `mv_b2b_contract_courserun` to `models/b2b_learner_records/`, grain org × contract × course run. It joins `bridge_organization_courserun`, `dim_contract`, `dim_organization` and `dim_course_run` (`platform = 'mitxonline'`, `is_current`), and carries the contract's active flag, start and end dates and `seat_limit` next to the run identifiers. `seat_limit` is `b2b_contract_max_learners` with 0 folded into null: MITx Online documents both 0 and blank as unlimited, and the API contract reserves null for uncapped.

`mv_b2b_learner_enrollment` can't back `/courses`. It inner-joins `tfact_enrollment`, so a run nobody has enrolled in drops out, and it has no contract status, dates or seat limit. The new view holds no personal data. It lives in `b2b_learner_records` so the tenant reads one database.

No Dagster change: the refresh asset and the drift check find MVs by the `starrocks` tag in the manifest, which the directory config already sets.

### How can this be tested?
- `ol-dbt validate`: 0 errors. The new model's only warnings are the unresolvable `source()` column lists, which `mv_b2b_learner` gets too. Pre-commit (sqlfluff, yamllint) passes.
- I ran the model's joins and filters against production dimensional Iceberg tables in DuckDB, re-registered 2026-09-15. It hasn't been built on StarRocks.
  - 2,770 rows, unique on (`contract_pk`, `courserun_pk`), across 73 contracts.
  - 1,577 of those runs have no row in `tfact_enrollment`, so a view derived from `mv_b2b_learner_enrollment` would miss them.
  - `b2b_contract_start_date` is DATE and `courserun_start_on` an ISO-8601 string, matching the spec's `date` and `date-time`. I didn't check the end-date columns' types separately.
- After #2669 and this merge, the next `starrocks_dbt_assets` run creates it. Confirm with `DESC b2b_learner_records.mv_b2b_contract_courserun`.

### Additional Context
What the API will return from this data today:
- `contract_start_date` is null everywhere. `dim_contract.b2b_contract_start_date` is null for all 74 production contracts, and `contract_start` is null in all 74 rows of the raw `b2b_contractpage` table, so the dbt models aren't dropping it. `b2b_contract_end_date` is null for 73 of the 74.
- `seat_limit` is null (uncapped) for 27 of 74 contracts.
- 2,034 of the 2,770 rows have a null `sso_organization_id`. It's the same gap #2669 reports: the API filters on it, so those organizations' runs aren't reachable.

### Checklist:
- [ ] Merge https://github.com/mitodl/ol-data-platform/pull/2669 first, then retarget this to `main`
- [ ] Confirm the MV exists before deploying https://github.com/mitodl/ol-analytics-api/pull/60

https://claude.ai/code/session_01EDJRuj6mDpGRqNbQSsXZ9o


src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
```diff
@@ -208,3 +208,58 @@ models:
...
+    description: Display name of the contract -- the API's contract_name.
+  - name: b2b_contract_is_active
+    description: Whether the contract is currently active -- the API's contract_is_active.
+  - name: b2b_contract_start_date
+    description: Contract start date. Null when the contract has none.
+  - name: b2b_contract_end_date
+    description: Contract end date. Null when the contract has none.
+  - name: seat_limit
+    description: b2b_contract_max_learners. Null means uncapped, not zero.
+  - name: courserun_pk
```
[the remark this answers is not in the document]
blarghmatey: Fixed in d92beadfd: seat_limit is now nullif(b2b_contract_max_learners, 0), documented as null = uncapped. MITx Online's max_learners help text says 'Set to zero or leave blank for unlimited', and the API contract reserves null for that.

src/ol_dbt/dbt_project.yml
```diff
@@ -182,6 +182,26 @@ models:
...
           enabled: false
+    # Same engine config as b2b_analytics above (see the comments there); only
+    # the schema differs. These views identify individual learners, so they get
+    # their own StarRocks database, provisioned in ol-infrastructure
+    # substructure/starrocks, and can be granted apart from the aggregate views.
+    b2b_learner_records:
+      +enabled: "{{ target.type == 'starrocks' }}"
+      +materialized: table
+      +tags: ["starrocks"]
+      +schema: b2b_learner_records
```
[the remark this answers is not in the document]
blarghmatey: Verified — readonly and app do get catalog-wide GRANT SELECT ON ALL MATERIALIZED VIEWS IN ALL DATABASES (ol-infrastructure substructure/starrocks/__main__.py:437,441), so the concern is real. But the b2b_learner_records grants block and the PII-bearing mv_b2b_learner/mv_b2b_learner_enrollment models were added by #2669, already merged to main — this PR shows it only because the branch has not been rebased since. mv_b2b_contract_courserun, the only model #2684 adds, carries no personal data. Resolving here as out of scope for this PR.

src/ol_dbt/models/b2b_learner_records/mv_b2b_contract_courserun.sql
```diff
@@ -0,0 +1,37 @@
...
+    org.organization_key,
+    org.sso_organization_id,
+    org.organization_name,
+    c.contract_pk,
+    c.contract_id,
+    c.b2b_contract_name,
+    c.b2b_contract_is_active,
+    c.b2b_contract_start_date,
+    c.b2b_contract_end_date,
+    c.b2b_contract_max_learners                                                         as seat_limit,
```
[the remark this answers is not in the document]
blarghmatey: Fixed in d92beadf: seat_limit now folds a zero b2b_contract_max_learners into null.

src/ol_dbt/models/b2b_learner_records/mv_b2b_contract_courserun.sql
```diff
@@ -0,0 +1,39 @@
...
+    c.b2b_contract_is_active,
+    c.b2b_contract_start_date,
+    c.b2b_contract_end_date,
+    -- MITx Online documents max_learners 0 and null both as unlimited. The API
+    -- contract reserves null for uncapped, so 0 is folded into it.
+    nullif(c.b2b_contract_max_learners, 0)                                              as seat_limit,
+    cr.courserun_pk,
+    cr.courserun_readable_id,
+    cr.courserun_title,
+    cr.courserun_start_on,
```
quazi-h, line 29: `courserun_start_on` and `courserun_end_on` are not valid `date-time` values.
`CourseRun` in the spec declares both `format: date-time` with the example
`2026-02-01T00:00:00Z` (ol-analytics-api#55). OpenAPI's `date-time` is RFC 3339,
which *requires* a UTC offset. What this MV emits has none.

Measured by running this model's exact SELECT against the production dimensional
Iceberg tables: all 2,770 non-null `courserun_start_on` values are shaped
`2026-02-09T15:00:00.000`. 0 end in `Z`; 0 carry a numeric offset.

The cause is upstream, and it is an adapter split inside one macro.
`dim_course_run.courserun_start_on` comes from `cast_timestamp_to_iso8601`
(`stg__mitxonline__app__postgres__courses_courserun.sql:29-30`). Its DuckDB and
StarRocks branches append a literal `Z`
(`macros/cast_timestamp_to_iso8601.sql:19,21,36,37`); its Trino branch does not —
`to_iso8601(try_cast({{ column_name }} AS timestamp))` at `:9`, where the
`try_cast` downcasts the zone away. The dimensional tables are built by Trino, so
the offset-less form is what reaches StarRocks. This is the exact hazard
`trino__format_timestamp_as_iso8601` was written to avoid; its own comment says
the generic `try_cast(... AS timestamp)` "drop[s] the zone"
(`macros/format_timestamp_as_iso8601.sql:6-11`).

Good news: the values *are* UTC, so this is a labelling bug, not a data bug.
`raw__mitxonline__app__postgres__courses_courserun.start_date` is `TIMESTAMP WITH
TIME ZONE`, and for all 4,996 mitxonline runs the staged string equals the raw
instant expressed in UTC — raw `2013-02-12 10:00 America/New_York` (EST) stages
as `2013-02-12T15:00:00.000`, and 4,996 of 4,996 match to the minute.

Failure scenario: `/courses` returns `"courserun_start_on":
"2026-02-09T15:00:00.000"`. A partner validating the response against the
published schema rejects the payload. A partner not validating parses it as a
local wall clock — in `America/New_York` that becomes 20:00Z, a 5-hour shift,
moving any run starting before 05:00Z onto the wrong calendar day. 706 of the
2,770 rows carry exactly that `2026-02-09T15:00:00.000` value, so it is not a
tail case.

I'm not asking for the macro fix: `cast_timestamp_to_iso8601` has 447 call sites
across 197 models, so changing the Trino branch reformats every ISO timestamp
column in the warehouse. The narrow options are to stamp it in the MV —
`concat(cr.courserun_start_on, 'Z')`, which is null-safe in StarRocks, so the
2,752 null `courserun_end_on` rows stay null — or to let ol-analytics-api#60
normalize and record that, in which case nothing changes here but the yml wording
at `_b2b_learner_records__models.yml:266,268` ("ISO-8601" is true, but it
under-specifies the thing that matters). Your call which layer owns it, and a
reply naming the layer is a fine outcome if the fix lands in the API.

Two things worth folding into whichever way you go. The merged
`mv_b2b_learner_enrollment` surfaces these same two `dim_course_run` columns at
`:27-28`, so they have the same shape and one decision covers both models.

And if the fix lands in dbt, it is cheaper now than later: appending `Z` changes
no *column name*, and `drifted_relations` is set equality over the yml's
`columns:` (`lib/starrocks_dbt.py:196-213`). So a later same-columns edit to this
SELECT drifts nothing, `_stale_materialized_views()` adds no `--full-refresh`,
and dbt-starrocks' `get_materialized_view_configuration_changes` is an empty
macro — the build logs "no configuration changes were identified" and `REFRESH
MATERIALIZED VIEW` re-runs the *old stored query*. Green build, unchanged data.
That costs nothing today, because the MV does not exist yet and the first build
creates it with whatever SELECT is merged. After it exists, the same one-line fix
needs a hand-run `--full-refresh`.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
```diff
@@ -209,3 +209,60 @@ models:
...
+  - name: contract_pk
+    description: Surrogate key of the B2B contract.
+  - name: contract_id
+    description: >
+      mitxonline's ContractPage.page_ptr_id (dim_contract.contract_id) -- the
+      contract identifier the API exposes.
+  - name: b2b_contract_name
+    description: Display name of the contract -- the API's contract_name.
+  - name: b2b_contract_is_active
+    description: Whether the contract is currently active -- the API's contract_is_active.
```
quazi-h: "Whether the contract is **currently** active" overstates what the flag is. Both
upstream definitions say the opposite in the same breath: `_dim_contract.yml:31`
and `_stg_mitxonline__models.yml:2136` each read "Date rules still apply." It is
MITx Online's manually-set `active` marker, and it does not consult
`b2b_contract_end`.

That matters more here than it would elsewhere, because this yml is what
ol-analytics-api is written against — `documented_columns`' docstring is explicit
that "a column nobody documented is a column no consumer projects"
(`lib/starrocks_dbt.py:137-139`) — and because `contract_is_active` is nearly the
only licence-window signal `/courses` can carry today, with
`contract_start_date` null on all 74 contracts and `contract_end_date` null on
73 of them.

Dated failure scenario, from the one contract that has an end date: contract
`1594` ("MIT Horizon Digital Program | Cohort 1") is `b2b_contract_is_active =
true` with `b2b_contract_end_date = 2026-09-30`. On 2026-10-01, unless someone
manually clears the flag in MITx Online, `/courses` reports
`contract_is_active: true` for an expired contract. A partner filtering on that
field — the natural choice, since it is the only non-null one — keeps serving its
learners a licence that ended.

The row does carry `contract_end_date`, so a consumer *can* apply the date
itself; it just has to be told to. Something like "Whether the contract is marked
active in MITx Online. A manual flag — the end date is not applied, so combine it
with `b2b_contract_end_date`."

quazi-h (commented):
Reviewed at `87cd2730`. Two findings, both about what the columns *mean* to the
API rather than about the SQL — the joins and the grain measure clean.

Heads up on one process thing: Copilot's review is against `91f48bba`, which the
rebase orphaned, so nothing has bot-reviewed the current head.

What I verified by running this model's exact SELECT against the production
dimensional Iceberg tables in DuckDB:

- **Grain holds.** 2,770 rows, 2,770 distinct `(contract_pk, courserun_pk)`, and
  2,770 distinct `(contract_id, courserun_readable_id)` — so the spec's "ordered
  by `(contract_id, courserun_id)`, which is unique" is true of the API's own key
  and not just the surrogate version. 2,770 distinct `courserun_pk`, i.e. no run
  appears under two contracts, which is the `b2b_contract_id`-is-a-single-FK
  claim measured rather than asserted.
- **No join drops anything.** bridge 2,770 → after `dim_contract` 2,770 → after
  `dim_organization` 2,770 → after `platform = 'mitxonline'` 2,770 → after
  `is_current` 2,770. All 74 contracts have a non-null `organization_fk`, so the
  inner join to `dim_organization` costs nothing today, and `boc.organization_fk`
  and `c.organization_fk` agree on all 2,770 rows.
- **`is_current = true` is load-bearing, not decorative.** 89 of the runs in this
  view have an `is_current = false` twin in `dim_course_run` (2,378 historical
  rows overall), so dropping that predicate would fan those 89 out.
- **Types line up with the spec** — including the end-date pair you flagged as
  unchecked: `b2b_contract_end_date` is `DATE` (spec `date`), `contract_id`
  `BIGINT` (spec `integer/int64`), `seat_limit` `BIGINT`. `courserun_end_on` is a
  VARCHAR of the same shape as `courserun_start_on`, which is the first finding.
- **No nulls in the spec's non-nullable fields**: 0 null across
  `organization_name`, `organization_key`, `contract_id`, `b2b_contract_name`,
  `b2b_contract_is_active`, `courserun_readable_id`, `courserun_title`. All 736
  non-null `sso_organization_id` values are UUID-shaped.
- **"Null for self-paced runs" checks out.** I went at this one expecting it to
  fail, since `courserun_end_on` is null on 2,752 of 2,770 rows and 99%
  self-paced seemed unlikely. It isn't: `is_self_paced` is true for 2,703 of the
  2,770 runs, and 2,689 of the 2,752 nulls are self-paced runs.
- **"No Dagster change" checks out.** `_materialized_view_nodes` selects on
  `materialized == 'materialized_view'` plus the `starrocks` tag, both set by the
  directory config, and the schema comes from the manifest node
  (`lib/starrocks_dbt.py:97,122`), so `b2b_learner_records.mv_b2b_contract_courserun`
  is picked up with no Python edit. The yml documents all 15 columns in SELECT
  order, which the drift check's set-equality needs, and `validate`'s
  `yaml_sql_sync` still errors in both directions here because the model has no
  `SELECT *` to defeat sqlglot.
- `ol-dbt validate`: 0 errors; the new model's only warnings are the four
  unresolvable-`source()` column lists, the same class both merged siblings get.

Not raising, since you disclosed both and I measured the same numbers: the null
`contract_start_date` / `contract_end_date` (0 of 2,770 start, 1 of 2,770 end)
and the 2,034 rows with no `sso_organization_id` (736 reachable rows, 25
contracts, 21 orgs).

The first finding would normally be request-changes. Posting as a comment so you
can decide which layer owns it, or push back.

src/ol_dbt/models/b2b_learner_records/mv_b2b_contract_courserun.sql
blarghmatey: The API layer owns this. ol-analytics-api#56 adds `UtcDatetime` (`models.py:27-35`), which attaches UTC to timestamps with no zone. #60 uses it for `CourseRun.courserun_start_on/end_on`, and its test checks that `2026-02-01T00:00:00.000` comes out as `2026-02-01T00:00:00Z`. The enrollment model's `courserun_start_on/end_on` use the same type (`models.py:218-221`), so that covers mv_b2b_learner_enrollment too. In 2985481 the yml for both models now says the value is UTC with no offset. The SQL is unchanged.

src/ol_dbt/models/b2b_learner_records/_b2b_learner_records__models.yml
blarghmatey: Fixed in 2985481: the description now calls it a manual MITx Online flag that ignores the end date, and tells consumers to combine it with b2b_contract_end_date.

quazi-h (approved):

PR #2691, opened 2026-09-15
fix(dbt): route the remaining raw dedups through the inventory

blarghmatey:

Follow-up to #2650 (#2443).

#2650 resolved the dedup ordering column through the ingestion inventory, but 13 staging models still named Airbyte's metadata columns in their own SQL, so they fail with `COLUMN_NOT_FOUND` when their unit cuts over. The macro's missing-column compile error can't catch them because they never call the resolver. This routes all 13 through it.

- **8 hand-rolled `row_number()` dedups** (6 xpro `app_postgres`, 2 edxorg S3) now call `deduplicate_raw_table(raw_table=...)`.
- **5 compound `order_by` callers** (the four `*__tracking_logs__user_activity` models and `program_entitlement`), which #2650 left out of scope, now call it too.
- **`raw_metadata_column` accepts a list.** The Airbyte S3-source tie-breaker `_ab_source_file_last_modified` is declared per table in the inventory, next to `_airbyte_extracted_at`, instead of in model SQL.
- **`deduplicate_raw_table` gains two parameters.** `then_by` adds a business-column tie-breaker (`program_entitlement`'s purchase date). `source_cte` lets one model dedup two raw tables (`cms_coursepage`, `cms_coursepage_topics`).

**Also fixes a dedup #2650 switched off.** `edxorg/course_structure` is `loader: dagster`, which resolves to no metadata column, so `stg__edxorg__s3__courserun`, `course_policy` and `course_video` compile as pass-throughs on main. The Dagster pipeline writes JSON files to S3, but all four `course_structure` raw tables carry `_airbyte_extracted_at` and `_ab_source_file_*`, the columns Airbyte's S3 source writes (`ol_warehouse_production_raw` in Glue, checked 2026-09-15). Those four tables now declare the column. In `ol_warehouse_production_staging`, the three affected tables were last updated 2026-06-05 (`courserun`, `course_policy`) and 2026-08-13 (`course_video`), both before #2650 merged. `course_xml_blocks` has no table in `ol_warehouse_production_raw` to check, so it keeps the default.

<details>
<summary><b>What the compiled SQL change amounts to</b></summary>
<br>

Compiled each affected model on `main` and on this branch (`-t dev`) and diffed:

- **Ordering is unchanged.** `desc` becomes `desc nulls last`, which Trino's docs give as the default: "The default null ordering is NULLS LAST, regardless of the ordering direction."
- **The tracking-logs models drop the trailing `"time"` sort key.** `"time"` is also in `partition by`, so it is constant within each partition and never sorted anything.
- **`program_learner_report` aggregates from `source` instead of `source_sorted`.** Same rows: `source_sorted` added `row_num` but never filtered, and the aggregates are `min`/`max`.
- **`course_certificate_signatory` keeps its newest-file filter** (`dense_rank` on `_ab_source_file_last_modified`), then dedups through the macro.
- **`courserun`, `course_policy` and `course_video` gain the `row_number()` dedup back.**

**Still coupled, not in this PR:**
- `course_certificate_signatory`'s `dense_rank`. It picks the newest file's snapshot per course, which isn't a dedup, and the inventory has nothing to express that.
- `stg__mitlearn__app__postgres__users_user` and `profiles_profile`. Their `order_by` puts a business column first and `_airbyte_extracted_at` second, which neither the inventory list nor `then_by` expresses.

</details>

### How can this be tested?

```
uv run pytest src/ol_dbt_cli/tests/
uv run ol-dbt inventory validate
uv run ol-dbt inventory metadata-columns    # check mode, exits 1 when stale
cd src/ol_dbt && uv run dbt compile -t dev --select staging
rg -l "row_number\(\) over" src/ol_dbt/models --glob '*.sql' | xargs rg -l "_airbyte_"   # expect nothing
```

What I ran and observed:

- 655 tests pass, including 3 new ones: list resolution, list rendering, and the `course_structure` declaration.
- `ol-dbt inventory validate`: 43 units, 0 errors. Its 8 warnings (xmin cursors) match `main` exactly.
- `dbt compile -t dev --select staging` exits 0.
- The generated map changes only the 9 intended tables.
- The grep returns nothing.
- `prek` passes on every changed file (sqlfluff, yamllint, mypy, ruff).

Not built against Trino. Compiled on DuckDB only.

https://claude.ai/code/session_01Et8ys3HNHVvdFcpFX9NbQy


ingestion/inventory/schema/unit.schema.json
```diff
@@ -52,8 +52,10 @@
...
+      "type": ["string", "array", "null"],
+      "items": {"type": "string"},
+      "minItems": 1,
+      "description": "Column a staging model orders by to pick the newest copy of a record key. Omit to take the loader default: `airbyte` -> `_airbyte_extracted_at`, `dlt` and `dagster` -> null. Null is a real answer meaning \"no metadata column, so do not deduplicate\" — measured, not assumed: the Iceberg schemas dlt writes here carry no metadata columns at all, neither `_airbyte_*` nor `_dlt_*`. Declare explicitly only where a table departs from its loader's convention, such as the two Salesforce tables still on Airbyte v1's `_airbyte_emitted_at`. A list orders by each column in turn, for tables whose models break ties on a second loader column, such as `[_airbyte_extracted_at, _ab_source_file_last_modified]` on Airbyte S3-source tables. Set here it covers every table in the unit; a table's own value wins."
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 98f2e88: description now says dlt -> `_dlt_load_id`, dagster -> null, and drops the stale "no metadata columns" claim.

src/ol_dbt/models/staging/edxorg/stg__edxorg__program_entitlement.sql
```diff
@@ -3,8 +3,9 @@ with source as (
 )
 
 {{ deduplicate_raw_table(
-    order_by="_airbyte_extracted_at desc, cast(date_parse(\"purchase date\", '%m/%d/%Y') as date)",
-    partition_columns='"program title", email'
+    raw_table='raw__edxorg__program_entitlement',
+    partition_columns='"program title", email',
+    then_by="cast(date_parse(\"purchase date\", '%m/%d/%Y') as date)"
```
rachellougee:
```suggestion
   then_by="cast(" ~ date_parse('"purchase date"', "'%m/%d/%Y'") ~ " as date)" 
```
to ensure it works on Trino and DuckDB

src/ol_dbt/models/staging/edxorg/stg__edxorg__s3__course_certificate_signatory.sql
```diff
@@ -1,31 +1,19 @@
-with source as (
-    select *
-    from {{ source('ol_warehouse_raw_data', 'raw__edxorg__s3__course_structure__course_certificate_signatory') }}
-)
-
-, source_sorted as (
+with source_files as (
     select
         *
         , dense_rank() over (partition by course_id order by _ab_source_file_last_modified desc) as dense_row
```
rachellougee: The still uses an Airbyte column `_ab_source_file_last_modified`

rachellougee (commented):
Looks good just two small comments

PR #2713, opened 2026-09-18
 Fix feedback_embeddings model/dim override overwriting existing vectors

### LINKED ISSUE #2688: feedback_embeddings model/dim override overwrites vectors and can mix models in one table
### Expected Behavior
Embedding a sample under a different model or dim keeps the existing vectors, so the embedding bake-off (#2543) can compare models on the same conversations. A run never leaves two models in `feedback_embeddings` without recording it.

### Current Behavior
- The upsert key is `feedback_conversation_pk` alone ([embed.py#L19](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/lib/embed.py#L19), [feedback_embeddings.py#L91](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/assets/feedback_embeddings.py#L91)), so the table holds one vector per conversation.
- `filter_unembedded` re-submits every conversation whose stored model or dim differs from the current run's ([embed.py#L342-L351](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/lib/embed.py#L342-L351)).
- `sample_limit` is applied after that filter ([feedback_embeddings.py#L163-L164](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/assets/feedback_embeddings.py#L163-L164)).

A full override re-embeds all ~190K conversations and overwrites the previous vectors. An override with `sample_limit` moves that many conversations to the new model and leaves the rest on the old one. `feedback_clusters` filters on model and dim ([feedback_clusters.py#L147-L150](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/assets/feedback_clusters.py#L147-L150)), so the next run clusters only the matching subset, and the run row's conversation count has no denominator to show it.

### Steps to Reproduce
Derived from the code, not run:
1. With `feedback_embeddings` populated under the default model, materialize it with run config `{embedding_model_version: "<other model>", sample_limit: 100}`.
2. Group `feedback_embeddings` by `embedding_model_version`: 100 rows are on the new model, the rest on the old one, and the old vectors for those 100 conversations are gone.
3. Materialize `feedback_clusters` with default config: those 100 conversations are excluded.

### Possible Solution
- Add model and dim to the upsert key so both versions coexist, or
- refuse to run when the resolved model/dim differs from what is stored unless `full_refresh` is set, or
- at minimum, document that the override rewrites the table and log the before/after model mix.

### Additional Details
- `sample_limit` means different things in sibling assets: `feedback_summaries` caps the upstream read ([feedback_summaries.py#L109-L110](https://github.com/mitodl/ol-data-platform/blob/777ac4e2820b0001f8aeb0271ac5dc7e8f0d689f/dg_projects/ml/ml/assets/feedback_summaries.py#L109-L110)), while `feedback_embeddings` applies it after the incremental filter.
- Found in a post-merge review of #2622. Blocks #2543.

rachellougee:

Fixes https://github.com/mitodl/ol-data-platform/issues/2688

  - Updates feedback_embeddings upsert key to account for different model and dimension, so overriding them no longer overwrites the existing vectors. 
       - This is a preventive fix for the next model comparison (#2543), not a change to current behavior                                                                  
  - Fixes sample_limit to cap the upstream read, to match feedback_summaries                                                                                
  - Updates int__feedback__embedding to handle multiple embedding rows per conversation. Also removes the embedded_at existence check, since it's already on production   

### How can this be tested?

```
 cd dg_projects/ml                                                                                                                                              
 uv run pytest ml_tests/test_embed.py -q   
```
dbt build --target dev_production --vars 'schema_suffix: <>' --select int__feedback__embedding


rachellougee:
pre-commit.ci run

PR #2722, opened 2026-09-19
feat(delivery): build MIT Learn's edX programs from the edxorg extraction

blarghmatey:
Part of the MIT Learn ETL migration: https://github.com/mitodl/hq/issues/11511. Addresses the edX programs review thread on https://github.com/mitodl/mit-learn/pull/3557#discussion_r4053409162. Stacked on #2720; depends on #2721.

The edX programs delivery read a dlt table that doesn't exist in QA or production: in production `data_loading` has no edX API credentials, and both of its runs there on 2026-09-19 failed with "Missing required credentials". Its payload also had no runs, an edX platform code MIT Learn doesn't have (`edxorg`), and child courses the loader couldn't resolve.

This builds MIT Learn's edX programs from data the `edxorg` code location already extracts from the same edX discovery API, moving the legacy ETL's semantics into dbt, and retires the dlt source.

On a day with no program to deliver, the asset skips delivery, since MIT Learn rejects an empty batch and the legacy ETL left programs published in that case. It then asks MIT Learn which edX programs it still publishes and opens one GitHub issue per program so a person decides whether to unpublish it. Production is in that state today: edX lists no active MIT program in the 2026-09-18 extraction, and MIT Learn still publishes _Circuits and Electronics_, which edX has retired.

<details>
<summary><b>Implementation details</b></summary>
<br>

- `int__edxorg__mitx_learn_programs`: which programs MIT Learn lists (latest extraction; MITx or MITx_PRO; active; not MicroMasters; not titled as deleted) and the program run's dates, summed price, pace, availability and level, from the program courses' runs.
- `int__edxorg__mitx_learn_course_best_runs`: the MITx catalog courses as MIT Learn loads them, with MIT Learn's [`best_run`](https://github.com/mitodl/mit-learn/blob/6286cce8e/learning_resources/models.py#L605) rule. Reusable beyond programs.
- `int__edxorg__mitx_learn_program_rollups` / `_program_instructors`: what legacy read back from MIT Learn's own course records. Topics resolve through the topic crosswalk from #2719, and weeks, weekly hours and instructors come from each published course's best run.
- `integrations__learn__mit_edx_program{s,_instructors}`: what gets delivered. `integrations__learn__mit_edx_programs` changes shape; only the delivery asset reads it.
- `stg__edxorg__api__course` parsed subjects with a Trino-only expression; it now uses a `json_array_field_values` macro so it also builds on DuckDB. Also adds a `json_array_string` macro.
- `MITLearnApiClient.get_published_programs` reads MIT Learn's public programs API, which lists published programs only.
- Review issues go to the asset config's `review_repository` (default `mitodl/mit-learn`). Any issue whose title names the program counts, open or closed: closing one records the decision, so a program that stays unlisted is raised once. Recently updated issues are checked too, since search indexes new issues with a delay.
- A listed program with no courses fails the run before anything is sent, as legacy's `_sum_course_prices` did: MIT Learn's `load_program` replaces a program's course links with the ones delivered, so an empty list would unlink them all. Program courses come from the programs' own extraction, so a program-course sync that lags the programs one ends up here rather than pairing programs with another day's courses.
- Only the latest programs and catalog extractions count, because the raw tables only append. If extractions stop arriving, the models keep the last one, so `integrations__learn__mit_edx_programs` carries the extraction's `retrieved_at` and the asset fails rather than send one more than 3 days old. `raw__edxorg__s3__program` also gets a freshness config with the same threshold, which only takes effect once something runs `dbt source freshness`.
- The dlt `mit_edx_programs` source, its asset, schedule, tests, staging model and inventory unit are removed.

Deliberate differences from legacy:
- The run description is sanitized like the program's. Legacy stored the run's copy raw.
- Instructors are deduplicated across courses.

Not ported: MIT Learn's course blocklist (its `BLOCKLISTED_COURSES_URL` setting). MIT Learn unpublishes blocklisted courses on load, so legacy's rollups left them out; the platform can't see that list, so a blocklisted course in a program would still count towards its topics, instructors and effort.
</details>

### How can this be tested?
There's no active MIT program to compare against, so parity was checked on fixtures derived from mit-learn's edX API fixtures. They cover an upcoming instructor-paced run, a course excluded from search, a course with no published run, a course missing from the catalog, and MicroMasters, retired, deleted-titled and non-MIT programs.

- Legacy side: mit-learn main's `mit_edx` course load and `mit_edx_programs.transform`, with the topic crosswalk matched to the seed.
- Platform side: the real `edxorg` asset code wrote the raw rows, then the dbt models and the delivery asset's `build_resources` produced the payload.
- Every payload key matches, including the DB-derived topics (after the parent expansion MIT Learn applies on load), instructors, weekly hours and duration.
- dbt unit tests pin the best-run tiers (including latest-extraction, deleted-title and zero-maximum-effort handling) and the listing and aggregation rules; the best-run cases were checked against mit-learn's `best_run` property.
- `dbt build` of the edX chain on the fixtures: 74 nodes pass. `ol-dbt validate`: 0 errors.
- `pytest`: delivery 32 pass, data_loading 29, ol-dlt source coverage 12, ol-dbt-cli inventory and QA contract 94, and the new client test.

Only run on DuckDB; the Trino branches haven't been executed.

### Checklist:
- [ ] `raw__edxorg__s3__program` and `raw__edxorg__s3__program_course` land in QA (`ol_warehouse_qa_raw` doesn't have either as of 2026-09-18, while `raw__edxorg__s3__mitx_course` is there), so this chain can be built and exercised in QA.
- [ ] #2721 is deployed, `edxorg_program_metadata` has run, and the Airbyte schema refresh is done, so the new raw columns exist. The staging models here select them.
- [ ] The GitHub token at `secret-data/pipelines/github` (read through `GithubApiClientFactory`) can open issues in the review repository. ol-infrastructure's Dagster server policy, as written, grants read on `secret-data/pipelines/*`; I haven't checked the deployed policy or the token's own repository access.
- [ ] Confirm `mitodl/mit-learn` is where unpublish reviews should go.

https://claude.ai/code/session_01WVriXpD74wThTxt9riaswR


src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
```diff
@@ -0,0 +1,208 @@
...
+with programs as (
+    select *
+    from {{ ref('stg__edxorg__s3__programs') }}
+    where program_retrieved_at = (select max(program_retrieved_at) from {{ ref('stg__edxorg__s3__programs') }})
+)
+
+, program_courses as (
+    select *
+    from {{ ref('stg__edxorg__s3__program_courses') }}
+    where program_course_retrieved_at = (
```
mbertrand: `program_course_retrieved_at` is new in this PR, so every pre-existing row in `raw__edxorg__s3__program_course` has it NULL until the next extraction lands. `x = (select max(x))` over an all-NULL column returns zero rows (checked in duckdb), so this CTE is empty in that window — while the programs CTE above stays populated, because `raw__edxorg__s3__program` already had the column. That asymmetry is what makes it an empty-courses delivery instead of a skip; see the note on line 203.

Separately, the two watermarks are taken independently here. The extraction writes one `data_retrieval_timestamp` per multi_asset, so the real invariant is that they're equal — if one file's Airbyte sync lands and the other's doesn't, this silently pairs the new programs with the previous extraction's courses. Same for `course_retrieved_at` vs `courserun_retrieved_at` in the best-runs model.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
```diff
@@ -0,0 +1,208 @@
...
+    , program_dates.enrollment_start
+    , program_dates.enrollment_end
+    , coalesce(program_prices.price, 0) as price
+    , coalesce(program_prices.currency, 'USD') as currency
+    , program_paces.pace
+    -- a program with no courses is available anytime, as all() of nothing is true
+    , coalesce(program_availability.availability, case when program_course_ids.program_uuid is null then 'anytime' end)
+        as availability
+    , program_course_ids.course_readable_ids
+from listed_programs
```
mbertrand: This is `from listed_programs` plus five left joins and no `where`, so a program with no course rows at all still comes through: `price` coalesces to 0, `currency` to USD, `availability` to anytime, `course_readable_ids` NULL. `rollups` is `from programs left join ...` as well, and the integrations model inner joins it 1:1, so every listed program reaches the payload. The asset turns the NULL into `[]` at mit_edx_programs.py:159 and delivers `"courses": []` with `"published": True`.

So an empty `program_courses` doesn't reach the skip path — it delivers every program with its courses, prices, dates and topics wiped. On the other side `load_program` does `learning_resource.children.all().delete()` and then bulk_creates the empty list (mit-learn `learning_resources/etl/loaders.py:777`), so the program-to-course rows go away.

I don't think this can fire yet: `/api/v1/webhooks/learning_resources/` isn't on mit-learn main (mitodl/mit-learn#3557 is still open) and the client `raise_for_status()`es, so a delivery 404s loudly. It goes live the moment that merges, which argues for closing it here rather than after.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
```diff
@@ -1132,6 +1132,15 @@ models:
       , HarvardX
   - name: program_updated_on
     description: str, timestamp when the data was last modified
+  - name: program_marketing_url
+    description: str, the program's edX marketing URL
+  - name: program_banner_image_url
+    description: str, URL of the program's medium banner image
+  - name: program_level_type_override
+    description: str, the program's level override, e.g. Intermediate
+  - name: program_retrieved_at
```
mbertrand: None of the four `*_retrieved_at` columns has a test, and the whole latest-extraction design rests on them. Worth `not_null` on this one plus `program_course_retrieved_at` (1173), `course_retrieved_at` (1218) and `courserun_retrieved_at` (1303):

```yaml
  - name: program_retrieved_at
    description: str, when the extraction that wrote this row ran. Programs with the
      latest value are the ones the programs API currently lists.
    tests: [not_null]
```

Cheapest half of the fail-open problem — right now a NULL watermark empties the CTE instead of failing the build.

dg_projects/delivery/delivery/assets/mit_edx_programs.py
```diff
@@ -46,86 +59,247 @@
...
+    """
+    repo = github.get_repo(repository)
+    recent_issues = list(
+        repo.get_issues(state="all", since=datetime.now(tz=UTC) - timedelta(days=2))
+    )
+    urls = []
+    for program in programs:
+        readable_id = program["readable_id"]
+        recent = next(
+            (issue for issue in recent_issues if readable_id in issue.title), None
```
mbertrand: This matches any title containing the uuid without checking `_REVIEW_TITLE_PREFIX`, and `get_issues` returns pull requests too — PyGithub doesn't filter them (`pull_request` appears nowhere in `Repository.get_issues`). So a PR whose title mentions a program uuid suppresses its review issue.

```python
        recent = next(
            (
                issue
                for issue in recent_issues
                if issue.title.startswith(_REVIEW_TITLE_PREFIX)
                and readable_id in issue.title
            ),
            None,
        )
```

and the prefix belongs in the search on 230 as well. Because a closed issue counts as a decision, a false match isn't noise — it silences the question permanently, which is worse than the delivery failure mode since that one a re-run fixes.

dg_projects/delivery/delivery/assets/mit_edx_programs.py
```diff
@@ -46,86 +59,247 @@
...
+        "certification": True,
+        "certification_type": "completion",
+        "availability": row["availability"],
+        "format": ["asynchronous"],
+        "pace": row["pace"] or [],
+        # MIT Learn looks program courses up rather than upserting them, so a
+        # reference is enough; ones it doesn't publish are skipped.
+        "courses": [
+            {"readable_id": course_id, "platform": _PLATFORM}
+            for course_id in row["course_readable_ids"] or []
```
mbertrand: The comment four lines up is the right reasoning applied to the other field: if MIT Learn clears topics on `[]` and leaves them alone on `None`, then `[]` is the destructive value for courses too.

Worth refusing the batch instead of delivering it when every row comes through with no courses. That's the one case legacy couldn't express — `_sum_course_prices` raised `IndexError` and aborted the run — so the stated deviation (a course-less program delivered at price 0) also silently covers "every program lost its courses at once".

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand: This can end up sending MIT Learn every edX program with its courses, prices, dates and topics wiped, which makes MIT Learn unlink the program's courses. I'm not sure how likely the window is in practice, but nothing stops it: it happens if the dbt build runs after the new column exists but before the next extraction lands — this block returns no rows, and nothing downstream drops a program whose courses came back empty. Line 203 has that half.

The reason it returns nothing is that the column is new, so the rows already in the raw table have no value for it until the next extraction runs, and asking for the rows whose value equals the largest value finds nothing when every value is missing. I checked that part in duckdb. The programs block just above is fine, because that table already had the column — and that difference is exactly why we get a bad delivery instead of a skip.

Separately, the two largest-value lookups are done separately, but the extraction stamps one timestamp per run, so they should always match. If one file's sync arrives and the other's doesn't, we could pair new programs with the previous run's courses.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand: This is the half that doesn't filter. A program with no courses comes straight through: the course-side joins are all optional and nothing drops those rows, so the price falls back to 0, availability to anytime, and the course list to nothing. Rollups behaves the same way, the model that feeds the webhook keeps one row per program, and the asset turns the missing list into an empty one at mit_edx_programs.py:159 — so we send instead of skipping.

Worth knowing that an empty course list isn't a no-op on MIT Learn's side: load_program deletes the program's existing course links before adding back whatever we sent (loaders.py:777).

It can't happen today, since the endpoint we post to isn't on mit-learn main yet and a delivery just 404s. It can after mitodl/mit-learn#3557 merges.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
mbertrand: None of the four retrieved_at columns has a test, and every "use the latest extraction" filter depends on them having a value. When one is missing the filter quietly matches nothing, which is how we get the emptied-out delivery described on int__edxorg__mitx_learn_programs.sql:203. Adding not_null here and on program_course_retrieved_at (1173), course_retrieved_at (1218) and courserun_retrieved_at (1303) fails the build instead.

```yaml
    tests: [not_null]
```

dg_projects/delivery/delivery/assets/mit_edx_programs.py
mbertrand: The comment just above applies here too: if an empty list clears topics, it clears courses. Worth refusing to send a batch where every program came through with no courses. The old ETL couldn't get into that state — _sum_course_prices raised an error on a program with no courses and took the whole run down with it.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand: This can end up sending MIT Learn every edX program with its courses, prices, dates and topics wiped, which makes MIT Learn unlink the program's courses. I'm not sure how likely the window is in practice, but nothing stops it: it happens if the dbt build runs after the new column exists but before the next extraction lands — this block returns no rows, and nothing downstream drops a program whose courses came back empty. [Line 203](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql#L203) has that half.

The reason it returns nothing is that the column is new, so the rows already in the raw table have no value for it until the next extraction runs, and asking for the rows whose value equals the largest value finds nothing when every value is missing. I checked that part in duckdb. The programs block just above is fine, because that table already had the column — and that difference is exactly why we get a bad delivery instead of a skip.

Separately, the two largest-value lookups are done separately, but the extraction stamps one timestamp per run, so they should always match. If one file's sync arrives and the other's doesn't, we could pair new programs with the previous run's courses.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand: This is the half that doesn't filter. A program with no courses comes straight through: the course-side joins are all optional and nothing drops those rows, so the price falls back to 0, availability to anytime, and the course list to nothing. Rollups behaves the same way, the model that feeds the webhook keeps one row per program, and the asset turns the missing list into an empty one at [mit_edx_programs.py:159](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/dg_projects/delivery/delivery/assets/mit_edx_programs.py#L159) — so we send instead of skipping.

Worth knowing that an empty course list isn't a no-op on MIT Learn's side: load_program deletes the program's existing course links before adding back whatever we sent ([loaders.py:777](https://github.com/mitodl/mit-learn/blob/38df6bfc3e0c87f204d947345e7816cee272995f/learning_resources/etl/loaders.py#L777)).

It can't happen today, since the endpoint we post to isn't on mit-learn main yet and a delivery just 404s. It can after mitodl/mit-learn#3557 merges.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
mbertrand: None of the four retrieved_at columns has a test, and every "use the latest extraction" filter depends on them having a value. When one is missing the filter quietly matches nothing, which is how we get the emptied-out delivery described on [int__edxorg__mitx_learn_programs.sql:203](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql#L203). Adding not_null here and on program_course_retrieved_at ([1173](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1173)), course_retrieved_at ([1218](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1218)) and courserun_retrieved_at ([1303](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1303)) fails the build instead.

```yaml
    tests: [not_null]
```

dg_projects/delivery/delivery/assets/mit_edx_programs.py
mbertrand: The comment just above applies here too: if an empty list clears topics, it clears courses. Worth refusing to send a batch where every program came through with no courses. The old ETL couldn't get into that state — [_sum_course_prices](https://github.com/mitodl/mit-learn/blob/38df6bfc3e0c87f204d947345e7816cee272995f/learning_resources/etl/openedx.py#L391) raised an error on a program with no courses and took the whole run down with it.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand: This column isn't in the extraction output yet, so once Airbyte adds it every row already in the raw table will have no value for it. Asking for the rows whose value equals the largest value finds nothing when every value is missing, so this block returns no rows — and then nothing downstream drops a program whose courses came back empty, which is the emptied-out delivery on [line 203](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql#L203).

I checked today's production extraction and the program-course records still have five fields: program_uuid, course_key, course_title, course_short_description, course_type. So the window is real, but it's narrow: it opens at the Airbyte schema refresh and closes at the next extraction, which runs daily around 01:02 UTC. It also needs edX to be listing at least one active MIT program that day — today it lists none, so we'd skip rather than deliver.

Worth closing anyway, because it reopens on any future schema change to this table and the failure is silent when it happens. The `not_null` tests are the cheap half.

The other way in is that the two largest-value lookups are done independently, and the extraction stamps one timestamp per run, so they should always match. If one file's sync arrives and the other's doesn't, we could pair new programs with the previous run's courses.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
mbertrand, line 200: This is the half that doesn't filter. A program with no courses comes straight through: the course-side joins are all optional and nothing drops those rows, so the price falls back to 0, availability to anytime, and the course list to nothing. Rollups behaves the same way, the model that feeds the webhook keeps one row per program, and the asset turns the missing list into an empty one at [mit_edx_programs.py:159](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/dg_projects/delivery/delivery/assets/mit_edx_programs.py#L159) — so we send instead of skipping.

Worth knowing that an empty course list isn't a no-op on MIT Learn's side: load_program deletes the program's existing course links before adding back whatever we sent ([loaders.py:777](https://github.com/mitodl/mit-learn/blob/38df6bfc3e0c87f204d947345e7816cee272995f/learning_resources/etl/loaders.py#L777)).

It can't happen today, since the endpoint we post to isn't on mit-learn main yet and a delivery just 404s. It can after mitodl/mit-learn#3557 merges.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
mbertrand, line 1141: None of the four retrieved_at columns has a test, and every "use the latest extraction" filter depends on them having a value. When one is missing the filter quietly matches nothing, which is how we get the emptied-out delivery described on [int__edxorg__mitx_learn_programs.sql:203](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql#L203). Adding not_null here and on program_course_retrieved_at ([1173](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1173)), course_retrieved_at ([1218](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1218)) and courserun_retrieved_at ([1303](https://github.com/mitodl/ol-data-platform/blob/33c0b5deca44c955d331ea64be9a9b26cb5faf07/src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml#L1303)) fails the build instead.

```yaml
    tests: [not_null]
```

dg_projects/delivery/delivery/assets/mit_edx_programs.py
mbertrand: The comment just above applies here too: if an empty list clears topics, it clears courses. Worth refusing to send a batch where every program came through with no courses. The old ETL couldn't get into that state — [_sum_course_prices](https://github.com/mitodl/mit-learn/blob/38df6bfc3e0c87f204d947345e7816cee272995f/learning_resources/etl/openedx.py#L391) raised an error on a program with no courses and took the whole run down with it.

src/ol_dbt/models/staging/edxorg/_edxorg_sources.yml
```diff
@@ -1033,6 +1047,15 @@ sources:
...
     - name: course_type
       description: str, type of the course. e.g., audit, verified, professional
+    - name: course_position
+      description: int, 1-based position of the course in the program
+    - name: excluded_from_search
+      description: boolean, whether edX excludes the course from search
+    - name: course_runs
+      description: str, JSON array of the course's runs as the programs API reports
+        them
+    - name: retrieved_at
```
mbertrand, line 1059: Neither `raw__edxorg__s3__program` nor `raw__edxorg__s3__program_course` exists in QA — `ingestion/inventory/qa_observation.json` has both as `present: false` in `ol_warehouse_qa_raw`, observed 2026-09-18, while `raw__edxorg__s3__mitx_course` is there with 448 rows. So none of this chain can be built or exercised in QA as things stand, which is worth sorting out before the #2721 checkbox, since that's where we'd otherwise confirm the new columns arrived.

mbertrand (changes requested):

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
blarghmatey: Fixed in ef2a457: program courses now come from the programs' extraction (`program_course_retrieved_at = max(program_retrieved_at)`) instead of their own latest, so a lagging or missing program-course sync leaves programs with no courses rather than another day's. That case now fails `not_null` on `integrations__learn__mit_edx_programs.course_readable_ids` and the asset refuses to send it (3bb1e43). The unit test has a program-course row from a newer extraction that the old filter picked up.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_programs.sql
blarghmatey: Fixed in 3bb1e43: `build_resources` raises before anything is sent when any listed program has no courses, matching legacy's `_sum_course_prices` failure, since `load_program` would unlink every course. ef2a457 adds `not_null` on `course_readable_ids` in the integrations model so the dbt build flags it too. I dropped the "delivered with price 0" deliberate difference from the description.

src/ol_dbt/models/staging/edxorg/_stg__edxorg__models.yml
blarghmatey: Added `not_null` on `program_retrieved_at`, `course_retrieved_at` and `courserun_retrieved_at` in ef2a457; production raw has no null in any of them. Not on `program_course_retrieved_at`: the latest production extraction has 3532 program-course pairs vs. 3548 in the table, so the 16 pairs edX has dropped keep a pre-#2721 row with no value after dedup and the test would fail on every build. The programs-extraction pairing plus the `course_readable_ids` test covers that column instead.

dg_projects/delivery/delivery/assets/mit_edx_programs.py
blarghmatey: Fixed in 3bb1e43, stricter than every-program: the batch is refused if any program has no courses, as legacy did, because a single empty list still unlinks that program's courses in MIT Learn. The missing list is no longer turned into [].

src/ol_dbt/models/staging/edxorg/_edxorg_sources.yml
blarghmatey: Confirmed against `qa_observation.json`. Getting these two tables landing in QA is outside this PR's code, so I added it to the checklist ahead of the #2721 item. Leaving this open until QA has them.

src/ol_dbt/models/staging/edxorg/_edxorg_sources.yml
```diff
@@ -1018,7 +1018,21 @@ sources:
...
+      description: str, the program's level override
+    - name: retrieved_at
+      description: str, when the extraction that wrote this row ran
+    config:
+      # The edX programs delivery rebuilds from the latest extraction; if the edxorg
+      # asset stops landing new ones it would keep re-sending a stale program set.
+      freshness:
+        warn_after: {count: 2, period: day}
+        error_after: {count: 3, period: day}
+      loaded_at_field: "from_iso8601_timestamp(retrieved_at)"
```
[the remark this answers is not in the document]
blarghmatey: Right, `full_dbt_project` only runs `dbt build`. Fixed in ba62cd4: `integrations__learn__mit_edx_programs` now carries the extraction's `retrieved_at`, and the delivery asset fails instead of sending one more than 3 days old (the config's `error_after`). The source freshness config stays for when `dbt source freshness` gets scheduled; its comment and the docstring now say the asset is the enforcing gate.

src/ol_dbt/models/intermediate/edxorg/int__edxorg__mitx_learn_program_course_runs.sql
```diff
@@ -0,0 +1,68 @@
...
+  price, pace and availability from these runs (learning_resources/etl/openedx.py).
+  The raw table only appends, so only rows from the latest extraction are current.
+#}
+
+with program_courses as (
+    select *
+    from {{ ref('stg__edxorg__s3__program_courses') }}
+    where program_course_retrieved_at = (
+        select max(program_course_retrieved_at) from {{ ref('stg__edxorg__s3__program_courses') }}
+    )
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 32467a5: program_course_runs, program_rollups and program_instructors now read program courses through the same `edxorg_current_program_courses` macro as int__edxorg__mitx_learn_programs, keyed on the programs' `retrieved_at`.

dg_projects/delivery/delivery/assets/mit_edx_programs.py
```diff
@@ -46,86 +65,266 @@
...
+    retry soon after a partial failure would otherwise file duplicates.
+    """
+    repo = github.get_repo(repository)
+    recent_issues = list(
+        repo.get_issues(state="all", since=datetime.now(tz=UTC) - timedelta(days=2))
+    )
+    urls = []
+    for program in programs:
+        readable_id = program["readable_id"]
+        recent = next(
```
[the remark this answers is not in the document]
blarghmatey: Not changing. `get_published_programs` reads MIT Learn's programs API, where `id`, `readable_id` and `title` are required fields (`url` is already read with `.get()` because it can be null). A response missing them means the API contract changed, and a KeyError failing the run is the right result, rather than filing a review issue with placeholder values.

src/ol_dbt/macros/edxorg_current_program_courses.sql
```diff
@@ -0,0 +1,15 @@
...
+      if one stream's sync lands without the other's, programs are left with no
+      courses (which the delivery refuses to send) instead of being paired with
+      another day's. Every model that reads program courses for MIT Learn's edX
+      programs goes through this so they all agree on the extraction.
+    #}
+    select *
+    from {{ ref('stg__edxorg__s3__program_courses') }}
+    where program_course_retrieved_at = (
+        select max(program_retrieved_at) from {{ ref('stg__edxorg__s3__programs') }}
+    )
```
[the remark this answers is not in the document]
blarghmatey: Not changing. Airbyte doesn't set `retrieved_at`: the edxorg_program_metadata asset writes one `data_retrieval_timestamp` per run into both the program and program-course records (`program_record` / `program_course_records` in #2721), so rows from the same extraction match exactly. They only differ when one stream's sync hasn't landed, and failing the delivery in that case is the point.

dg_projects/delivery/delivery/assets/mit_edx_programs.py
```diff
@@ -46,86 +64,291 @@
...
+        context.add_output_metadata(
+            {
+                "delivered_count": 0,
+                "webhook_status": "skipped",
+                "review_issues": MetadataValue.json(issue_urls),
+            }
+        )
+        return {"delivered_count": 0, "webhook_status": "skipped"}
 
+    check_extraction_age(programs_df.iter_rows(named=True), now=datetime.now(tz=UTC))
```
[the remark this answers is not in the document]
blarghmatey: Not changing. The gate stops the asset from re-sending an old extraction to MIT Learn, and the empty path sends nothing. It only opens review issues, at most one per program, and they ask a person to check the program on edX before unpublishing. With no listed program the integrations table has no `retrieved_at` to check, and plumbing one through for that path would only change whether a review issue gets opened.

mbertrand (approved):
👍

